PR #1, opened 2022-11-19
Update interface for verify_kzg_proof

xrchz:
Match the public interface in the specs which takes bytes for the field elements.


bindings/csharp/ckzg.c
```diff
@@ -63,15 +63,12 @@ C_KZG_RET compute_aggregate_kzg_proof_wrap(uint8_t out[48], const Blob blobs[],
...
-  BLSFieldElement fx, fy;
   bool out;
 
-  bytes_to_bls_field(&fx, x);
-  bytes_to_bls_field(&fy, y);
   if (bytes_to_g1(&commitment, c) != C_KZG_OK) return -1;
   if (bytes_to_g1(&proof, p) != C_KZG_OK) return -1;
 
-  if (verify_kzg_proof(&out, &commitment, &fx, &fy, &proof, s) != C_KZG_OK)
+  if (verify_kzg_proof(&out, &commitment, x, y, &proof, s) != C_KZG_OK)
```
kevaundray, line 71: probably want to be consistent in terms of naming the input_point and the output_point, I believe specs use f(z)=y, which is what you also do in the other edited file

PR #2, opened 2022-11-19
Sync C# wrapper changes, update tests, minor changes for c-kzg

flcl42:
C# wrapper changes:
- Add generic test and new setup
- Improve building

C changes:
- Fix minor issue, FILE not being closed
- Extend access to hashing for testing purposes - not necessary needed, I can remove its usage from tests if you see it worsens library interface


src/c_kzg_4844.h
```diff
@@ -121,6 +121,8 @@ C_KZG_RET verify_kzg_proof(bool *out,
     const KZGProof *kzg_proof,
     const KZGSettings *s);
 
+void hash(uint8_t md[32], const uint8_t input[], size_t n);
```
xrchz, line 124: why does this need to be exposed?

xrchz: OK I saw your comment. I prefer to not include it in the interface if you're only using it for tests. If you think it's very valuable to have it, then I would want more comments in the header file and readme saying that it's only there for tests. But preferably just remove it.

PR #4, opened 2022-11-21
Rust bindings

pawanjay176:


bindings/rust/src/lib.rs
```diff
@@ -0,0 +1,273 @@
...
+
+    fn generate_random_blob(rng: &mut ThreadRng) -> Blob {
+        let mut arr: Blob = [0; 131072];
+        rng.fill(&mut arr[..]);
+        arr
+    }
+
+    #[test]
+    fn test_simple() {
+        {
```
michaelsproul, line 241: seems like there's an extra scope `{ .. }` here that can be deleted

pawanjay176:
@xrchz Cleaned this up and it is ready for review now.

xrchz:
Is it possible to add any automatic CI checks for this, that it at least builds or maybe even passes some tests?
Also could check that re-generating the generated file doesn't produce something different than committed?

PR #6, opened 2022-11-22
NodeJS bindings version 1.0.7

dgcoffman:
This version will build on Alpine with newer g++, which is what EthereumJS requires.

### DIFF
--- a/bindings/node.js/Dockerfile
+++ b/bindings/node.js/Dockerfile
@@ -1,6 +1,9 @@
 # Exists as a test harness for building and running tests in Linux
 
-FROM node:16
+
+FROM node:16-alpine
+RUN apk update && apk add --no-cache g++ make python3
+
 
 COPY ./dist/ /app/dist/
 COPY test.ts /app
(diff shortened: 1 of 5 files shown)


bindings/node.js/Dockerfile
```diff
@@ -1,6 +1,9 @@
 # Exists as a test harness for building and running tests in Linux
 
-FROM node:16
+
+FROM node:16-alpine
+RUN apk update && apk add --no-cache g++ make python3
```
dgcoffman, line 5: Makes the `make linux-test` command replicate EthereumJS's environment https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/Dockerfile#L1-L3

dgcoffman: This also makes explicit the build dependencies: a C/C++ compiler, make, and python3.

bindings/node.js/kzg.cxx
```diff
@@ -9,8 +9,8 @@
 #include "blst.h"
 
 Napi::Value throw_invalid_arguments_count(
-  const uint expected,
-  const uint actual,
+  const unsigned int expected,
+  const unsigned int actual,
```
dgcoffman, line 13: Fixes the EthereumJS issue https://github.com/ethereumjs/ethereumjs-monorepo/pull/2349#issuecomment-1322197299

PR #7, opened 2022-11-22
Run NodeJS bindings unit tests in CI

dgcoffman:
Adds a GH actions workflow which runs the NodeJS bindings unit tests in CI.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -1,5 +1,5 @@
 INCLUDE_DIRS = ../inc
-CFLAGS += -O2
+CFLAGS += -O2 -fPIC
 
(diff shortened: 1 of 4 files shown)


src/Makefile
```diff
@@ -1,5 +1,5 @@
 INCLUDE_DIRS = ../inc
-CFLAGS += -O2
+CFLAGS += -O2 -fPIC
```
dgcoffman, line 2: I believe this to be what we want all the time

https://stackoverflow.com/questions/19364969/compilation-fails-with-relocation-r-x86-64-32-against-rodata-str1-8-can-not

PR #9, opened 2022-11-24
Update Fiat-Shamir protocol to match specs

xrchz:


src/c_kzg_4844.c
```diff
@@ -1111,8 +1112,23 @@ static C_KZG_RET hash_to_bytes(uint8_t out[32],
   for (i = 0; i < n; i++)
     bytes_from_g1(&bytes[np + i * 48], &comms[i]);
 
-  hash(out, bytes, np + n * 48);
+  uint8_t* hash_output = calloc(32, sizeof(uint8_t));
```
asn-d6, line 1115: Out of curiosity why do you allocate hash_output on the heap? and not like: `uint8_t hash_output[32] = {0};`

apart from marginally better performance you also get less bookkeepping

src/c_kzg_4844.c
```diff
@@ -1111,8 +1112,23 @@ static C_KZG_RET hash_to_bytes(uint8_t out[32],
...
+  if (hash_output == NULL) { free(bytes); return C_KZG_MALLOC; }
+  hash(hash_output, bytes, nb);
+  memcpy(bytes, hash_output, 32);
+  bytes[32] = 0x0;
```
asn-d6, line 1119: If `n==0` then you won't have enough space in `bytes` for this write.
Maybe add an assert or something if you are sure that `n>0`.

xrchz: You're right - fixed.

PR #15, opened 2022-11-29
Bring compute_challenges() and compute_powers() closer to the spec

asn-d6:
Reading compute_challenges() I noticed that the c-kzg code actually diverges a bit from the spec logic to get performance gains.

In this PR, I attempt to make `compute_challenges()` and `compute_powers()` closer to the spec in terms of readability, as well as remove some funky practices like conditionals inside `calloc()` and using the `bytes` array for multiple uses within the same function.

I split the PR into two mostly orthogonal commits.

### DIFF
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -1121,12 +1121,14 @@ static C_KZG_RET compute_challenges(BLSFieldElement *out, BLSFieldElement r_powe
   uint8_t hashed_data[32] = {0};
   hash(hashed_data, bytes, nb);
 
+  /* We will use hash_input in the computation of both challenges */
+  uint8_t hash_input[33];
+
   /* Compute r */
   uint8_t r_bytes[32] = {0};
-  uint8_t hash_input_0[33] = {0}; // hashed_data + b'\x00'
-  memcpy(hash_input_0, hashed_data, 32);
-  hash_input_0[32] = 0x0;
-  hash(r_bytes, hash_input_0, 33);
+  memcpy(hash_input, hashed_data, 32);
+  hash_input[32] = 0x0;
+  hash(r_bytes, hash_input, 33);
 
   /* Compute r_powers */
   BLSFieldElement r;
@@ -1135,10 +1137,8 @@ static C_KZG_RET compute_challenges(BLSFieldElement *out, BLSFieldElement r_powe
 
   /* Compute eval_challenge */
   uint8_t eval_challenge[32] = {0};
-  uint8_t hash_input_1[33] = {0}; // hashed_data + b'\x01'
-  memcpy(hash_input_1, hashed_data, 32);
-  hash_input_1[32] = 0x1;
-  hash(eval_challenge, hash_input_1, 33);
+  hash_input[32] = 0x1;
+  hash(eval_challenge, hash_input, 33);
   bytes_to_bls_field(out, eval_challenge);
 
   free(bytes);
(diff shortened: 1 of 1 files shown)


src/c_kzg_4844.c
```diff
@@ -825,10 +825,12 @@ void free_trusted_setup(KZGSettings *s) {
   free_kzg_settings(s);
 }
 
-static void compute_powers(fr_t out[], uint64_t n) {
-    out[0] = fr_one;
-    for (uint64_t i = 2; i < n; i++)
-      fr_mul(&out[i], &out[i-1], &out[1]);
+static void compute_powers(fr_t out[], BLSFieldElement *x, uint64_t n) {
```
xrchz, line 828: It's probably better for readability to call them all `fr_t` or all `BLSFieldElement` rather than mixing them.

src/c_kzg_4844.c
```diff
@@ -1098,34 +1100,46 @@ static C_KZG_RET compute_challenges(BLSFieldElement *out, BLSFieldElement r_powe
...
-  bytes[32] = 0x1;
-  hash(hash_output, bytes, 33);
-  bytes_to_bls_field(out, hash_output);
+  /* Now let's create challenges! */
+  uint8_t hashed_data[32] = {0};
+  hash(hashed_data, bytes, nb);
+
+  /* Compute r */
+  uint8_t r_bytes[32] = {0};
+  uint8_t hash_input_0[33] = {0}; // hashed_data + b'\x00'
```
xrchz, line 1126: This one I would prefer to reuse (just one `hash_input` instead of two). I'm fine with not re-using `bytes` to improve readability though.

src/c_kzg_4844.c
```diff
@@ -1098,34 +1100,46 @@ static C_KZG_RET compute_challenges(BLSFieldElement *out, BLSFieldElement r_powe
...
-
-  bytes[32] = 0x1;
-  hash(hash_output, bytes, 33);
-  bytes_to_bls_field(out, hash_output);
+  /* Now let's create challenges! */
+  uint8_t hashed_data[32] = {0};
+  hash(hashed_data, bytes, nb);
+
+  /* Compute r */
+  uint8_t r_bytes[32] = {0};
```
xrchz, line 1125: Would at prefer to allocate only one of `r_bytes` and `eval_challenge`, and possibly neither.

src/c_kzg_4844.c
asn-d6: For the record, stack allocation is an extremely cheap (like [usually a single CPU instruction](https://stackoverflow.com/a/161073)) so IMO this is mostly a question of readability than performance. If you think the code is cleaner with a single `hash_input` then let's go with that. Other than that, the performance overhead should be negligible and we have the luxury to optimize for niceness :)

(To be frank, in the above there is also the extra tiny overhead of the zeroing with `= {0};`. In this code the zeroing is not really necessary as we are writing over the entire array, and can be omitted for extra (negligible) "speed". Negligible because memset is also extremely fast.)

src/c_kzg_4844.c
asn-d6: Similar to the case above, this is largely a matter of readability. I thought that these two extra variables make the code look closer to the spec. If you think they are confusing, we can remove them!

xrchz: I could go either way on these, but would find it clearer with reusing `hashed_data` for all hash outputs. (It could be called `hash_output` if the other one is called `hash_input`.)

src/c_kzg_4844.c
asn-d6: Done in bcf14e7

src/c_kzg_4844.c
asn-d6: Done in bdfa79c

src/c_kzg_4844.c
asn-d6: Since you could go either way I went naughty and didn't change this part of the code. I think this way it's more verbose but the variable names stay close to the spirit of the spec. IMO that will be better for people who review/audit the code.

PR #16, opened 2022-11-29
Make FIELD_ELEMENTS_PER_BLOB compile-time configurable

xrchz:


bindings/java/Makefile
```diff
@@ -2,7 +2,9 @@ INCLUDE_DIRS = ../../src ../../blst/bindings
 
 LIBRARY_FOLDER=src/main/resources/ethereum/ckzg4844/lib
 
-CC_FLAGS=
+FIELD_ELEMENTS_PER_BLOB ?= 4096
+
+CC_FLAGS = -DFIELD_ELEMENTS_PER_BLOB=$(FIELD_ELEMENTS_PER_BLOB)
```
StefanBratanov, line 7: Is it possible to add this in the command itself similar to src/Makefile? We override the CC_FLAGS in our CI to create the java library itself.

StefanBratanov:
Is it possible as part of this PR a version of https://github.com/ethereum/c-kzg-4844/blob/main/src/trusted_setup.txt to be included which will work with FIELD_ELEMENTS_PER_BLOB set to 4

xrchz:
If you have one handy I'm happy to add it :)

xrchz:
Thanks for suggesting this - it uncovered some bugs in the PR.

bindings/node.js/binding.gyp
```diff
@@ -8,7 +8,8 @@
         "CLANG_CXX_LIBRARY": "libc++",
         "MACOSX_DEPLOYMENT_TARGET": "13.0"
       },
-      "defines": ["NAPI_DISABLE_CPP_EXCEPTIONS"],
+      "defines": ["NAPI_DISABLE_CPP_EXCEPTIONS",
+                  "FIELD_ELEMENTS_PER_BLOB=<!(echo ${FIELD_ELEMENTS_PER_BLOB:-4096})"],
```
dgcoffman, line 12: This also needs to go in `binding.dist.gyp`

Issue #11, opened 2022-11-25
check for non-canonical field element representations


xrchz:
Considering the conversation here https://github.com/ethereum/consensus-specs/issues/3057 and ongoing discussion, just want to record that the C library may need to detect and fail on non-canonical inputs (i.e., greater than the modulus).

sauliusgrigaitis:
Key question - is there a valid case that MODULUS <= value <= MAX_DATA_TYPE_VALUE at the application level. Seems not. So the crypto library should just throw an informative error that it got a value that doesn't fit into MODULUS. This would also help to debug the application.

sauliusgrigaitis:
I would vote for adding an informative error at the crypto library just because at the application level most of the devs are not that familiar with cryptography. By application level, I mean client level.

xrchz:
Does this need to happen for the arguments of any other API functions than `verify_kzg_proof`?
Or is it all the functions that take a `blob` or `blobs` too?

xrchz:
Following the latest specs (after https://github.com/ethereum/consensus-specs/commit/23d3aeebba3b5da0df4bd25108461b442199f406) I think the way to do it is put the check in `bytes_to_bls_field` (and have an unchecked version `hash_to_bls_field` for `compute_challenges`)

xrchz:
Closed by #20

PR #20, opened 2022-12-02
Add checks that field elements are canonical

xrchz:
Addresses #11


kevaundray:
Hmm bindings are failing here, might need to ping the respective code owners about making sure the blobs are actually canonical and not just random bytes

StefanBratanov:
> Hmm bindings are failing here, might need to ping the respective code owners about making sure the blobs are actually canonical and not just random bytes

Once this PR is rebased, I can work on fixing the java bindings. Think only the pointEvaluation test needs fixing. All random blobs are canonical in the java tests after the last PR.

PR #21, opened 2022-12-06
allocation_check

Ahaif:
check for allocation in case of failure


xrchz:
Thanks I've implemented this here instead: 2bd4c3b

PR #28, opened 2022-12-12
Do not allow blob fields to overflow

dgcoffman:
https://github.com/ethereum/c-kzg-4844/pull/20 causes the NodeJS bindings tests to fail because they generate truly random bytes, and that isn't valid blob data.

This PR makes the test blobs be valid.


bindings/node.js/test.ts
```diff
@@ -50,8 +62,7 @@ describe("C-KZG", () => {
     );
   });
 
-  // Just don't call verifyAggregateKzgProof when there are no blobs or commitments
-  it.skip("verifies the aggregate proof of empty blobs and commitments", () => {
```
dgcoffman, line 54: this test passes now

PR #31, opened 2022-12-13
Correctly set FIELD_ELEMENTS_PER_BLOB for distributable build

dgcoffman:


bindings/node.js/binding.dist.gyp
```diff
@@ -38,6 +40,7 @@
           "action": [
             "cc",
             "-I<(module_root_dir)/dist/deps/blst/bindings",
+            "-DFIELD_ELEMENTS_PER_BLOB=<!(echo ${FIELD_ELEMENTS_PER_BLOB:-4096})",
```
dgcoffman, line 43: This is the fix. The C compilation of c_kzg_4844 requires this define.

PR #37, opened 2022-12-17
Make the interface bytes-only

### LINKED ISSUE #27: Make the interfaces all bytes-only
Currently there's a mix of byte arrays and internal types (like `g1_t`) in the interface. It should all be byte arrays. The conversion functions (like `bytes_to_g1`) should be removable from the interface.

xrchz:
Fixes #27

### DIFF
--- a/src/c_kzg_4844.h
+++ b/src/c_kzg_4844.h
@@ -40,14 +40,9 @@ extern "C" {
 #define BYTES_PER_BLOB FIELD_ELEMENTS_PER_BLOB * BYTES_PER_FIELD_ELEMENT
 static const char *FIAT_SHAMIR_PROTOCOL_DOMAIN = "FSBLOBVERIFY_V1_";
 
-typedef blst_p1 g1_t;         /**< Internal G1 group element type */
-typedef blst_p2 g2_t;         /**< Internal G2 group element type */
-typedef blst_fr fr_t;         /**< Internal Fr field element type */
-
-typedef g1_t KZGCommitment;
-typedef g1_t KZGProof;
-typedef fr_t BLSFieldElement;
 typedef uint8_t Blob[BYTES_PER_BLOB];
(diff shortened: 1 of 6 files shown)


bardo-hub (approved):

src/c_kzg_4844.h
```diff
@@ -40,14 +40,9 @@ extern "C" {
...
 
-typedef blst_p1 g1_t;         /**< Internal G1 group element type */
-typedef blst_p2 g2_t;         /**< Internal G2 group element type */
-typedef blst_fr fr_t;         /**< Internal Fr field element type */
-
-typedef g1_t KZGCommitment;
-typedef g1_t KZGProof;
-typedef fr_t BLSFieldElement;
 typedef uint8_t Blob[BYTES_PER_BLOB];
+typedef uint8_t KZGCommitment[BYTES_PER_COMMITMENT];
```
asn-d6, line 44: All in all this PR looks sensible but with one caveat:

I'm not the biggest fan of these kind of `typedefs` (e.g. `uint8_t KZGCommitment[BYTES_PER_COMMITMENT];`)

On one hand it's convenient. On the other hand, it leads to surprising stuff like:
```
- C_KZG_RET blob_to_kzg_commitment(KZGCommitment *out, const Blob blob, const KZGSettings *s) {
+ C_KZG_RET blob_to_kzg_commitment(KZGCommitment out, const Blob blob, const KZGSettings *s) {
```
where the out argument is not seen as a pointer.

But my biggest fear is that it might be a footgun in bindings: where someone passes a KZGCommitment to a function and expects that its sizeof() will return the object's size, but in reality it will return the pointer size.

I guess the alternative approach would be to make KZGCommitment a struct instead of a typedef which will result in more boilerplate but also more traditional pointer style.

Some additional context and examples can be found [here](https://stackoverflow.com/questions/4523497/typedef-fixed-length-array).

xrchz:
Replaced by #62

PR #38, opened 2022-12-17
Fully free FFTSettings

benjaminion:
Plug a little memory leak associated with `FFTSettings`.

### DIFF
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -701,7 +701,7 @@ static void free_fft_settings(FFTSettings *fs) {
     free(fs->expanded_roots_of_unity);
     free(fs->reverse_roots_of_unity);
     free(fs->roots_of_unity);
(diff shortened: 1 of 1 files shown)


xrchz:
Nice find!

It might be good to use different naming schemes for `free_fft_settings` and `free_kzg_settings`, since now the former also frees the passed pointer but the latter does not. Alternatively manually call this free outside the function (at its sole call site) instead.

benjaminion:
Valgrind is my friend :)

Yes, it could be cleaner. Mostly just keen to see if it resolves the issues for the Java API fuzzing. I'm happy for you to tidy it up as you see fit.

xrchz:
My alt: #39 
If that also looks good to you I'll use it.

benjaminion:
Heh, looks good. I did exactly the same once I'd decided not to be so lazy.

PR #41, opened 2022-12-18
byte_only_csharp

Ahaif:
make c-kzg wrapper in csharp binding bytes-only as c-kzg interface

### DIFF
--- a/bindings/csharp/ckzg.c
+++ b/bindings/csharp/ckzg.c
@@ -4,6 +4,8 @@
 #include "c_kzg_4844.h"
 #include "ckzg.h"
 
+
+
 KZGSettings* load_trusted_setup_wrap(const char* file) {
   KZGSettings* out = malloc(sizeof(KZGSettings));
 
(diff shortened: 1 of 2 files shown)


bindings/csharp/ckzg.c
```diff
@@ -24,29 +26,17 @@ void free_trusted_setup_wrap(KZGSettings *s) {
   free(s);
 }
 
-void blob_to_kzg_commitment_wrap(uint8_t out[48], const Blob blob, const KZGSettings *s) {
+C_KZG_RET blob_to_kzg_commitment_wrap(uint8_t out[48], const Blob blob, const KZGSettings *s) {
```
xrchz, line 29: You need to put the result in `out` in this function - in the proposed changes `out` has become ignored.

bindings/csharp/ckzg.c
```diff
@@ -26,32 +23,34 @@ void free_trusted_setup_wrap(KZGSettings *s) {
   free(s);
 }
 
-C_KZG_RET blob_to_kzg_commitment_wrap(uint8_t out[48], const Blob blob, const KZGSettings *s) {
-  KZGCommitment c;
-  C_KZG_RET ret = blob_to_kzg_commitment(c, blob, s);
+C_KZG_RET blob_to_kzg_commitment_wrap(KZGCommitment out, const Blob blob, const KZGSettings *s) {
+
+  C_KZG_RET ret = blob_to_kzg_commitment(out, blob, s);
   return ret;
```
xrchz, line 29: can't you just return this directly, without the intermediate variable `ret`?

bindings/csharp/ckzg.h
```diff
@@ -9,14 +12,21 @@
 #define DLLEXPORT
 #endif
 
+typedef blst_p1 g1_t;         /**< Internal G1 group element type */
```
xrchz, line 15: what are these needed for?

xrchz:
@flcl42 - what do you think?

xrchz:
Replaced in #62

PR #43, opened 2022-12-19
Add ARM support for Java bindings

jtraglia:
Tried to build/test the Java bindings but my platform (ARM64) wasn't supported by the Makefile.

* On Linux, `uname -m` reports "x86_64" instead of "amd64".
* On macOS, `uname -m` reports "arm64" instead of "aarch64" like Linux does.

### DIFF
--- a/bindings/java/Makefile
+++ b/bindings/java/Makefile
@@ -19,7 +19,11 @@ else
   ifeq ($(UNAME_S),Linux)
     CLANG_FLAGS=-fPIC -shared
     JNI_INCLUDE_FOLDER=linux
-    OS_ARCH=$(UNAME_M)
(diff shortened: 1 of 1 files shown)


StefanBratanov:
I like having the defaults in the Makefile. Thanks! For reference, currently we hardcode them here for ARM machines: https://github.com/ConsenSys/jc-kzg-4844/blob/master/.circleci/config.yml

jtraglia:
Ah I see now. Thanks for pointing that out.

xrchz:
Is this still supposed to be merged? It's a bit unclear from the discussion.

jtraglia:
Yes, please merge. The link he shared just tells us how ConsenSys builds the library for all platforms. This PR is still useful.

PR #44, opened 2022-12-19
Check that malloc'd settings isn't null

jtraglia:
I think it's a good idea to check the return values for these two `malloc` calls.

In the event of memory exhaustion, it will provide a better error message than a segfault crash will.


jtraglia:
Oh duh, you're right. Good catch. Will fix.

PR #45, opened 2022-12-19
Refactor error handling

jtraglia:
I noticed that when the `TRY` macro is used, it doesn't free the previous allocations if there's a problem. The safer alternative is to have a single exit point & free the necessary variables there. I cleaned up those & a few other missing free calls.

Additionally, this PR will:

* Remove the `TRY` macro.
* Replace calls to `malloc` with `c_kzg_malloc`.
* Add some return value checks that were missing.

I know `goto` calls could be a point of contention, but personally I like them. Reduces code duplication and makes it more human readable. I'm somewhat used to Linux kernel development, so maybe I'm biased.

For example, this PR will convert this...

to something like this:

In this example, there is a single exit point AND it frees the memory allocated for the trusted setup (`fs`, `g1_values`, and `g2_values`) if something goes wrong, like if the call to `reverse_bit_order` fails. This was not being done before.

### DIFF
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -811,10 +811,6 @@ C_KZG_RET load_trusted_setup(KZGSettings *out, const uint8_t g1_bytes[], size_t 
   if (ret != C_KZG_OK) goto error_free_out;
   ret = fft_g1(out->g1_values, g1_projective, true, n1, out->fs);
   if (ret != C_KZG_OK) goto error_free_out;
-
-  free(g1_projective);
-  g1_projective = NULL;
-
   ret = reverse_bit_order(out->g1_values, sizeof(g1_t), n1);
   if (ret != C_KZG_OK) goto error_free_out;
 
@@ -824,8 +820,8 @@ error_free_out:
   if (out->fs != NULL) free((void *)out->fs);
   if (out->g1_values != NULL) free(out->g1_values);
   if (out->g2_values != NULL) free(out->g2_values);
-  if (g1_projective != NULL) free(g1_projective);
 success_out:
+  if (g1_projective != NULL) free(g1_projective);
   return ret;
 }
 
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -264,7 +264,7 @@ static C_KZG_RET fr_batch_inv(fr_t *out, const fr_t *a, size_t len) {
     size_t i;
 
     ret = new_fr_array(&prod, len);
-    if (ret != C_KZG_OK) goto free_out;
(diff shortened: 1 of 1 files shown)


src/c_kzg_4844.c
```diff
@@ -989,22 +1006,28 @@ C_KZG_RET verify_kzg_proof(bool *out,
...
+  ret = new_fr_array(&inverses, FIELD_ELEMENTS_PER_BLOB);
+  if (ret != C_KZG_OK) goto out_error;
+
   for (i = 0; i < FIELD_ELEMENTS_PER_BLOB; i++) {
     if (fr_equal(x, &roots_of_unity[i])) {
       *out = p[i];
-      free(inverses_in);
-      free(inverses);
-      return C_KZG_OK;
+      goto out_error;
```
xrchz, line 1024: this should set `ret` to `C_KZG_OK` first. the label is also a bit misleading here.

src/c_kzg_4844.c
```diff
@@ -1194,32 +1221,43 @@ C_KZG_RET compute_aggregate_kzg_proof(KZGProof *out,
...
     const KZGSettings *s) {
-  KZGCommitment* commitments = calloc(n, sizeof(KZGCommitment));
-  if (0 < n && commitments == NULL) return C_KZG_MALLOC;
+  C_KZG_RET ret;
+  Polynomial* polys = NULL;
+  KZGCommitment* commitments = NULL;
+
+  commitments = calloc(n, sizeof(KZGCommitment));
+  if (0 < n && commitments == NULL)
+  {
```
xrchz, line 1230: nit: brace formatting inconsistent with the prevailing style

src/c_kzg_4844.c
jtraglia: Oh good catch. And yeah, I'll fix that.

xrchz: Maybe the labels should all be `out`? I think there's only one case where the `_error` part is needed to distinguish from `success`?

jtraglia: Sounds good to me. Fixed. For some reason I thought there would be a conflict with the `out` parameter.

src/c_kzg_4844.c
jtraglia: Will fix. Also, I was planning to do a formatting PR later.

* Use consistent indentation (it's split between 4 spaces & 2 spaces).
* Use consistent brackets. For this PR I used the func's style.

jtraglia: Will fix. Also, I plan to make a PR later to use consistent indentation. It's split between 4 spaces & 2 spaces.

PR #46, opened 2022-12-19
Add wrapper, improve build

flcl42:
Covers #150 
Depends on https://github.com/ethereum/c-kzg-4844/pull/165, merge after it

- Add support for ARM
- Add wrapper with checks
- Clean up tests


asn-d6:
Great work here! What kind of testing have you performed on this code, and how can we test it ourselves as part of review?

bindings/csharp/Ckzg.Bindings/Ckzg.cs
```diff
@@ -1,115 +1,231 @@
...
-    /// Given a blob and a KZG proof, verify that the blob data corresponds to the provided commitment.
+    ///     Given a blob and a KZG proof, verify that the blob data corresponds to the provided commitment.
     /// </summary>
-    /// <param name="result">True if the proof is valid</param>
     /// <param name="blob"></param>
     /// <param name="commitment"></param>
     /// <param name="proof"></param>
-    /// <param name="ts">Trusted setup settings</param>
+    /// <param name="ckzgSetup">Trusted setup settings</param>
     /// <returns></returns>
```
asn-d6, line 182: I think it's important to document the return value for these verify functions.

asn-d6 (changes requested):

bindings/csharp/Ckzg.Bindings/Ckzg.cs
```diff
@@ -1,115 +1,231 @@
...
+
+        IntPtr ckzgSetup = InternalLoadTrustedSetup(filename);
+
+        if (ckzgSetup == IntPtr.Zero) throw new InvalidOperationException("Unable to load trusted setup");
+        return ckzgSetup;
+    }
 
     /// <summary>
-    /// Frees memory allocated for trusted setup settings
+    ///     Frees memory allocated for trusted setup settings
```
jtraglia, line 54: Missing punctuation here.

bindings/csharp/Ckzg.Bindings/Ckzg.cs
```diff
@@ -1,115 +1,231 @@
...
-                Architecture.X64 => "x64",
-                Architecture.Arm64 => "arm64",
-                _ => ""
-            }}/native/{path}.{(RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "dll" : "so")}");
+        return false;
+    }
 
     /// <summary>
-    /// Load trusted setup settings from file
+    ///     Load trusted setup settings from file
```
jtraglia, line 39: Missing punctuation here.

bindings/csharp/Ckzg.Bindings/Ckzg.cs
```diff
@@ -1,115 +1,231 @@
...
     /// <param name="proofs">Proofs as a flattened byte array</param>
     /// <param name="count">The number of blobs/commitments/proofs</param>
-    /// <param name="ts">Trusted setup settings</param>
+    /// <param name="ckzgSetup">Trusted setup settings</param>
     /// <returns></returns>
-    [DllImport("ckzg", EntryPoint = "verify_blob_kzg_proof_batch", CallingConvention = CallingConvention.Cdecl)]
-    public unsafe static extern Ret VerifyBlobKzgProofBatch(bool* result, byte* blobs, byte* commitments_bytes, byte* proofs_bytes, int count, IntPtr ts);
-}
+    public static unsafe bool VerifyBlobKzgProofBatch(ReadOnlySpan<byte> blobs, ReadOnlySpan<byte> commitments,
+        ReadOnlySpan<byte> proofs, int count, IntPtr ckzgSetup)
```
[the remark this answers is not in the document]
asn-d6: Maybe the idea is that this way the bindings don't need to go to the effort of allocating a flat array to pass it to c-kzg (like the node bindings do).

bindings/csharp/Ckzg.Bindings/Ckzg.cs
```diff
@@ -1,115 +1,231 @@
...
+    {
+        if (blobs.Length != BytesPerBlob * count) throw new ArgumentException("Invalid blob size", nameof(blobs));
+
+        if (commitments.Length != BytesScalar * count)
```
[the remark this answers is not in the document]
flcl42: Yep it's just a PoC to get feedback asap

bindings/csharp/Ckzg.Bindings/Ckzg.cs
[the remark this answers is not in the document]
flcl42: Yep, we can probably optimize by removing splitting while deserializing from SSZ and than flatten while passing to c-kzg. Still can be two interfaces if a consumer will want to verify a batch of blobs from different transactions or like that.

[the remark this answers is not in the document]
asn-d6: I guess the ideal approach would be to pass a list of byte arrays (so that you don't need `count`) but then be able to interpret that list as a flat array of bytes inside C so that you can pass them to verify_batch in c-kzg. You can do this in languages [like Rust](https://stackoverflow.com/a/68905410) but I don't know enough about C#.

No strong opinion from my part, as long as its convenient from the client's PoV and the code is reasonable (which I think it currently is).

bindings/csharp/Ckzg.Bindings/Ckzg.cs
flcl42: Could you clarify

jtraglia: Ah yeah. Just add a period to the end, like:

```
///     Frees memory allocated for trusted setup settings.
```

bindings/csharp/Ckzg.Test/ReferenceTests.cs
```diff
@@ -0,0 +1,337 @@
...
+    }
+
+    private IntPtr _ts;
+    private const string TestDir = "../../../../../../tests";
+    private readonly string _blobToKzgCommitmentTests = Path.Join(TestDir, "blob_to_kzg_commitment");
+    private readonly string _computeKzgProofTests = Path.Join(TestDir, "compute_kzg_proof");
+    private readonly string _computeBlobKzgProofTests = Path.Join(TestDir, "compute_blob_kzg_proof");
+    private readonly string _verifyKzgProofTests = Path.Join(TestDir, "verify_kzg_proof");
+    private readonly string _verifyBlobKzgProofTests = Path.Join(TestDir, "verify_blob_kzg_proof");
+    private readonly string _verifyBlobKzgProofBatchTests = Path.Join(TestDir, "verify_blob_kzg_proof_batch");
```
jtraglia, line 37: I believe the `TestDir` variable should be renamed:

```suggestion
    private const string _testDir = "../../../../../../tests";
    private readonly string _blobToKzgCommitmentTests = Path.Join(_testDir, "blob_to_kzg_commitment");
    private readonly string _computeKzgProofTests = Path.Join(_testDir, "compute_kzg_proof");
    private readonly string _computeBlobKzgProofTests = Path.Join(_testDir, "compute_blob_kzg_proof");
    private readonly string _verifyKzgProofTests = Path.Join(_testDir, "verify_kzg_proof");
    private readonly string _verifyBlobKzgProofTests = Path.Join(_testDir, "verify_blob_kzg_proof");
    private readonly string _verifyBlobKzgProofBatchTests = Path.Join(_testDir, "verify_blob_kzg_proof_batch");
```

jtraglia (approved):
Looks great, I only have one little nit:

bindings/csharp/Ckzg.Test/ReferenceTests.cs
jtraglia: Also, should it be `readonly` too? What's the difference between `const` and `readonly`?

jtraglia: Ah nevermind, I just [googled it](https://stackoverflow.com/a/56024) and it makes sense as a `const` value.

flcl42: `const` is for primitive values only, we can't init it with calls also. `readonly` covers the rest of the cases for immutable fields. 

Not very significant fact: `const` behaves more like C macros.

flcl42: The default convention is pascal case for constants. Looks a bit weird, but will show warnings in IDEs otherwise.

jtraglia: Gotcha. Weird but I understand.

jtraglia (approved):
LGTM 👍

PR #49, opened 2022-12-20
Add go bindings (and fuzzing!)

jtraglia:
Thought it would be really cool if this project had some fuzzers. This PR provides a basic Go binding & a collection of fuzzing methods for the exported KZG functions. Ideally, the next steps would be to import [go-kzg](https://github.com/protolambda/go-kzg), fuzz both at the same time, and compare the results. I'd like for that to be a separate PR though.

Thoughts?


jtraglia:
Hmm, I've encountered an interesting issue.

Using the same go version... on amd64, I get this error, but it works fine on arm64:

```
cannot use (*_Ctype_Blob)(unsafe.Pointer(&blobs)) (value of type *_Ctype_Blob) as type *[131072]_Ctype_uchar in variable declaration
```

On arm64, I need this cast to be:

```go
(*C.Blob)(unsafe.Pointer(&blobs))
```

On amd64, I need this cast to be:

```go
(*[blobSize]C.uchar)(unsafe.Pointer(&blobs)),
```

Which doesn't work on arm64 [🤦](https://emojipedia.org/person-facepalming/)

jtraglia:
Hmm, I'm having trouble converting this to a package. Go discards most files in the repository. I don't think it's possible to use `embed` with cgo. Not sure what can be done here. I think this is still useful for fuzzing though.

jtraglia:
If I can't figure this out, should we move this out of `bindings` and into a new `fuzz` directory, like I originally had it?

jtraglia:
So I think our next best option is to a plugin "plugin" build.

```
go build -buildmode=plugin -o gockzg.so
```

Then you can do something like:

```go
p, err := plugin.Open("gockzg.so")
if err != nil {
    panic(err)
}

symbol, err := p.Lookup("BytesFromG1")
if err != nil {
    panic(err)
}

BytesFromG1, ok := symbol.(func([144]byte) [48]byte)
if !ok {
    panic("missing func")
}

var g1 [144]byte
ret := BytesFromG1(g1)
fmt.Println(ret)
```

I'll continue to explore this. Unsure how to handle the C return types.

jtraglia:
Regarding plugins, nevermind. It would be a big hassle. You can't do static builds and, as a result, there would need to be a plugin for each architecture/libc/golang combination, which would be a whole lot. If we didn't need `libblst`, we could probably find a way to make this a package. And since there's `go-kzg`, these bindings aren't strictly necessary. Unless someone knows how to do this, I suggest we keep the bindings for fuzzing and differential fuzzing with `go-kzg`.

jtraglia:
Closing. Going to make a separate repository for this work.

PR #51, opened 2022-12-24
Memset Settings structs before modifying them

z3n-chada:
When using new_fft_settings and load_trusted_setup the structure memory for FFTSettings/KZGSettings is not guaranteed to be zeroed out. In error cases, its possible that they are not NULL. As a result we may unintentionally free these pointers.

@jtraglia and I found this while doing a review.

### DIFF
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -663,6 +663,8 @@ static C_KZG_RET new_fft_settings(FFTSettings *fs, unsigned int max_scale) {
     C_KZG_RET ret;
     fr_t root_of_unity;
 
+    memset(fs, 0, sizeof(FFTSettings));
+
     fs->max_width = (uint64_t)1 << max_scale;
 
     CHECK((max_scale < sizeof scale2_root_of_unity / sizeof scale2_root_of_unity[0]));
@@ -784,6 +786,8 @@ C_KZG_RET load_trusted_setup(KZGSettings *out, const uint8_t g1_bytes[], size_t 
   blst_p2_affine g2_affine;
   g1_t *g1_projective = NULL;
   C_KZG_RET ret;
+
+  memset(out, 0, sizeof(KZGSettings));
 
   ret = new_g1_array(&out->g1_values, n1);
   if (ret != C_KZG_OK) goto out_error;


jtraglia:
Thanks @z3n-chada. This is something that I should have done in #45. My bad. Glad we spotted this.

xrchz:
I'm not sure I understand how `memset` to 0 helps with the issue of them possibly being or not being `NULL` - would you mind explaining it a bit more?

jtraglia:
Of course, happy to explain. I'll do the `FFTSettings` one because it's simpler.

So the memory for `FFTSettings` comes from `c_kzg_malloc` (`malloc`):

https://github.com/ethereum/c-kzg-4844/blob/08a1cdcd379c0093467a2e952c8c523666c2a4b8/src/c_kzg_4844.c#L808-L811

This memory is uninitialized and could be anything. It could be zerod out memory but it just as likely contains data from a previous allocation. This means that the pointers in `FFTSettings` may not be null. See [this stack overflow answer](https://stackoverflow.com/a/66840190) for more details.

> The `malloc` function allocates space for an object whose size is specified by `size` and whose value is indeterminate.

Given the following block of code, imagine it fails at the first allocation. The `reverse_roots_of_unity` and `roots_of_unity` pointers will remain uninitialized.

https://github.com/ethereum/c-kzg-4844/blob/08a1cdcd379c0093467a2e952c8c523666c2a4b8/src/c_kzg_4844.c#L672-L677

During this cleanup section, those two pointers will likely not be null. As a result, `free` will be called on whatever indeterminate value it was. This will cause the program to crash.

https://github.com/ethereum/c-kzg-4844/blob/08a1cdcd379c0093467a2e952c8c523666c2a4b8/src/c_kzg_4844.c#L695-L700

The solution is to `memset` the entire structure so we _know_ those values start as null.

xrchz:
I'm not sure that memset does what you need it to here actually. See https://stackoverflow.com/questions/29227487/memset-an-array-of-void-pointers

Also this seems to have broken something in the csharp bindings?

jtraglia:
So these are pointers in a structure. The `memset` zeros them out (makes them null).

```c
fs->expanded_roots_of_unity = (void*)0x11111111;
fs->reverse_roots_of_unity = (void*)0x22222222;
fs->roots_of_unity = (void*)0x33333333;

printf("fs->expanded_roots_of_unity: %p\n", fs->expanded_roots_of_unity);
printf("fs->reverse_roots_of_unity: %p\n", fs->reverse_roots_of_unity);
printf("fs->roots_of_unity: %p\n", fs->roots_of_unity);

memset(fs, 0, sizeof(FFTSettings));
printf("----------------------\n");

printf("fs->expanded_roots_of_unity: %p\n", fs->expanded_roots_of_unity);
printf("fs->reverse_roots_of_unity: %p\n", fs->reverse_roots_of_unity);
printf("fs->roots_of_unity: %p\n", fs->roots_of_unity);
```

```
fs->expanded_roots_of_unity: 0x11111111
fs->reverse_roots_of_unity: 0x22222222
fs->roots_of_unity: 0x33333333
----------------------
fs->expanded_roots_of_unity: (nil)
fs->reverse_roots_of_unity: (nil)
fs->roots_of_unity: (nil)
```

That stack overflow question/answer isn't really relevant. They are passing the address of the pointer to memset, which isn't the same thing. What they are doing is definitely wrong.

As for the C# test, I think it's just a buggy test. It passed during the CI checks in this PR.

xrchz:
> The memset zeros them out (makes them null).

No I don't think this is how it works: the value zero is not the same as a null pointer. That's what the link I sent is supposed to be saying. To get a null pointer you need to use a literal/constant zero expression, but setting the value via `memset` won't achieve this afaiu. I think you would need to set them manually like `fs->expanded_roots_of_unity = NULL;`.

I agree that null pointers will be represented by 0 values on many platforms, but my understanding is that this is not portable. I could be wrong though!

jtraglia:
Oh I see what you're saying now! Sorry, I didn't understand before & I was focused on a different stack overflow answer (speaking of, my comment about it being wrong was incorrect). But yes, you're right. To be extra portable, your suggestion is good. I will make a PR with those changes. [This stack overflow answer](https://stackoverflow.com/a/2759875) was particularly helpful; the [seriously, have any actual machines really used nonzero null pointers](https://c-faq.com/null/machexamp.html) article was really interesting too.

PR #60, opened 2023-01-11
Convert Polynomial type to struct

jtraglia:
This PR converts the `Polynomial` from a typedef'd fixed-length array to a structure with a single member. This type is only used internally so we don't need to update any of the bindings.

PS: I will update the other types next. Wanting to do this work in small chunks.

In my opinion, a wrapped struct is safer & easier to read. See this link for details:
* https://stackoverflow.com/questions/4523497/typedef-fixed-length-array

Also, there's practically no performance difference. Benchmarked with go bindings.

Before:
```
$ ./bench.sh
goos: darwin
goarch: arm64
pkg: github.com/jtraglia/cgo-kzg-4844
BenchmarkBytesFromG1-20                         462160016               77.99 ns/op          192 B/op          2 allocs/op
BenchmarkBytesToG1-20                           89689077               401.0 ns/op           192 B/op          2 allocs/op
BenchmarkBytesToBlsField-20                     87684872               412.5 ns/op            64 B/op          2 allocs/op
BenchmarkComputeAggregateKzgProof-20                 730          49268169 ns/op             336 B/op          3 allocs/op
BenchmarkVerifyAggregateKzgProof-20                15490           2327210 ns/op             145 B/op          2 allocs/op
BenchmarkBlobToKzgCommitment-20                    18070           1996157 ns/op          131408 B/op          4 allocs/op
BenchmarkVerifyKzgProof-20                         41712            860376 ns/op             353 B/op          5 allocs/op
PASS
ok      github.com/jtraglia/cgo-kzg-4844        320.149s
```

After:
```
$ ./bench.sh
goos: darwin
goarch: arm64
pkg: github.com/jtraglia/cgo-kzg-4844
BenchmarkBytesFromG1-20                         463204014               78.04 ns/op          192 B/op          2 allocs/op
BenchmarkBytesToG1-20                           89752320               400.7 ns/op           192 B/op          2 allocs/op
BenchmarkBytesToBlsField-20                     87820915               410.6 ns/op            64 B/op          2 allocs/op
BenchmarkComputeAggregateKzgProof-20                 730          49246427 ns/op             336 B/op          3 allocs/op
BenchmarkVerifyAggregateKzgProof-20                15475           2327501 ns/op             145 B/op          2 allocs/op
BenchmarkBlobToKzgCommitment-20                    18154           1983811 ns/op          131408 B/op          4 allocs/op
BenchmarkVerifyKzgProof-20                         41888            860085 ns/op             353 B/op          5 allocs/op
PASS
ok      github.com/jtraglia/cgo-kzg-4844        320.093s
```

### DIFF
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -1310,7 +1310,7 @@ out:
 
 C_KZG_RET verify_aggregate_kzg_proof(bool *out,
                                      const Blob blobs[],
-                                     const KZGCommitment *expected_kzg_commitments,
+                                     const KZGCommitment expected_kzg_commitments[],
                                      size_t n,
                                      const KZGProof *kzg_aggregated_proof,
                                      const KZGSettings *s) {
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -790,8 +790,7 @@ static bool pairings_verify(const g1_t *a1, const g2_t *a2, const g1_t *b1, cons
     return blst_fp12_is_one(&gt_point);
 }
 
-
-typedef BLSFieldElement Polynomial[FIELD_ELEMENTS_PER_BLOB];
+typedef struct { BLSFieldElement elements[FIELD_ELEMENTS_PER_BLOB]; } Polynomial;
 
 void bytes_from_g1(uint8_t out[48], const g1_t *in) {
     blst_p1_compress(out, in);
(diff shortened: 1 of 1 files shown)


jtraglia:
@xrchz Could we change the repo settings to squash commits for PRs?

src/c_kzg_4844.c
```diff
@@ -1310,7 +1310,7 @@ C_KZG_RET compute_aggregate_kzg_proof(KZGProof *out,
 
 C_KZG_RET verify_aggregate_kzg_proof(bool *out,
                                      const Blob blobs[],
-                                     const KZGCommitment *expected_kzg_commitments,
+                                     const KZGCommitment expected_kzg_commitments[],
```
asn-d6, line 1313: Why was this reverted? Is it just to keep changes minimal?

Otherwise, [it's the same thing](https://c-faq.com/aryptr/aryptrparam.html). Ppersonally I prefer the pointer form and I'm a bit sad that c-kzg uses both forms.

jtraglia: I made that change before I decided to split things up into multiple PRs. Reverted it because it because that PR was focussed on polynomials and didn't want to create confusion. I'm pretty sure I made this same change in #62.

asn-d6: Makes sense!

src/c_kzg_4844.c
```diff
@@ -790,8 +790,7 @@ static bool pairings_verify(const g1_t *a1, const g2_t *a2, const g1_t *b1, cons
     return blst_fp12_is_one(&gt_point);
 }
 
-
-typedef BLSFieldElement Polynomial[FIELD_ELEMENTS_PER_BLOB];
+typedef struct { BLSFieldElement elements[FIELD_ELEMENTS_PER_BLOB]; } Polynomial;
```
asn-d6, line 793: I would maybe rename `elements` to `evals` or `evaluations` here since this is a polynomial in evaluation form.

I feel like `elements` is a vague term and `polynomial->element[i]` not actually meaningful in this context.

asn-d6 (commented):
All in all, looks great! Just a small nitpick on the naming.

src/c_kzg_4844.c
jtraglia: Gotcha. That works for me! Didn't really know what to name it.

PR #61, opened 2023-01-11
Convert Blob type to struct

jtraglia:
Convert the `Blob` type to a typedef'd anonymous struct. See #60 for rationale.

### DIFF
--- a/src/c_kzg_4844.h
+++ b/src/c_kzg_4844.h
@@ -47,7 +47,7 @@ typedef blst_fr fr_t;         /**< Internal Fr field element type */
 typedef g1_t KZGCommitment;
 typedef g1_t KZGProof;
 typedef fr_t BLSFieldElement;
-typedef uint8_t Blob[BYTES_PER_BLOB];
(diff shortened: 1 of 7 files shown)


src/c_kzg_4844.h
```diff
@@ -47,7 +47,7 @@ typedef blst_fr fr_t;         /**< Internal Fr field element type */
 typedef g1_t KZGCommitment;
 typedef g1_t KZGProof;
 typedef fr_t BLSFieldElement;
-typedef uint8_t Blob[BYTES_PER_BLOB];
+typedef struct { uint8_t data[BYTES_PER_BLOB]; } Blob;
```
asn-d6, line 50: Perhaps renaming `data` to `bytes` would be more meaningful in the context of a blob. After all, everything is "data" in computers.

asn-d6 (commented):
LGTM. Yet another tiny nit on naming.

src/c_kzg_4844.h
jtraglia: I agree. I like `bytes` better too. Thanks!

PR #62, opened 2023-01-11
Bytes-only interface

jtraglia:
This PR is an alternative to #37. See the notable changes below:

For C code:

* Replace `KZGCommitment`, `KZGProof`, and `BLSFieldElement` with bytes-only struct.
* In exported C funcs, add conversions to/from `g1_t`/`fr_t` point types.
* Remove exported conversions functions (like `bytes_to_g1`).
  * Make those functions static.
* Change type for field elements from `uint8_t z[BYTES_PER_FIELD_ELEMENT]` to `BLSFieldElement *z`.
  * Easier to read/understand IMO.
* Where applicable, use pointer style instead of array brackets.
  * For example, `void f(char a[])` -> `void f(char *a)`.
  * Not a strong reason for this besides consistency.
* For `verify_aggregate_kzg_proof`, use singular exit point during allocations.
* Wrap `BYTES_PER_BLOB` value in parentheses.
  * Ran into an issue with python bindings, where `n = 393216` (3 blobs).
  * `n = n / BYTES_PER_BLOB` would result the incorrect value: 3072.
  * This is because it was evaluated like `(n/32)*4096` instead of `n/(32*4096)`.

For Java bindings:

* Remove calls to `bytes_from_g1` and `bytes_to_g1`.
  * Also got to remove a bunch of allocations associated with that.
* Add `.PHONY` tags to rules so these get executed every time.
  * Ran into a cached build issue that misled me into thinking things worked.

For C# bindings:

* Remove calls to `bytes_from_g1` and `bytes_to_g1`.
* Remove `blob_to_kzg_commitment_wrap` and `compute_aggregate_kzg_proof_wrap` functions.
  * These are identical to their respective functions now.
* Remove `returns 0 on success` comment. Dup of docstring comment right above it.
* Fix some real minor spacing nits in the tests.

For Node.js bindings:

* Remove calls to `bytes_from_g1` and `bytes_to_g1`.
* Update expected exception in test.
  * It would expect a failure in `bytes_to_g1` which got removed.

For Python bindings:

* Change parameter/return types from `PyCapsule` to `PyBytes`.
  * The capsule bytes cannot be accessed which makes it difficult to actually use.
* Flatten `commitments` in `verify_aggregate_kzg_proof_wrap`.
  * Don't need to iterate of the sequence & extract anymore.
* Return value is allocated with `PyBytes_FromStringAndSize` now.
  * This is faster and simpler.
  * Can remove the `free_G1` method now.
* Remove `bytes_from_g1_wrap` as it's not needed anymore.
* Specify Python version in Makefile.
  * Spent about 5 minutes trying to figure out why python development headers couldn't be found.
  * On my system (and probably others), `python` doesn't point to anything.


jtraglia:
Definitely still needs some work! I'll do that tomorrow.

asn-d6:
In what way is this PR an alternative to #37? The idea with #37 is that client devs would never need to interact with any cryptographic objects and they would just pass bytes to the library. Does this PR have the same goals?

jtraglia:
Also, this is ready for review now. A few things to note:

* The python bindings compile, but I was unable to run the tests on my development system. Something about an unknown option `-Bsymbolic` during linking. I was uncertain about the change in [py_ecc_tests.py](https://github.com/ethereum/c-kzg-4844/pull/62/files#diff-565a335cf3223b4d9fec8fe48c188a0abdfb53928769fa5222004dd8bbdff84eL64-R64).
* I'm not confident that the C# tests are properly testing. I made some bad changes, ran the test, and it seems fine. It could be a cached build or something, not sure. Worth looking into. Also, if it were up to me, I'd remove the other two wrapped functions (`verify_aggregate_kzg_proof_wrap` and `verify_kzg_proof_wrap`) but I see the appeal to having `ret=0` be the success value for all of the functions.

jtraglia:
Regarding performance, benchmarks are just a hair slower but that makes sense because the conversion from G1 to bytes was done in setup and wasn't including in the benchmark time. In reality, it's probably the exact same overall.

Before:
```
$ ./bench.sh 
goos: darwin
goarch: arm64
pkg: github.com/jtraglia/cgo-kzg-4844
BenchmarkBytesFromG1-20                         463127452               78.02 ns/op          192 B/op          2 allocs/op
BenchmarkBytesToG1-20                           89462080               402.3 ns/op           192 B/op          2 allocs/op
BenchmarkBytesToBlsField-20                     87647565               411.7 ns/op            64 B/op          2 allocs/op
BenchmarkComputeAggregateKzgProof-20                 730          49276212 ns/op             336 B/op          3 allocs/op
BenchmarkVerifyAggregateKzgProof-20                15478           2330489 ns/op             145 B/op          2 allocs/op
BenchmarkBlobToKzgCommitment-20                    18087           1990456 ns/op          131408 B/op          4 allocs/op
BenchmarkVerifyKzgProof-20                         41899            858431 ns/op             353 B/op          5 allocs/op
```

After:
```
$ ./bench.sh
goos: darwin
goarch: arm64
pkg: github.com/jtraglia/cgo-kzg-4844
BenchmarkComputeAggregateKzgProof-20                 721          49401536 ns/op              48 B/op          1 allocs/op
BenchmarkVerifyAggregateKzgProof-20                15142           2381334 ns/op              49 B/op          2 allocs/op
BenchmarkBlobToKzgCommitment-20                    17972           2009970 ns/op          131120 B/op          2 allocs/op
BenchmarkVerifyKzgProof-20                         40689            891162 ns/op             161 B/op          5 allocs/op
```

(note the removed functions are no longer benchmarked)

bindings/python/ckzg.c
```diff
@@ -106,27 +95,30 @@ static PyObject* verify_aggregate_kzg_proof_wrap(PyObject *self, PyObject *args)
...
 
   Blob* blobs = (Blob*)PyBytes_AsString(b);
+  KZGProof *proof = (KZGProof*)PyBytes_AsString(p);
 
   for (Py_ssize_t i = 0; i < n; i++) {
     e = PySequence_GetItem(c, i);
-    if (!PyCapsule_IsValid(e, "G1")) {
+    if (!PyBytes_Check(e)) {
+      free(commitments);
+      return PyErr_Format(PyExc_ValueError, "expected KZGCommitment capsules");
```
xrchz, line 109: expected bytes?

jtraglia: Good catch. Thanks.

bindings/python/ckzg.c
```diff
@@ -39,21 +35,19 @@ static PyObject* blob_to_kzg_commitment_wrap(PyObject *self, PyObject *args) {
...
-  KZGCommitment *k = (KZGCommitment*)malloc(sizeof(KZGCommitment));
-
-  if (k == NULL) return PyErr_NoMemory();
+  PyObject *out = PyBytes_FromStringAndSize(NULL, BYTES_PER_COMMITMENT);
+  if (out == NULL) return PyErr_NoMemory();
 
+  Blob *blob = (Blob *)PyBytes_AsString(b);
+  KZGCommitment *k = (KZGCommitment *)PyBytes_AsString(out);
   if (blob_to_kzg_commitment(k, blob, PyCapsule_GetPointer(s, "KZGSettings")) != C_KZG_OK) {
-    free(k);
```
asn-d6, line 52: Do we need a `Py_DECREF(out);` here?

jtraglia:
> WRT "this is ready for review", there were no modifications required for the Rust bindings?

Hey George, this is associated with some chats in telegram. After marking it as ready to review, it was decided to merge the rust bindings (#4) into main. Then main got merged into this PR's branch. @pawanjay176 (who wrote the rust bindings) said that he will update them to the bytes interface today.

bindings/rust/src/bindings.rs
```diff
@@ -2,185 +2,181 @@
...
+impl DerefMut for Blob {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.bytes
+    }
+}
+
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+pub struct KZGProof {
+    bytes: [u8; 48usize],
```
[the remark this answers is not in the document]
jtraglia: Thanks. I agree. Fixed in all.

asn-d6 (approved):
LGTM! Thanks!

src/c_kzg_4844.c
```diff
@@ -1315,26 +1331,46 @@ C_KZG_RET verify_aggregate_kzg_proof(bool *out,
...
+    if (ret != C_KZG_OK) goto out;
+
+    commitments = calloc(n, sizeof(g1_t));
+    if (commitments == NULL) {
+        ret = C_KZG_MALLOC;
+        goto out;
+    }
+
+    polys = calloc(n, sizeof(Polynomial));
+    if (polys == NULL) {
```
xrchz, line 1348: We should also be checking `0 < n` here, since `calloc(0, _)` isn't portably specified (e.g., may return `NULL` on success). I can fix it later if you want though.

jtraglia: I agree. Fixed. It's a simple change, might as well do it here.

PR #71, opened 2023-01-17
Reorganize functions to match spec

jtraglia:
Sorry this is a lot. This PR re-organizes functions in the C file to *better* match [the spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/polynomial-commitments.md). No implementation changes.

* Attempt to group items that were previously spread out such as constants.
* Move all of the trusted setup code to the bottom of the file.

Note: when there's a paired "impl" function, he spec likes to define the `func` first then `func_impl`. We cannot do that because in C the impl function needs to be defined above the regular function.

See the screenshot below:

image

### DIFF
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -20,6 +20,139 @@
 #include <stdlib.h>
 #include <string.h>
 
+///////////////////////////////////////////////////////////////////////////////
+// Macros
+///////////////////////////////////////////////////////////////////////////////
+
+#define CHECK(cond)                                                                                                    \
+    if (!(cond)) return C_KZG_BADARGS
+
+///////////////////////////////////////////////////////////////////////////////
+// Types
+///////////////////////////////////////////////////////////////////////////////
+
+typedef struct { fr_t evals[FIELD_ELEMENTS_PER_BLOB]; } Polynomial;
+
+///////////////////////////////////////////////////////////////////////////////
+// Constants
+///////////////////////////////////////////////////////////////////////////////
+
+/** The G1 identity/infinity. */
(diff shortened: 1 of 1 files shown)


asn-d6:
> 

Patch looks sensible given an analysis with `git show --color-moved=default`.

Two minor things:
- You added `extern` in front of the sha256 funcs which was not there before. Why?
- `We cannot do that because in C the impl function needs to be defined above the regular function.` We can still do that if we add a *forward function declaration* of the `impl` function above so that the regular function knows what to expect.

jtraglia:
> You added `extern` in front of the sha256 funcs which was not there before. Why?

Oops good spot, I didn't intend to commit that. But when I moved those declarations to the top of the file, I wanted to explicitly clarify that those functions are implemented elsewhere. I'm pretty sure the `extern` is implicit anyway. Should I remove those changes?

> We can still do that if we add a _forward function declaration_ of the `impl` function above so that the regular function knows what to expect.

Yes, I didn't think about that. That's reasonable, I'll make those changes.

jtraglia:
Hmm that CI rust bindings failure was a bit weird.

```
process didn't exit successfully: `/home/runner/work/c-kzg-4844/c-kzg-4844/bindings/rust/target/release/deps/c_kzg-9a815dc584cbdc83` (signal: 11, SIGSEGV: invalid memory reference)
```

I was able to reproduce this issue on main, then about a minute later I was unable to reproduce the issue. Something to investigate. We're using rust stable, so it's probably not that. Maybe one of the dependencies?

jtraglia:
So that segfault happened again. It's a problem on main too. I've made an issue:
* https://github.com/ethereum/c-kzg-4844/issues/73

PR #72, opened 2023-01-17
Rename poly_from_blob to match spec

jtraglia:
I believe this function should be named [`blob_to_polynomial`](https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/polynomial-commitments.md#blob_to_polynomial) to match the spec.


asn-d6:
LGTM!

PR #77, opened 2023-01-19
Add go bindings

### LINKED ISSUE #55: Go bindings
So I think I finally figured out how to provide Go bindings. I've made a proof of concept here:

* https://github.com/jtraglia/cgo-kzg-4844

You can import this package and use the exported C-KZG-4844 functions.

Is this something we want in this repository?

jtraglia:
This PR adds go bindings. Includes tests & benchmarks. Uses test vectors from the rust bindings.

A necessary evil is that we need to add the blst headers (`blst.h` and `blst_aux.h`) to the repo.

Also, this requires the sha256 functions to be exposed in blst:
* https://github.com/supranational/blst/pull/146

Making it a draft while we wait. The tests are expected to fail until then.

Fixes #55.


bindings/go/main_test.go
```diff
@@ -122,12 +123,34 @@ func TestVerifyKZGProof(t *testing.T) {
...
+func GetRandFieldElement(seed int64) Bytes32 {
+	rand.Seed(seed)
+	bytes := make([]byte, 31)
+	_, err := rand.Read(bytes)
+	if err != nil {
+		panic("failed to get random field element")
+	}
+
+	var fieldElementBytes Bytes32
+	copy(fieldElementBytes[:], bytes)
```
xrchz, line 135: I'm guessing this leaves the last byte as 0? Could be worth commenting.

jtraglia: Yes, that's how it works. I'll add a comment 👍

bindings/go/blst_headers/blst.h
```diff
@@ -0,0 +1,483 @@
+/*
```
xrchz, line 1: Is this file generated?

jtraglia: Unfortunately no, it was manually copied. It will need to be updated manually if it changes.

xrchz: Maybe we could remove the mentions of swig to avoid confusion?

jtraglia: Maybe. How about a CI check that compares those files to the bindings in blst & fails if they don't perfectly match?

PR #80, opened 2023-01-23
Expose `compute_kzg_proof`

### LINKED ISSUE #34: Expose `compute_kzg_proof` in interface
I suggest adding `compute_kzg_proof` to the interface as it is required in order to compute valid inputs to the point evaluation precompile.

It will also be useful for testing purposes (easier to generate test cases for point evaluation precompile).

jtraglia:
This exposes the `compute_kzg_proof` in the way described here:

* https://github.com/ethereum/consensus-specs/pull/3219#pullrequestreview-1267235849

This PR will:

* Replace `BLSFieldElement` with `Bytes32` in many places.
  * These are not guaranteed to be valid field elements and what I did before was misleading. Sorry. 
* Add `compute_kzg_proof` to the readme.
  * Also sort the entries in a more logical order.
* Create `compute_kzg_proof_impl` and move bulk of work there.
  * This allows us to pass a polynomial to it from `compute_aggregate_kzg_proof`. 
* Use `f->bytes` for calls to `bytes_to_g1` & `bytes_to_bls_field`.
  * This is a lot cleaner in my opinion. 

Fixes #34.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,10 +3,11 @@
 This is a copy of [C-KZG](https://github.com/benjaminion/c-kzg) stripped-down to support the
 [Polynomial Commitments](https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/polynomial-commitments.md) API:
 
-- `compute_aggregate_kzg_proof`
-- `verify_aggregate_kzg_proof`
 - `blob_to_kzg_commitment`
+- `compute_kzg_proof`
+- `compute_aggregate_kzg_proof`
 - `verify_kzg_proof`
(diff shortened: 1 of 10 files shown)


jtraglia:
So I just learned that I changed some things I shouldn't have and it no longer matches the spec. Let me fix those things and then it will be ready for review. Sorry.

asn-d6:
Let's not merge this just yet. Actively workign on it.

jtraglia:
Okay I think this is ready now. Please review.

PR #83, opened 2023-01-24
Use downward gotos in g1_lincomb

jtraglia:
Wanted to clean up this function a little.


asn-d6:
Here is an approach to simplify error handling further here: https://github.com/asn-d6/c-kzg-4844/commit/56b412d242d4ce7b45dc144b18722a506af63c1e

Start from bad ret and only make it good if you get to the bottom.

asn-d6:
lgtm

PR #85, opened 2023-01-24
Upgrade blst & remove sha256 patch

### LINKED ISSUE #23: Move away from patching sha256.h?
The current approach to making the sha256 functions visible is to patch the header file. This works fine for bindings built via some script/make process, where patching is just a step amongst others to be executed. 

In the case of Nim however, c bindings are integrated with the regular compilation (basically, the c code is just "inlined"), so there's no good place to insert a patching step. 

For the time being I have forked this repo and added a change which [removes the patch and instead includes sha256.h](https://github.com/henridf/c-kzg-4844/commit/df76c552312f2b67291a0ac4c00760635e708ed8). Can similarly remove the patch here?

jtraglia:
This PR will do the following:

* Update the `blst` submodule to the latest commit, which exposes `blst_sha256` now.
* Remove the `ignore=dirty` submodule option because we no longer patch it.
* Remove patching from the Makefiles.
* Replace `hash` calls with `blst_sha256`.

Fixes #23.


xrchz:
very nice

Issue #84, opened 2023-01-24
Include subgroup checks when deserialising points


kevaundray:
Checking the code here : https://github.com/ethereum/c-kzg-4844/blob/main/src/c_kzg_4844.c#L498 It seems that we are not checking whether points are in the subgroup and are just checking whether points are in the group.

xrchz:
Is there a test we could add to one of the bindings that would expose this issue?

asn-d6:
Yes we should figure out if we can add a test.

asn-d6:
This is kinda what we are looking for https://github.com/ethereum/bls12-381-tests/blob/master/main.py#L599

asn-d6:
We added unittests for this in https://github.com/ethereum/c-kzg-4844/pull/102

xrchz:
So we can close this now right? I will close but reopen/message me if incorrect.

PR #86, opened 2023-01-25
Use Bytes48 for commitments/proofs

jtraglia:
This PR implements the changes defined here:

* https://github.com/ethereum/consensus-specs/pull/3224

Notable:

* `KZGProof`/`KZGCommitment` are considered trusted types. `Bytes48` is untrusted.
  * Technically they are the same value though, it's just a typedef.
* [deviation] Used `z_bytes` instead of `z` for consistency, this will be renamed later anyway.
* Renamed all of the static constants to be ALL_UPPER_CASE.
* Removed `bytes_to_g1` function (`bytes_from_g1` is still required).
* There's a duplicate check in `validate_kzg_g1`. Not sure which is better.

### DIFF
--- a/bindings/rust/src/lib.rs
+++ b/bindings/rust/src/lib.rs
@@ -185,14 +185,14 @@ impl KZGProof {
     pub fn verify_aggregate_kzg_proof(
         &self,
         blobs: &[Blob],
-        expected_commitments: &[Bytes48],
+        commitments_bytes: &[Bytes48],
         kzg_settings: &KZGSettings,
     ) -> Result<bool, Error> {
-        if blobs.len() != expected_commitments.len() {
+        if blobs.len() != commitments_bytes.len() {
             return Err(Error::MismatchLength(format!(
                 "There are {} blobs and {} commitments",
                 blobs.len(),
-                expected_commitments.len()
+                commitments_bytes.len()
             )));
         }
         let mut verified: MaybeUninit<bool> = MaybeUninit::uninit();
@@ -200,7 +200,7 @@ impl KZGProof {
             let res = verify_aggregate_kzg_proof(
                 verified.as_mut_ptr(),
                 blobs.as_ptr(),
-                expected_commitments.as_ptr(),
+                commitments_bytes.as_ptr(),
                 blobs.len(),
                 &self.to_bytes48(),
                 kzg_settings,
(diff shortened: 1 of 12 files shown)


bindings/rust/src/lib.rs
```diff
@@ -105,6 +107,33 @@ impl Drop for KZGSettings {
...
+                48,
+                bytes.len(),
+            )));
+        }
+        let mut new_bytes = [0; 48];
+        new_bytes.copy_from_slice(bytes);
+        Ok(Self { bytes: new_bytes })
+    }
+
+    pub fn to_proof(self) -> Result<KZGProof, Error> {
```
asn-d6, line 124: Are we using these helper functions? If not, maybe we can remove them? If yes, i think they should do the `validate_kzg_G1` logic since we consider KZGProof and KZGCommitment to be trusted types.

jtraglia: Good call. They weren't actually being used. Removed.

asn-d6:
LGTM!

asn-d6 (approved):

src/c_kzg_4844.c
```diff
@@ -664,6 +656,66 @@ static C_KZG_RET bytes_to_bls_field(fr_t *out, const Bytes32 *b) {
...
+ * @retval C_KZG_BADARGS Invalid input bytes
+ *
+ * @remark This function deviates from the spec because it returns (via an
+ *     output argument) the g1 point. This way is more efficient (faster) but
+ *     the function name is a bit misleading.
+ */
+static C_KZG_RET validate_kzg_g1(g1_t *out, const Bytes48 *b) {
+    /* Fast check without needing to uncompress */
+    if (memcmp(G1_POINT_AT_INFINITY.bytes, b->bytes, sizeof(b)) == 0)
+        return C_KZG_OK;
```
xrchz, line 674: Don't we still need to fill `out` in this case?

src/c_kzg_4844.h
```diff
@@ -45,10 +45,12 @@ typedef blst_p2 g2_t;         /**< Internal G2 group element type */
 typedef blst_fr fr_t;         /**< Internal Fr field element type */
 
 typedef struct { uint8_t bytes[32]; } Bytes32;
-typedef struct { uint8_t bytes[BYTES_PER_COMMITMENT]; } KZGCommitment;
-typedef struct { uint8_t bytes[BYTES_PER_PROOF]; } KZGProof;
+typedef struct { uint8_t bytes[48]; } Bytes48;
 typedef struct { uint8_t bytes[BYTES_PER_BLOB]; } Blob;
 
+typedef Bytes48 KZGCommitment;
```
xrchz, line 51: I probably missed it, but are these used for something?

xrchz:
Does this address #84 too?

asn-d6:
> Does this address #84 too?

It does yes. Altho we might want to leave that ticket open to figure out how to add a small subgroup unittest.

src/c_kzg_4844.c
asn-d6: Good catch!

src/c_kzg_4844.h
asn-d6: They are used as inputs/outputs in internal functions as in the spec. They signify a KZG proof or commitment which has been deserialized and normalized.

src/c_kzg_4844.c
jtraglia: Yes, thank you. I'm just going to remove that check, since there's a infinity check below.

PR #88, opened 2023-01-26
Move go.mod & go.sum to repo root

jtraglia:
Ugh, so in its current state, the go package cannot be imported.

```
# github.com/ethereum/c-kzg-4844/bindings/go
../../../go/pkg/mod/github.com/ethereum/c-kzg-4844/bindings/go@v0.0.0-20230126171313-363c7d7593b4/main.go:6:11: fatal error: 'c_kzg_4844.c' file not found
 #include "c_kzg_4844.c"
          ^~~~~~~~~~~~~~
1 error generated.
FAIL	fuzz [build failed]
```

It's because go packages cannot reference files outside of its package directory, and it adds `-I${SRCDIR}/../../src` to the include path. This is my fault for not remembering this limitation. This worked in my proof of concept because I copied the source files into a local directory.

https://github.com/ethereum/c-kzg-4844/blob/363c7d7593b4c844415f38735515a4dbdd6101bb/bindings/go/main.go#L3-L7

The best solution I can think of is to move the `go.mod` and `go.sum` files to the project root. Then the source files will be available. This is what `blst` does too, but it's a little cleaner because it has zero dependencies.

https://github.com/supranational/blst/blob/master/go.mod

This PR fixes the issue.


xrchz:
Maybe a note in the README about why those are there?

jtraglia:
Good idea. Will add. One minute.

jtraglia:
Put the note in the go bindings README. I think that's what you wanted. Makes sense to not pollute the root README.

PR #96, opened 2023-01-27
Add a basic unittest framework + a compute KZG proof test

asn-d6:
We had a unittest framework in the past, but not anymore.

As we get closer to audit, we will want to get more confidence on the internal functions as well, which cannot be directly reached from the bindings.

This PR adds a basic testing framework and a basic test for computing a KZG proof.

### DIFF
--- a/src/test_c_kzg_4844.c
+++ b/src/test_c_kzg_4844.c
@@ -1,0 +1,81 @@
+/* Tests for myfunctions.c, using TinyTest. */
+
+#define UNIT_TESTS
+
+#include "tinytest.h"
+#include "blst.h"
+
+#include "c_kzg_4844.h"
+#include <stdio.h>
+#include <string.h>
+#include <assert.h>
+
+KZGSettings s;
+
+void load_setup() {
+  FILE *fp;
+  C_KZG_RET ret;
+  fp = fopen("trusted_setup.txt", "rb");
+
+  ret = load_trusted_setup_file(&s, fp);
+  assert(ret == C_KZG_OK);
+
+  fclose(fp);
+}
+
+void get_32_rand_bytes(uint8_t *out) {
+  static uint64_t seed = 0;
+  seed++;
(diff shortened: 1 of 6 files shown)


src/test_c_kzg_4844.c
```diff
@@ -0,0 +1,81 @@
...
+
+  ret = load_trusted_setup_file(&s, fp);
+  assert(ret == C_KZG_OK);
+
+  fclose(fp);
+}
+
+void get_32_rand_bytes(uint8_t *out) {
+  static uint64_t seed = 0;
+  seed++;
```
xrchz, line 28: what is the point of this?

asn-d6: I added a few helper functions to build blobs and field elements to be used by the unittests.

That's the function that produces random data from a (built-in) seed to generate those objects.

xrchz: I mean why do we always set `seed` to `1`, or do I misunderstand what this is doing?

xrchz: I was confused thinking we always are setting `seed` to `1` but now I see the `static`.

src/test_c_kzg_4844.c
```diff
@@ -0,0 +1,81 @@
...
+#include <stdio.h>
+#include <string.h>
+#include <assert.h>
+
+KZGSettings s;
+
+void load_setup() {
+  FILE *fp;
+  C_KZG_RET ret;
+  fp = fopen("trusted_setup.txt", "rb");
```
xrchz, line 18: why `b`? This should be opened as text

jtraglia: It was just a copy/paste of the code in this comment. Yeah the "b" should be removed.

* https://github.com/ConsenSys/teku/issues/6622#issuecomment-1356376309

xrchz: OK I'll do it

PR #99, opened 2023-01-30
Make get_rand_blob return a full blob

jtraglia:
These memsets were required because only the first 128 field elements in the blob were being populated. The remaining field elements had "random" and were most likely were invalid. Do all 4096 (FIELD_ELEMENTS_PER_BLOB).


asn-d6:
Dang. Good find.

LGTM!

PR #100, opened 2023-01-30
Add some blob_to_kzg_commitment tests

jtraglia:
This tests a valid field element & an invalid field element.


jtraglia:
Ooo those are good test names. I'll rename these tests to closer match those.

PR #104, opened 2023-01-31
Clean up the header file

jtraglia:
Instead of building `test_c_kzg_4844.o`, build the test binary with `c_kzg_4844.c` which makes everything accessible. No need to export internal functions in the header. Also, move `Polynomial` back to the C file and `FIAT_SHAMIR_PROTOCOL_DOMAIN` to the C file since it's only used internally. Also, make everything that was `static` -> `STATIC` so it can be tested.

### DIFF
--- a/.github/workflows/ckzg-test.yml
+++ b/.github/workflows/ckzg-test.yml
@@ -14,8 +14,12 @@ jobs:
       - uses: actions/checkout@v3
         with:
           submodules: recursive
-      - name: Test
+      - name: Build
         run: |
           cd src
           make blst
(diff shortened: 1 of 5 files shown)


jtraglia:
Huh for some reason this works on my local setup but not in CI. Trying to figure it out.

jtraglia:
Very weird. Not worth the trouble right now. Maybe I'll figure it out later.

xrchz:
Oh - this was brilliant - would love to see it working if possible :) I might have a look some time.

PR #106, opened 2023-01-31
Generate coverage report

jtraglia:
* Generate coverage report for the tests.
* There is a generated html report available via CI artifact.
* For now, this doesn't fail if it's not 100% covered.
* You should be able to run `make test_cov` for a detailed view.

### DIFF
--- a/.github/workflows/ckzg-test.yml
+++ b/.github/workflows/ckzg-test.yml
@@ -34,8 +34,3 @@ jobs:
         with:
           name: coverage
           path: src/coverage.html
-      - name: Check coverage
-        run: |
-          cd src
-          make test_cov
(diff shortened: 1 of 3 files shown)


jtraglia:
It's not very obvious, but there's an artifact (`coverage.html`) that gets uploaded. Find it on this page:

* https://github.com/ethereum/c-kzg-4844/actions/runs/4059028996

jtraglia:
@asn-d6 what do you think about this?

asn-d6:
> 

Really cool!

jtraglia:
Also, it's showing that there isn't a failing case for the `blst_p1_on_curve` check:

image

PR #107, opened 2023-02-01
Remove duplicate p1_on_curve check

jtraglia:
[asn edit:] We probably don't need `blst_p1_on_curve()` since we are checking `blst_p1_uncompress()`'s retval. 

In reference to the comment below, this PR removes this check.

* https://github.com/ethereum/c-kzg-4844/pull/106#issuecomment-1411237332

Also, there's a test for this here:

https://github.com/ethereum/c-kzg-4844/blob/471245001daf2c3f4990351e600b0cb2c4d2a3ac/src/test_c_kzg_4844.c#L232-L240

### DIFF
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -646,14 +646,13 @@ static C_KZG_RET bytes_to_bls_field(fr_t *out, const Bytes32 *b) {
 static C_KZG_RET validate_kzg_g1(g1_t *out, const Bytes48 *b) {
     /* Convert the bytes to a p1 point */
     blst_p1_affine p1_affine;
+    /* The uncompress routine also checks that the point is on the curve */
     if (blst_p1_uncompress(&p1_affine, b->bytes) != BLST_SUCCESS)
         return C_KZG_BADARGS;
     blst_p1_from_affine(out, &p1_affine);
 
     /* The point at infinity is accepted! */
     if (blst_p1_is_inf(out)) return C_KZG_OK;
-    /* The point must be on the curve */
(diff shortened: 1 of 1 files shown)


asn-d6:
This seems potentially right, but let's look at this more carefully before merging.

jtraglia:
I agree. Also worth sharing, this is the call order and the point on curve check:

* [`blst_p1_uncompress`](https://github.com/supranational/blst/blob/ca03e11a3ff24d818ae390a1e7f435f15bf72aee/src/e1.c#L293)
* [`POINTonE1_Uncompress_Z`](https://github.com/supranational/blst/blob/ca03e11a3ff24d818ae390a1e7f435f15bf72aee/src/e1.c#L261)
* [`POINTonE1_Uncompress_BE`](https://github.com/supranational/blst/blob/ca03e11a3ff24d818ae390a1e7f435f15bf72aee/src/e1.c#L236)


```c
if (!sqrt_fp(ret.Y, ret.Y))
    return (limb_t)0 - BLST_POINT_NOT_ON_CURVE;
```

xrchz:
Is the thought that `blst_p1_from_affine` can't produce a point not on the curve?

jtraglia:
I think it's that `blst_p1_uncompress` will catch this first. @asn-d6 have you had a chance to think about this?

asn-d6:
Let's leave this open a bit more. I would like to look at this with @asanso at some point.

asanso:
@asn-d6  @jtraglia  lgtm I agree that extra check can be safely removed

PR #109, opened 2023-02-01
Simplify log_2_byte & add test

jtraglia:
In my opinion, `log_2_byte` was a little over complicated. It's easy/faster/simpler to make this a few if conditions.

Also, remove todos which are only related to loading the trusted setup.


xrchz:
Is there any significant speed difference between these? Just wondering if the test and implementation should be swapped.

jtraglia:
So I did benchmark it and found it to be almost the exact same. I'm guessing the compiler probably generates the same/similar code for both implementations. I think the one that gets used should be the simpler one.

PR #110, opened 2023-02-01
Add test for compute_powers

jtraglia:
Adds a simple test case for compute_powers to ensure it's working properly.

### DIFF
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -881,10 +881,18 @@ static void poly_lincomb(Polynomial *out, const Polynomial *vectors, const fr_t 
  * @param[in]  n   The number of powers to compute
  */
 STATIC void compute_powers(fr_t *out, fr_t *x, uint64_t n) {
-    fr_t current_power = FR_ONE;
-    for (uint64_t i = 0; i < n; i++) {
-        out[i] = current_power;
-        blst_fr_mul(&current_power, &current_power, x);
+    /* If n is 0, bail without touching out */
+    if (n == 0) return;
+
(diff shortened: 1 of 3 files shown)


jtraglia:
Also, it seems that `compute_powers` is computing one extra power than it needs to.

https://github.com/ethereum/c-kzg-4844/blob/1e7353c435face227c4037b929db21373013c3ba/src/c_kzg_4844.c#L883-L889

It assigns the power, then computes the next power. The last one won't be assigned to anything.

jtraglia:
Okay, I made some changes to `compute_powers` that I think will speed things up.

@asn-d6 what do you think about this? I think the spec could be improved here.

jtraglia:
Hmm something's a little wonky with node.js. Passes all of the tests but fails at the end.

```
Test Suites: 1 passed, 1 total
Tests:       9 passed, 9 total
Snapshots:   0 total
Time:        7.379 s
Ran all test suites.
free(): invalid next size (fast)
Aborted (core dumped)
error Command failed with exit code 134.
```

Will investigate.

asn-d6:
> Okay, I made some changes to `compute_powers` that I think will speed things up.
> 
> @asn-d6 what do you think about this? I think the spec could be improved here.

Hmm, so one field multiplication is about 40 nanoseconds. I think it doesn't make sense to optimize for this kind of gain. I would suggest we stick with the spec logic given that it's simpler and has less special case checks.

jtraglia:
Okay sounds good. Will revert.

asn-d6:
LGTM!

PR #117, opened 2023-02-06
Fix python bindings on macOS

### LINKED ISSUE #81: Remove "-Bsymbolic" compilation requirement from python bindings
Currently the python bindings do not compile on MacOS. It seems like the "-Bsymbolic" compiler option is not available, and without it it's reporting lots of missing python symbols.

Tagging @hwwhww

jtraglia:
This works on Linux & macOS. Ran into issues removing the `-Bsymbolic` flag and linking the Python library.

I was able to get things working by making my own `distutils` setup file.
 
Fixes #81.

### DIFF
--- a/.github/workflows/python-bindings-test.yml
+++ b/.github/workflows/python-bindings-test.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Build Python bindings
         run: |
           cd bindings/python
-          make ckzg.so
+          make install
       - name: Test
         run: |
           cd bindings/python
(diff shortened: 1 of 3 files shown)


jtraglia:
@hwwhww could you verify this works for you?

hwwhww (approved):
@jtraglia Yes! Thank you so much!!! 😭

Finally saw:
```
...
python3 tests.py
tests passed
python3 py_ecc_tests.py
comparison to py_ecc passed
```

PR #119, opened 2023-02-07
Cleanup the Makefile

jtraglia:
Okay, so the main purpose of this PR was to cleanup the Makefile but it evolved a bit:

* Split the Makefile into two sections (configuration options & rules).
* Do my best to simplify all of the Makefile rules.
* All lines are fewer than 80 chars in width.
* Added `test_c_kzg_4844_cov` to the clean rule.
* Remove the `lib` rule that was only used for node.js bindings.
* Don't echo the commands being run (personal preference).
* Rid ourselves of the `UNIT_TESTS` macro & section of exported test functions.
  * We no longer need to differentiate between `STATIC` and `static` now too.
* Add `-Wextra -Werror` to the compiler flags.
  * I noticed a warning for this while fixing the python bindings for macOS.
  * Fix a few `sign-compare` warnings, like this:

```
../../src/c_kzg_4844.c:825:27: warning: comparison of integers of different signs: 'int' and 'const uint64_t' (aka 'const unsigned long long') [-Wsign-compare]
        for (int i = 0; i < len; i++) {
                        ~ ^ ~~~
1 warning generated.
```

PS: I tried this before but ran into issues on Linux. The issue was the blst library needed to be specified after the source files. See this link for a little more information. What a dumb issue 🤦
* https://stackoverflow.com/a/2487723

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 test_c_kzg_4844
+test_c_kzg_4844_cov
 coverage.html
 *.profraw
 *.profdata
--- a/bindings/node.js/Makefile
+++ b/bindings/node.js/Makefile
@@ -9,7 +9,7 @@ clean:
 	rm -f *.o
 
 build: kzg.cxx kzg.ts package.json binding.gyp Makefile
-	cd ../../src; make lib
+	cd ../../src && make c_kzg_4844.o && cp c_kzg_4844.o ../bindings/node.js
 	yarn install
 	yarn node-gyp rebuild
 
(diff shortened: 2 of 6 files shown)


jtraglia:
@asn-d6 @xrchz how does this look to you guys?

jtraglia:
Hey George. I agree that including a `.c` file isn't very common (nor recommended), but it an option and it does simplify things quite a lot. This is also just for the testing suite and allows to us to keep the header as is. I'm not a fan of having to put function definitions for all of the static functions in the header.

asn-d6:
Sounds good. No strong opinion here. Let's proceed if @xrchz  is also OK with it.

xrchz:
I like this approach. And c-kzg-4844 is going to stay a small single file :)

PR #120, opened 2023-02-07
Add benchmark for g1Lincomb in go bindings

jtraglia:
This PR starts to add some internal (private) functions to the go bindings. These could be used for benchmarking.


jtraglia:
Eh, I don't really think this is that useful anymore, especially because we have to convert types. Closing.

PR #123, opened 2023-02-08
Interface changes to free the blobs

asn-d6:
This PR implements the specs changes from this PR:
* https://github.com/ethereum/consensus-specs/pull/3236

### DIFF
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -21,6 +21,7 @@
  */
 #include "c_kzg_4844.h"
 
+#include <assert.h>
 #include <inttypes.h>
 #include <stdlib.h>
 #include <string.h>
@@ -38,8 +39,9 @@
 
 // clang-format off
 
-/** The Fiat-Shamir protocol domain. */
+/** The Fiat-Shamir protocol domains. */
 static const char *FIAT_SHAMIR_PROTOCOL_DOMAIN = "FSBLOBVERIFY_V1_";
+static const char *RANDOM_CHALLENGE_KZG_BATCH_DOMAIN = "RCKZGBATCH___V1_";
 
 /** Deserialized form of the G1 identity/infinity point. */
 static const g1_t G1_IDENTITY = {
@@ -650,10 +652,8 @@ static C_KZG_RET validate_kzg_g1(g1_t *out, const Bytes48 *b) {
 
     /* The point at infinity is accepted! */
     if (blst_p1_is_inf(out)) return C_KZG_OK;
-
     /* The point must be on the curve */
     if (!blst_p1_on_curve(out)) return C_KZG_BADARGS;
-
     /* The point must be on the right subgroup */
     if (!blst_p1_in_g1(out)) return C_KZG_BADARGS;
 
@@ -698,37 +698,24 @@ static C_KZG_RET blob_to_polynomial(Polynomial *p, const Blob *blob) {
     return C_KZG_OK;
 }
 
-/* Forward function definition */
-static void compute_powers(fr_t *out, fr_t *x, uint64_t n);
+/* Input size to the Fiat-Shamir challenge computation */
+#define CHALLENGE_INPUT_SIZE 32 + BYTES_PER_BLOB + 48
 
 /**
- * Return the Fiat-Shamir challenges required by the rest of the protocol.
+ * Return the Fiat-Shamir challenge required to verify `blob` and `commitment`.
  *
  * @remark This function should compute challenges even if `n==0`.
  *
  * @param[out] eval_challenge_out The evaluation challenge
- * @param[out] r_powers_out       The powers of r, where r is a randomly
- *                                generated scalar
- * @param[in]  polys              The array of polynomials
- * @param[in]  comms              The array of commitments
+ * @param[in]  blob               A blob
+ * @param[in]  commitment         A commitment
  * @param[in]  n                  The number of polynomials and commitments
  */
-static C_KZG_RET compute_challenges(
-    fr_t *eval_challenge_out,
-    fr_t *r_powers_out,
-    const Polynomial *polys,
-    const g1_t *comms,
-    uint64_t n
+static void compute_challenge(
+    fr_t *eval_challenge_out, const Blob *blob, const g1_t *commitment
 ) {
-    C_KZG_RET ret;
-    size_t i;
-    uint64_t j;
-    uint8_t *bytes = NULL;
-
-    // len(FIAT_SHAMIR_PROTOCOL_DOMAIN) + 8 + 8 + n blobs + n commitments
-    size_t input_size = 32 + (n * BYTES_PER_BLOB) + (n * 48);
-    ret = c_kzg_malloc((void **)&bytes, input_size);
-    if (ret != C_KZG_OK) goto out;
+    Bytes32 eval_challenge;
+    uint8_t bytes[CHALLENGE_INPUT_SIZE];
 
     /* Pointer tracking `bytes` for writing on top of it */
     uint8_t *offset = bytes;
@@ -736,52 +723,27 @@ static C_KZG_RET compute_challenges(
     /* Copy domain separator */
     memcpy(offset, FIAT_SHAMIR_PROTOCOL_DOMAIN, 16);
     offset += 16;
+    bytes_from_uint64(
+        offset, 0
+    ); /* need to fill 16 bytes with the degree in little-endian */
+    offset += 8;
     bytes_from_uint64(offset, FIELD_ELEMENTS_PER_BLOB);
     offset += 8;
-    bytes_from_uint64(offset, n);
-    offset += 8;
 
-    /* Copy polynomials */
-    for (i = 0; i < n; i++) {
-        for (j = 0; j < FIELD_ELEMENTS_PER_BLOB; j++) {
-            bytes_from_bls_field((Bytes32 *)offset, &polys[i].evals[j]);
-            offset += BYTES_PER_FIELD_ELEMENT;
-        }
-    }
+    /* Copy blob */
+    memcpy(offset, blob->bytes, BYTES_PER_BLOB);
+    offset += BYTES_PER_BLOB;
 
-    /* Copy commitments */
-    for (i = 0; i < n; i++) {
-        bytes_from_g1((Bytes48 *)offset, &comms[i]);
-        offset += BYTES_PER_COMMITMENT;
-    }
+    /* Copy commitment */
+    bytes_from_g1((Bytes48 *)offset, commitment);
+    offset += BYTES_PER_COMMITMENT;
 
-    /* Now let's create challenges! */
-    uint8_t hashed_data[32] = {0};
-    blst_sha256(hashed_data, bytes, input_size);
+    /* Make sure we wrote the entire buffer */
+    assert(offset == bytes + CHALLENGE_INPUT_SIZE);
 
-    /* We will use hash_input in the computation of both challenges */
-    uint8_t hash_input[33];
-
-    /* Compute r */
-    Bytes32 r_bytes;
-    memcpy(hash_input, hashed_data, 32);
-    hash_input[32] = 0x0;
-    blst_sha256(r_bytes.bytes, hash_input, 33);
-
-    /* Compute r_powers */
-    fr_t r;
-    hash_to_bls_field(&r, &r_bytes);
-    compute_powers(r_powers_out, &r, n);
-
-    /* Compute eval_challenge */
-    Bytes32 eval_challenge;
-    hash_input[32] = 0x1;
-    blst_sha256(eval_challenge.bytes, hash_input, 33);
+    /* Now let's create the challenge! */
+    blst_sha256(eval_challenge.bytes, bytes, CHALLENGE_INPUT_SIZE);
     hash_to_bls_field(eval_challenge_out, &eval_challenge);
-
-out:
-    free(bytes);
-    return ret;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


asn-d6:
Here are benchmarks from my crappy laptop:

```
❯ go test -bench=Benchmark 
goos: linux                                                                       
goarch: amd64
pkg: github.com/ethereum/c-kzg-4844/bindings/go
cpu: Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
Benchmark/BlobToKZGCommitment-8         	      20	  58965077 ns/op
Benchmark/ComputeKZGProof-8             	      20	  62236844 ns/op
Benchmark/VerifyKZGProof-8              	     856	   1541587 ns/op
Benchmark/ComputeBlobKZGProof-8         	       9	 123533698 ns/op
Benchmark/VerifyBlobKZGProof-8          	     380	   3000938 ns/op
Benchmark/VerifyBlobKZGProofMulti(blobs=1)-8         	     426	   3189670 ns/op
Benchmark/VerifyBlobKZGProofMulti(blobs=2)-8         	     250	   5062706 ns/op
Benchmark/VerifyBlobKZGProofMulti(blobs=4)-8         	     127	   8934038 ns/op
Benchmark/VerifyBlobKZGProofMulti(blobs=8)-8         	      72	  16760772 ns/op
Benchmark/VerifyBlobKZGProofMulti(blobs=16)-8        	      36	  32171933 ns/op
Benchmark/VerifyBlobKZGProofMulti(blobs=32)-8        	      21	  55817548 ns/op
Benchmark/VerifyBlobKZGProofMulti(blobs=64)-8        	      10	 119034498 ns/op
```

jtraglia:
> Need to run this on another laptop as well (since @jtraglia's laptop was 4 times faster than mine), and need to also look deeper into the code to see what's happening and why it's so slow.

I'm pretty sure it was only twice as fast 😄 ... Here are those benchmarks from my laptop:

```
goos: darwin
goarch: arm64
pkg: github.com/ethereum/c-kzg-4844/bindings/go
Benchmark/BlobToKZGCommitment-8         	      28	  41118077 ns/op
Benchmark/ComputeKZGProof-8             	      26	  44418431 ns/op
Benchmark/VerifyKZGProof-8              	    1234	    979579 ns/op
Benchmark/ComputeBlobKZGProof-8         	      13	  85605051 ns/op
Benchmark/VerifyBlobKZGProof-8          	     717	   1673759 ns/op
Benchmark/VerifyBlobKZGProofMulti(blobs=1)-8         	     684	   1774757 ns/op
Benchmark/VerifyBlobKZGProofMulti(blobs=2)-8         	     427	   2844018 ns/op
Benchmark/VerifyBlobKZGProofMulti(blobs=4)-8         	     235	   4936515 ns/op
Benchmark/VerifyBlobKZGProofMulti(blobs=8)-8         	     134	   8868204 ns/op
Benchmark/VerifyBlobKZGProofMulti(blobs=16)-8        	      72	  16470036 ns/op
Benchmark/VerifyBlobKZGProofMulti(blobs=32)-8        	      37	  31140634 ns/op
Benchmark/VerifyBlobKZGProofMulti(blobs=64)-8        	      19	  61290555 ns/op
```

Which reminds me of something interesting, I mentioned in private that these benchmarks are lot (almost 2x) faster on macOS than on Linux (via Asahi Linux), same computer. Originally, I thought it was a golang optimization, but I ran the Java benchmarks on both operating systems and it had the same problem. So now I'm thinking it's either some low-level operating system thing or BLST has some optimizations enabled for Darwin that aren't for Linux. I checked BLST but didn't see anything that would cause this. There are preprocessor differences, but they include the same files. Not sure what's going on but something to investigate later.

jtraglia:
For reference, this is on a somewhat modern x86_64 server CPU:

```
CC=clang go test -bench=Benchmark
goos: linux
goarch: amd64
pkg: github.com/ethereum/c-kzg-4844/bindings/go
cpu: AMD Ryzen 9 5950X 16-Core Processor
Benchmark/BlobToKZGCommitment-32         	      14	  76162702 ns/op
Benchmark/ComputeKZGProof-32             	      14	  79659143 ns/op
Benchmark/VerifyKZGProof-32              	     625	   1850039 ns/op
Benchmark/ComputeBlobKZGProof-32         	       7	 159006511 ns/op
Benchmark/VerifyBlobKZGProof-32          	     321	   3671053 ns/op
Benchmark/VerifyBlobKZGProofMulti(blobs=1)-32         	     325	   3671062 ns/op
Benchmark/VerifyBlobKZGProofMulti(blobs=2)-32         	     201	   5925584 ns/op
Benchmark/VerifyBlobKZGProofMulti(blobs=4)-32         	     100	  11034293 ns/op
Benchmark/VerifyBlobKZGProofMulti(blobs=8)-32         	      62	  20514069 ns/op
Benchmark/VerifyBlobKZGProofMulti(blobs=16)-32        	      31	  36407820 ns/op
Benchmark/VerifyBlobKZGProofMulti(blobs=32)-32        	      16	  67926201 ns/op
Benchmark/VerifyBlobKZGProofMulti(blobs=64)-32        	       8	 133235901 ns/op
```

Which is the slowest of the three. It's somewhat faster if I use `clang` as the compiler.

xrchz:
Another run for reference
```
goos: linux
goarch: amd64
pkg: github.com/ethereum/c-kzg-4844/bindings/go
cpu: 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz
Benchmark/BlobToKZGCommitment-16         	      27	  41626739 ns/op
Benchmark/ComputeKZGProof-16             	      26	  44164153 ns/op
Benchmark/VerifyKZGProof-16              	    1222	    993900 ns/op
Benchmark/ComputeBlobKZGProof-16         	      13	  84093680 ns/op
Benchmark/VerifyBlobKZGProof-16          	     588	   2033791 ns/op
Benchmark/VerifyBlobKZGProofMulti(blobs=1)-16         	     564	   2120081 ns/op
Benchmark/VerifyBlobKZGProofMulti(blobs=2)-16         	     342	   3554287 ns/op
Benchmark/VerifyBlobKZGProofMulti(blobs=4)-16         	     187	   6289255 ns/op
Benchmark/VerifyBlobKZGProofMulti(blobs=8)-16         	     100	  11464281 ns/op
Benchmark/VerifyBlobKZGProofMulti(blobs=16)-16        	      54	  21864610 ns/op
Benchmark/VerifyBlobKZGProofMulti(blobs=32)-16        	      27	  41963752 ns/op
Benchmark/VerifyBlobKZGProofMulti(blobs=64)-16        	      14	  81968819 ns/op
PASS
ok  	github.com/ethereum/c-kzg-4844/bindings/go	27.530s
```

potuz:
FWIW, here are some on a 12th gen i7
```
goos: linux
goarch: amd64
pkg: github.com/ethereum/c-kzg-4844/bindings/go
cpu: 12th Gen Intel(R) Core(TM) i7-1260P
Benchmark/BlobToKZGCommitment-16         	      31	  37319180 ns/op
Benchmark/ComputeKZGProof-16             	      30	  38965704 ns/op
Benchmark/VerifyKZGProof-16              	    1471	    823844 ns/op
Benchmark/ComputeAggregateKZGProof(blobs=1)-16         	      15	  74691875 ns/op
Benchmark/ComputeAggregateKZGProof(blobs=2)-16         	       9	 111621830 ns/op
Benchmark/ComputeAggregateKZGProof(blobs=4)-16         	       6	 185623088 ns/op
Benchmark/ComputeAggregateKZGProof(blobs=8)-16         	       4	 339695622 ns/op
Benchmark/ComputeAggregateKZGProof(blobs=16)-16        	       2	 652552524 ns/op
Benchmark/ComputeAggregateKZGProof(blobs=32)-16        	       1	1232714504 ns/op
Benchmark/ComputeAggregateKZGProof(blobs=64)-16        	       1	2552187462 ns/op
Benchmark/VerifyAggregateKZGProof(blobs=1)-16          	     630	   1824820 ns/op
Benchmark/VerifyAggregateKZGProof(blobs=2)-16          	     462	   2636964 ns/op
Benchmark/VerifyAggregateKZGProof(blobs=4)-16          	     312	   3982597 ns/op
Benchmark/VerifyAggregateKZGProof(blobs=8)-16          	     181	   6620619 ns/op
Benchmark/VerifyAggregateKZGProof(blobs=16)-16         	     100	  11829743 ns/op
Benchmark/VerifyAggregateKZGProof(blobs=32)-16         	      55	  22202064 ns/op
Benchmark/VerifyAggregateKZGProof(blobs=64)-16         	      27	  43071931 ns/op
PASS
ok  	github.com/ethereum/c-kzg-4844/bindings/go	32.685s
```

jtraglia:
The bindings have been updated to support the new interface. Additionally, I updated the benchmarks for Go/Rust/Java to reflect these changes. Didn't really do much with the tests; updated what I could but removed/commented out lots of them. I would like to focus on tests in a future PR, as this one is already quite big. If it compiles it should be fine, right?

asn-d6 (commented):
Bindings changes look reasonable but it's hard to say without tests or clients using them.

bindings/csharp/ckzg.c
```diff
@@ -23,19 +23,3 @@ void free_trusted_setup_wrap(KZGSettings *s) {
   free_trusted_setup(s);
   free(s);
 }
-
-int verify_aggregate_kzg_proof_wrap(const Blob *blobs, const Bytes48 *commitments_bytes, size_t n, const Bytes48 *aggregated_proof_bytes, const KZGSettings *s) {
-  bool b;
-  C_KZG_RET ret = verify_aggregate_kzg_proof(&b, blobs, commitments_bytes, n, aggregated_proof_bytes, s);
```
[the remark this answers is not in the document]
jtraglia: These were wrappers that changed the interface. See that they don't take a boolean as the first parameter. When removed, they will be able to directly call the C implementation, which doesn't require a wrapper. Personally, I think this is the right approach.

bindings/node.js/kzg.cxx
```diff
@@ -152,190 +152,185 @@ Napi::Value BlobToKzgCommitment(const Napi::CallbackInfo& info) {
...
     return throw_invalid_arguments_count(expected_argument_count, argument_count, env);
   }
 
-  auto commitment_bytes = extract_byte_array_from_param(info, 0, "commitmentBytes");
-  auto z_bytes = extract_byte_array_from_param(info, 1, "zBytes");
-  auto y_bytes = extract_byte_array_from_param(info, 2, "yBytes");
-  auto proof_bytes = extract_byte_array_from_param(info, 3, "proofBytes");
+  auto blobs_param = info[0].As<Napi::Array>();
+  auto commitments_param = info[1].As<Napi::Array>();
+  auto proofs_param = info[1].As<Napi::Array>();
```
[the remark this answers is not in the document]
jtraglia: Yes, it should be. Good spot!

bindings/node.js/kzg.cxx
```diff
@@ -152,190 +152,185 @@ Napi::Value BlobToKzgCommitment(const Napi::CallbackInfo& info) {
...
+  auto commitments_count = commitments_param.Length();
+  auto proofs_count = proofs_param.Length();
+
+  if (blobs_count != commitments_count || blobs_count != proofs_count) {
+    Napi::Error::New(env, "verifyBlobKzgProofBatch requires equal number of blobs/commitments/proofs")
+      .ThrowAsJavaScriptException();
+    return env.Null();
+  }
+
+  auto blobs_bytes = extract_byte_array_from_param(info, 0, "blobs");
```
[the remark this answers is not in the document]
jtraglia: Nope, they're wrong too. Good catch.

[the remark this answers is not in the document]
jtraglia: Yeah I think we need tests...

src/c_kzg_4844.c
```diff
@@ -698,90 +698,52 @@ static C_KZG_RET blob_to_polynomial(Polynomial *p, const Blob *blob) {
     return C_KZG_OK;
 }
 
-/* Forward function definition */
-static void compute_powers(fr_t *out, fr_t *x, uint64_t n);
+/* Input size to the Fiat-Shamir challenge computation */
+#define CHALLENGE_INPUT_SIZE 32 + BYTES_PER_BLOB + 48
```
xrchz, line 702: Would `static const int` be better here? or even `enum`?

src/c_kzg_4844.c
```diff
@@ -698,90 +698,52 @@ static C_KZG_RET blob_to_polynomial(Polynomial *p, const Blob *blob) {
...
 
     /* Pointer tracking `bytes` for writing on top of it */
     uint8_t *offset = bytes;
 
     /* Copy domain separator */
     memcpy(offset, FIAT_SHAMIR_PROTOCOL_DOMAIN, 16);
     offset += 16;
-    bytes_from_uint64(offset, FIELD_ELEMENTS_PER_BLOB);
+    bytes_from_uint64(
+        offset, 0
```
xrchz, line 727: shouldn't the polynomial degree (`FIELD_ELEMENTS_PER_BLOB`) go here instead of `0`?

xrchz: oh I see you need extra zero padding...

src/c_kzg_4844.c
```diff
@@ -1218,170 +1153,330 @@ C_KZG_RET compute_kzg_proof_impl(
...
+    size_t n,
+    const KZGSettings *s
+) {
+    C_KZG_RET ret;
+    g1_t proof_lincomb, proof_z_lincomb, C_minus_y_lincomb, rhs_g1;
+    fr_t *r_powers = NULL;
+    g1_t *C_minus_y = NULL;
+    fr_t *r_times_z = NULL;
+
+    assert(n > 0);
```
xrchz, line 1339: should we return `C_KZG_BADARGS` here?

xrchz: I guess the idea is we control all the call sites, so really don't expect this

xrchz:
I'm happy with the changes to `src`. Haven't looked at the others yet.

src/c_kzg_4844.c
asn-d6: Hmm. I think it's standard (and reasonable) practice to have public constants as `#define`s. We do the same thing for `BYTES_PER_COMMITMENT` etc.

I don't see a clear benefit to making it `static const int`, and I'm not sure how we would use an `enum`.

asn-d6:
lgtm!

src/c_kzg_4844.c
xrchz: Some thoughts on the benefits can be found here https://stackoverflow.com/questions/1674032/static-const-vs-define-vs-enum. In this case I was thinking of it primarily because we're doing arithmetic (`+`) unlike `BYTES_PER_COMMITMENT` which is a bare integer literal. (`BYTES_PER_BLOB` however would face the same argument.) We count on the compiler to constant fold this arithmetic when using macros.

asn-d6: No strong opinion here. I will defer to @jtraglia .

xrchz: yeah to be clear I don't really have a strong opinion either.

jtraglia: No strong opinion either, but I'm leaning towards a static const. In my opinion it's safer. If you decide to stick with a macro, you should wrap it in parentheses in-case it's every used in a multiplication/division operation. This was a problem with `BYTES_PER_BLOB` at one point in time.

asn-d6: ACK! Fixed in https://github.com/ethereum/c-kzg-4844/pull/123/commits/53534347ec7d0d16f8be84aa0285916bcf7ac798 .

jtraglia (approved):
LGTM 💯

PR #129, opened 2023-02-10
Run static analysis in CI

jtraglia:
* Add `make analyze` makefile rule that runs clang's static analyzer.
* It will exit with an error if there are any findings.
* Add this command to the CI checks.
* Fix the three findings.

```
$ make analyze
c_kzg_4844.c:315:13: warning: Array access (from variable 'prod') results in a null pointer dereference [core.NullDereference]
    prod[0] = a[0];
    ~~~~    ^
c_kzg_4844.c:573:38: warning: The result of the right shift is undefined due to shifting by '32', which is greater or equal to the width of type 'uint32_t' [core.UndefinedBinaryOperatorResult]
        uint32_t r = reverse_bits(i) >> unused_bit_len;
                                     ^  ~~~~~~~~~~~~~~
c_kzg_4844.c:1452:16: warning: Dereference of null pointer (loaded from variable 'in') [core.NullDereference]
        *out = *in;
               ^~~
3 warnings generated.
make: *** [analyze] Error 1
```

Somewhat valid findings, all in trusted setup code though. Thanks clang 👍 

Modified the `c_kzg_malloc` function to fix one of them. I don't think we should return `C_KZG_OK` if the result is null. I think we should just return `C_KZG_BADARGS` if the requested amount is zero. Is this okay?


xrchz:
I think there may still be an open question of whether to use `c_kzg_malloc` at all, or to just remove it. Or use it everywhere. I think there are some places where zero size allocation does make sense, so if we were to use it everywhere this wouldn't work. But I personally think the places where `c_kzg_malloc` is used would be better served by `calloc` anyway.

I don't mind merging this now and deciding about it later. Thoughts?

jtraglia:
> open question of whether to use c_kzg_malloc at all, or to just remove it. Or use it everywhere.

Originally, I thought it would be a good idea to remove all of the memory helper functions, but after making those changes I felt that it was less readable and easier to make a mistake. I think we should keep `c_kzg_malloc`, use it everywhere, and make a new `c_kzg_calloc` function that returns `C_KZG_MALLOC` on failures too. But let's do that in another PR.

xrchz:
OK I'll wait to see what you do in a subsequent PR. Merging this for static analysis.

PR #130, opened 2023-02-11
Add c_kzg_calloc function

jtraglia:
* Adds new `c_kzg_calloc` func that returns a `C_KZG_RET` value.
* Replace all instances of `malloc` and `calloc` with helper funcs.
* Rename parameters to match the func its wrapping.
* Use `c_kzg_calloc` in `new_*_array` funcs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 test_c_kzg_4844
 test_c_kzg_4844_*
 coverage.html
+analysis-report/
 *.profraw
 *.profdata
 *.prof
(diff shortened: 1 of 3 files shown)


jtraglia:
Associated with #129.

jtraglia:
Part of the reason I wanted to make this change, it's easy to make mistakes like this before:

https://github.com/ethereum/c-kzg-4844/blob/44fe79fd424cf3f8fb46e7ca09ab2f8e613ca825/src/c_kzg_4844.c#L1315-L1326

Above, `ret` is initialized to `C_KZG_MALLOC` but overwritten when converting the proof to bytes. If there were a memory error, this would have return `C_KZG_OK` instead of an error.

jtraglia:
Oops. Still need to free the allocations in the tests. Not able to do it right now, please wait to merge.

PR #133, opened 2023-02-13
Disable compiler optimizations for tests

jtraglia:
* Disable optimizations for tests.
* Rename `test_cov` to `coverage`.
* Add two new tests.

Ran into an interesting problem with the tests. It appears that some of them were being optimized out. I added two new `c_kzg_*alloc` tests and they would fail unless I printed the results. Opened the test binary in Ghidra and found that it was just automatically failing without actually running the test. Also, it inlined all of the tests into main.

Screenshot 2023-02-13 at 10 50 06 AM

Disabled optimizations and it looked like:

Screenshot 2023-02-13 at 11 06 53 AM

Doing a string search, I found that it was only executing 24/32 of the tests:

Screenshot 2023-02-13 at 10 50 30 AM

Without optimizations, this properly showed 32 strings:

Screenshot 2023-02-13 at 10 53 50 AM

### DIFF
--- a/.github/workflows/ckzg-test.yml
+++ b/.github/workflows/ckzg-test.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Generate coverage report
         run: |
           cd src
-          make test_cov
+          make coverage
       - name: Save coverage report
         uses: actions/upload-artifact@v3
         with:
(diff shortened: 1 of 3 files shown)


xrchz:
nice find!

xrchz:
We assume the compiler optimisations don't introduce bugs this way...

PR #135, opened 2023-02-14
Add method to load trusted setup from a resource

fab-10:


bindings/java/src/main/java/ethereum/ckzg4844/CKZG4844JNI.java
```diff
@@ -106,6 +105,31 @@ public static int getBytesPerBlob() {
...
+  public static void loadTrustedSetupFromResource(String resource, Class clazz) {
+    InputStream is = clazz.getResourceAsStream(resource);
+    if (is == null) {
+      throw new IllegalArgumentException("Resource " + resource + " does not exist.");
+    }
+
+    try (is) {
+      Path jniWillLoadFrom = tempDir().resolve("kzg-trusted-setup.txt");
+      jniWillLoadFrom.toFile().deleteOnExit();
+      Files.copy(is, jniWillLoadFrom, StandardCopyOption.REPLACE_EXISTING);
```
jtraglia, line 125: This feels like there could be a race condition. Is it possible to create a unique temp directory each time `tempDir()` is called, instead of reusing the same one? I guess it could be a limitation of `deleteOnExit`.

bindings/java/src/main/java/ethereum/ckzg4844/CKZG4844JNI.java
```diff
@@ -168,4 +192,11 @@ public static native boolean verifyAggregateKzgProof(byte[] blobs, byte[] commit
   public static native boolean verifyKzgProof(byte[] commitment_bytes, byte[] z_bytes,
       byte[] y_bytes, byte[] proof_bytes);
 
+  private static synchronized Path tempDir() throws IOException {
+    if(tempDir == null) {
```
jtraglia, line 196: Minor nit, add a space before the opening parentheses.

jtraglia (commented):
Thanks, I like this 👍 just a minor nit/concern.

bindings/java/src/main/java/ethereum/ckzg4844/CKZG4844JNI.java
fab-10: testing on Linux I had no issues, in any case I have reverted the change, so now it just create a temp file in the default tmp dir, instead of trying to reuse the same dir of the lib

jtraglia: Thanks, personally I like this better 👍

jtraglia:
A bit of an issue with CI tests. There's a segfault in `freeTrustedSetup` for some reason.

fab-10:
> A bit of an issue with CI tests. There's a segfault in `freeTrustedSetup` for some reason.

any idea on which test it fails?

jtraglia:
Ah gotcha, makes sense. Thanks for fixing that. LGTM now :+1:

PR #137, opened 2023-02-16
Add more unit tests

dankrad:
More unit tests for c-kzg-4844


jtraglia:
Looking good, two things:

1. That test (which is unrelated to yours) is failing because of an ordering issue that we should probably fix. If you move your new tests after the `test_blob_to_kzg_commitment` section they should all pass. I'll fix that other one later.
2. You can fix the formatting issues with `make format`.

dankrad:
> 1. That test (which is unrelated to yours) is failing because of an ordering issue that we should probably fix. If you move your new tests after the `test_blob_to_kzg_commitment` section they should all pass. I'll fix that other one later.

Yes, I'm aware. I think that test should be implemented to use a fixed seed, and not be dependent on its place in the sequence.

jtraglia:
> > 1. That test (which is unrelated to yours) is failing because of an ordering issue that we should probably fix. If you move your new tests after the `test_blob_to_kzg_commitment` section they should all pass. I'll fix that other one later.
> 
> Yes, I'm aware. I think that test should be implemented to use a fixed seed, and not be dependent on its place in the sequence.

Okay. This has been fixed.

src/test_c_kzg_4844.c
```diff
@@ -486,6 +743,103 @@ static void test_reverse_bits__succeeds_all_bits_are_one(void) {
...
+    uint32_t reversed[128];
+
+    for (size_t i = 0; i < 128; i++) {
+        get_rand_uint32(&original[i]);
+        reversed[i] = original[i];
+    }
+    ret = bit_reversal_permutation(&reversed, sizeof(uint32_t), 128);
+    ASSERT_EQUALS(ret, C_KZG_OK);
+
+    ASSERT_EQUALS(reversed[0], original[0]);
```
jtraglia, line 780: Could you add a short comment here reminding the reader that it's reordering based on array indices and not the array values? At first, I thought this was a position-dependent test because I forgot how it worked.

jtraglia (commented):
This is great, thanks for this! Didn't see anything wrong, just a few minor nits.

PR #141, opened 2023-02-18
Add test files

jtraglia:
This PR introduces test files which will be used to verify the all the bindings are working. This version only contains a success test with valid values. More tests should be added later.


asn-d6:
Interesting approach.

Two thoughts:
a) I imagine that this text-based format is convenient for c-kzg and that's great. I wonder if we can unify this text-based approach with @kevaundray's cross-library tests. For example, perhaps we could have a script that fetches @kevaundray's repo and turns its json to this text-based format? Maybe. Just throwing this out here.
b) Given the verbosity of the data here, I would put all of that data in a separate `tests/data/` directory, to more clearly separate it from the actual test code.

jtraglia:
@asn-d6 I was hoping the test generator could be the consensus spec-tests, but maybe that's too hopeful. Another option, these could be generated in the C tests. These were generated somewhat manually. I used the Go binding's benchmarks to generate valid inputs/outputs. The others were copied from the old tests. If @kevaundray's repo could do this, I guess that would be fine too. I suspect it will be a bit awkward to add failure test cases in JSON test files, for example given invalid inputs there should be no result. For now, I'm against making a "data" directory as it seems unnecessary.

xrchz:
Including as-is for now. We can continue discussing source and format and dir structure in future as necessary.

PR #142, opened 2023-02-18
Update go bindings

jtraglia:
This PR updates the go bindings to the new interface.

### DIFF
--- a/bindings/go/main.go
+++ b/bindings/go/main.go
@@ -160,6 +160,26 @@ func ComputeKZGProof(blob Blob, zBytes Bytes32) (KZGProof, CKzgRet) {
 }
 
 /*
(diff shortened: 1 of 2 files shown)


asn-d6:
Given that no golang clients are planning to use c-kzg, would it be crazy to just `rm -rf` those bindings for now? No reason to spend time maintaining and updating dead weight.

/cc @kevaundray @dankrad

jtraglia:
> Given that no golang clients are planning to use c-kzg, would it be crazy to just `rm -rf` those bindings for now? No reason to spend time maintaining and updating dead weight.

Yes, that would be crazy. These are very useful for comparing against the go implementations, for example in my differential fuzzer. The work to update them has already been done.

PR #143, opened 2023-02-18
Update rust bindings

jtraglia:
This PR updates the rust bindings to the new interface.

Notable changes:

* Delete old test vectors as I'd like for us to transition to new tests.
* Update benchmarking to test all public functions.
* For `verify_blob_kzg_proof_batch` test counts 1, 2, 4, 8, 16, 32, 64.
* Fix a couple of bugs I spotted while updating, see comments.


bindings/rust/src/lib.rs
```diff
@@ -162,18 +178,13 @@ impl KZGProof {
     }
 
     pub fn compute_kzg_proof(
-        blob: &[Blob],
```
jtraglia, line 165: Just to note, this was a problem before. It is supposed to take a single blob. Testing helps.

bindings/rust/benches/kzg_benches.rs
```diff
@@ -16,53 +23,95 @@ fn generate_random_blob_for_bench(rng: &mut ThreadRng) -> Blob {
...
-                            kzg_commitments.as_slice(),
-                            &kzg_settings,
-                        )
-                        .unwrap()
-                })
-            },
-        );
+        count *= 2;
     }
+    group.finish();
```
jtraglia, line 114: Not sure if it matters, but this wasn't being done before.

bindings/rust/benches/kzg_benches.rs
```diff
@@ -16,53 +23,95 @@ fn generate_random_blob_for_bench(rng: &mut ThreadRng) -> Blob {
...
+                        .take(count)
+                        .collect::<Vec<Bytes48>>(),
+                    &proofs
+                        .clone()
+                        .into_iter()
+                        .take(count)
+                        .collect::<Vec<Bytes48>>(),
+                    &kzg_settings,
+                )
+                .expect("verify failed");
```
jtraglia, line 108: This looks weird. Not sure why I added it. Should look into removing it.

bindings/rust/src/lib.rs
```diff
@@ -283,15 +316,19 @@ impl KZGCommitment {
         hex::encode(self.bytes)
     }
 
-    pub fn blob_to_kzg_commitment(blob: Blob, kzg_settings: &KZGSettings) -> Self {
+    pub fn blob_to_kzg_commitment(blob: Blob, kzg_settings: &KZGSettings) -> Result<Self, Error> {
         let mut kzg_commitment: MaybeUninit<KZGCommitment> = MaybeUninit::uninit();
         unsafe {
-            blob_to_kzg_commitment(
+            let res = blob_to_kzg_commitment(
```
jtraglia, line 322: This was a problem before too. It didn't check the return value. We wouldn't have known if it failed.

asn-d6: Good catch. Let's add `#[must_use]` on top of `pub enum C_KZG_RET` in `bindings.rs` so that the compiler warns us if we are ignoring any `C_KZG_RET` in the future.

asn-d6: Done in 9559dd081f9044013b164f08a33347da6a31b4f3.

jtraglia: Thank you!

bindings/rust/src/lib.rs
jtraglia: Also, just to be clear, this was my fault. I added this function a while ago.

bindings/rust/src/bindings.rs
```diff
@@ -212,41 +213,48 @@ extern "C" {
 
     pub fn free_trusted_setup(s: *mut KZGSettings);
 
+    pub fn blob_to_kzg_commitment(
```
asn-d6, line 216: Just out of curiosity: Did you do these changes to `bindings.rs` by running `bindgen`?

jtraglia: No, I did not. Updated these manually.

bindings/rust/benches/kzg_benches.rs
jtraglia: Ah it was complaining about an unused result. Replaced with unwrap.

jtraglia:
@pawanjay176 when you get a chance, could you review these changes?

bindings/rust/src/bindings.rs
asn-d6: @pawanjay176 Hey hey. Have you considered integrating bindgen to the build.rs? https://rust-lang.github.io/rust-bindgen/tutorial-3.html

pawanjay176: I used bindgen to generate the first pass of the bindings. It's a bit of a pain to keep it automated as the auto generated code needs a few tweaks to get working even after using all available config options in the bindgen builder. 
Also, I figured if we end up changing the interface, it will need more work to get stuff compiling and all tests and benchmarks running again.

asn-d6: > I used bindgen to generate the first pass of the bindings. It's a bit of a pain to keep it automated as the auto generated code needs a few tweaks to get working even after using all available config options in the bindgen builder. Also, I figured if we end up changing the interface, it will need more work to get stuff compiling and all tests and benchmarks running again.

Hmm, the interface has changed since the first pass of the bindings and there was also a new release of bindgen two weeks ago. Would you be up for running bindgen once more for the latest interface (which is most likely gonna be the final one) so that we have the latest and greatest?

bindings/rust/src/lib.rs
```diff
@@ -107,6 +108,21 @@ impl Drop for KZGSettings {
     }
 }
 
+impl Blob {
+    pub fn from_bytes(bytes: &[u8]) -> Result<Self, Error> {
```
asn-d6, line 112: Where do we use these methods? If we use them just for tests, can we perhaps make this clear?

In particular, perhaps at this point it would be a good idea to split tests to a different file and move all these helpers there too? I think it will be nice to know which functions are security-related (and need to be audited) and which functions are only used in tests.

pawanjay176: Unless there's some reason for not allowing the user of the lib to create a `Blob` from a slice, it's a useful function to expose to the user imo.

bindings/rust/src/lib.rs
```diff
@@ -2,6 +2,7 @@
 #![allow(non_camel_case_types)]
 #![allow(non_snake_case)]
 
+extern crate core;
```
pawanjay176, line 5: Not sure if we need this. Seems to build okay without this.

pawanjay176 (changes requested):
Looks good on the whole. Some style related nits and potential issue with the benchmarks

bindings/rust/src/bindings.rs
asn-d6: > Sorry I'm a bit confused. If I understood you correctly earlier, you were suggesting that we auto-generate the bindings in `build.rs` instead of manually changing bindings.rs everytime there's a change in the C interface.

Sorry for the confusion. Auto-generating was indeed what I suggested in the beginning, but after you explained that it's complicated to bake bindgen into `build.rs`, I changed my mind and just suggested to run the latest bindgen manually one more time.

I suggested that because it's been many months since that original bindgen run and the interface has changed quite a bit since then. Also there has been a new bindgen version since then and the [bindgen docs](https://rust-lang.github.io/rust-bindgen/tutorial-1.html)  say "Please always use the latest version of bindgen, it has the most fixes and best compatibility".

bindings/rust/src/lib.rs
asn-d6: Good point, because `Blob` is part of the public interface.

However, objects like `KZGCommitment` and `KZGProof` are not part of the public interface, so I don't think a user of the lib would need to create a KZGCommitment or KZGProof from bytes.

bindings/rust/benches/kzg_benches.rs
```diff
@@ -16,53 +23,94 @@ fn generate_random_blob_for_bench(rng: &mut ThreadRng) -> Blob {
     arr.into()
 }
 
+fn generate_random_commitment(rng: &mut ThreadRng, s: &KZGSettings) -> Bytes48 {
+    let blob = generate_random_blob(rng);
+    KZGProof::compute_blob_kzg_proof(blob, s)
```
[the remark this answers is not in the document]
jtraglia: Yes, that would probably be better (faster). But it doesn't really matter. We just need a valid G1 point.

bindings/rust/benches/kzg_benches.rs
```diff
@@ -16,53 +23,94 @@ fn generate_random_blob_for_bench(rng: &mut ThreadRng) -> Blob {
...
+    c.bench_function("compute_blob_kzg_proof", |b| {
+        b.iter(|| KZGProof::compute_blob_kzg_proof(*blobs.first().unwrap(), &kzg_settings))
+    });
 
-        group.bench_with_input(
-            BenchmarkId::new("compute_aggregate_kzg_proof", *num_blobs),
-            &blobs,
-            |b, blobs| b.iter(|| KZGProof::compute_aggregate_kzg_proof(blobs, &kzg_settings)),
-        );
+    c.bench_function("verify_blob_kzg_proof", |b| {
```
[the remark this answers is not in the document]
jtraglia: I think it's fine if it only returns false. It's going to do the exact same work. It won't short circuit.

bindings/rust/benches/kzg_benches.rs
```diff
@@ -16,53 +23,94 @@ fn generate_random_blob_for_bench(rng: &mut ThreadRng) -> Blob {
...
+        })
+    });
 
-        let kzg_commitments: Vec<Bytes48> = blobs
-            .clone()
-            .into_iter()
-            .map(|blob| KZGCommitment::blob_to_kzg_commitment(blob, &kzg_settings).to_bytes())
-            .collect();
-        let proof = KZGProof::compute_aggregate_kzg_proof(&blobs, &kzg_settings).unwrap();
+    let mut group = c.benchmark_group("verify_blob_kzg_proof_batch");
```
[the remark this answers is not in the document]
jtraglia: Yeah I think this is fine. As long as the inputs are valid, we should be good.

bindings/rust/benches/kzg_benches.rs
```diff
@@ -16,53 +23,94 @@ fn generate_random_blob_for_bench(rng: &mut ThreadRng) -> Blob {
...
+    let mut count: usize = 1;
+    while count <= max_count {
```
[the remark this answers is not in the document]
jtraglia: Definitely, but with `[1, 2, 4, 8, 16, 32, 64]` to be clear.

bindings/rust/src/lib.rs
jtraglia: By the way, this is used when loading a blob from a test file. I also think it's best to keep it.

bindings/rust/src/lib.rs
```diff
@@ -390,51 +441,102 @@ mod tests {
         test_simple(trusted_setup_file);
     }
 
+    fn getBlob(path: PathBuf) -> Blob {
```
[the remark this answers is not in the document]
jtraglia: Definitely. That's my Go/Java-brain mistake.

bindings/rust/benches/kzg_benches.rs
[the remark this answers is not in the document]
asn-d6: Fixed in 1036a79

bindings/rust/src/lib.rs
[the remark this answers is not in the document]
asn-d6: Fixed in 2493e02

bindings/rust/src/lib.rs
```diff
@@ -390,51 +441,102 @@ mod tests {
...
+        let inputBytes = hex::decode(inputStr.as_bytes()).unwrap();
+        Bytes48::from_bytes(inputBytes.as_slice()).unwrap()
+    }
+
+    fn getBoolean(path: PathBuf) -> bool {
+        let inputStr = fs::read_to_string(path).unwrap();
+        inputStr.contains("true")
+    }
+
+    const blob_to_kzg_commitment_tests: &str = "../../tests/blob_to_kzg_commitment/";
```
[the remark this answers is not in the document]
asn-d6: Fixed in 6d817df

bindings/rust/benches/kzg_benches.rs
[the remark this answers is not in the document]
asn-d6: Fixed in cbeea24

bindings/rust/src/lib.rs
asn-d6: Fixed in fb41231

jtraglia:
Thanks for making those changes @asn-d6!

PR #144, opened 2023-02-18
Update java bindings

jtraglia:
This PR updates the java bindings to the new interface.

Notable changes:

* For benchmarking `verifyBlobKzgProofBatch` do counts 1, 2, 4, 8, 16, 32, 64.
* For `verifyKzgProof` change time scale to ms (instead of ns).
* Tried to update relevant tests to reflect new interface functions.


bindings/java/c_kzg_4844_jni.c
```diff
@@ -125,6 +125,39 @@ JNIEXPORT void JNICALL Java_ethereum_ckzg4844_CKZG4844JNI_freeTrustedSetup(JNIEn
...
+    throw_exception(env, TRUSTED_SETUP_NOT_LOADED);
+    return NULL;
+  }
+
+  size_t blob_size = (size_t)(*env)->GetArrayLength(env, blob);
+  if (blob_size != BYTES_PER_BLOB)
+  {
+    throw_invalid_size_exception(env, "Invalid blob size.", blob_size, BYTES_PER_BLOB);
+    return NULL;
+  }
```
asn-d6, line 141: Hmm, why do we need to do this verbose checking for basic invariants? My suggestion would be to replace these 11 lines of code with two simple assertions:
```suggestion
assert(settings);
assert((size_t)(*env)->GetArrayLength(env, blob) == BYTES_PER_BLOB)
```

I don't understand why we need to fail gracefully if the caller has not initialized the trusted setup, or has provided us with junk data. Let's ensure we don't segfault (i.e. assert) but no reason to do graceful handling of these bad cases.

Same goes for similar boilerplate in all the other functions. I think this change will result in a non-trivial simplfication of this file.

asn-d6 (changes requested):

bindings/java/c_kzg_4844_jni.c
```diff
@@ -146,151 +179,170 @@ JNIEXPORT jbyteArray JNICALL Java_ethereum_ckzg4844_CKZG4844JNI_computeKzgProof(
...
     return 0;
   }
 
-  Bytes48 *commitment_native = (Bytes48 *)(*env)->GetByteArrayElements(env, commitment_bytes, NULL);
-  Bytes48 *proof_native = (Bytes48 *)(*env)->GetByteArrayElements(env, proof_bytes, NULL);
-  Bytes32 *z_native = (Bytes32 *)(*env)->GetByteArrayElements(env, z_bytes, NULL);
-  Bytes32 *y_native = (Bytes32 *)(*env)->GetByteArrayElements(env, y_bytes, NULL);
+  size_t count_native = (size_t)count;
+  size_t blobs_size = (size_t)(*env)->GetArrayLength(env, blobs);
+  if (blobs_size != count_native * BYTES_PER_BLOB)
```
asn-d6, line 312: Similar to above, I think we should panic horribly if the inputs are not the right size, and not do any attempt at failing gracefully. I don't see why a caller would ever give us a wrong-sized blob or commitment.

All these checks can be compressed into three lines of code.

jtraglia: I really disagree with the idea of "panic horribly" with the bindings. A little extra code that handles things properly is the way to go IMO. @StefanBratanov what are your thoughts on this?

StefanBratanov: Hi, my preference is to have concise exceptions in case things go bad. Also I think it's cleaner (even though more lines) since you can see exactly what exception will be thrown with each check. Also not sure just using assert is going to work and will propagate the exception correctly to java.

jtraglia:
@StefanBratanov when you get a chance, could you review these changes?

StefanBratanov:
> @StefanBratanov when you get a chance, could you review these changes?

LGTM

PR #145, opened 2023-02-18
Update python bindings

jtraglia:
This PR updates the python bindings to the new interface.


asn-d6:
Similar thoughts to here as in https://github.com/ethereum/c-kzg-4844/pull/142#issuecomment-1436721428
What is the use of these bindings?

jtraglia:
> What is the use of these bindings?

If/when there are spec tests for c-kzg, these will be useful.

PR #146, opened 2023-02-18
Update csharp bindings

jtraglia:
This PR updates the csharp bindings to the new interface.

### DIFF
--- a/bindings/csharp/Ckzg.Bindings/Ckzg.cs
+++ b/bindings/csharp/Ckzg.Bindings/Ckzg.cs
@@ -5,10 +5,18 @@ namespace Ckzg;
 
 public class Ckzg
 {
-    public const int CommitmentLength = 48;
-    public const int BlobElementLength = 32;
-    public const int BlobLength = BlobElementLength * 4096;
-    public const int ProofLength = 48;
+    public const int BytesPerFieldElement = 32;
+    public const int BytesPerBlob = BytesPerFieldElement * 4096;
+    public const int BytesPerCommitment = 48;
+    public const int BytesPerProof = 48;
+
+    public enum Ret
+    {
+        Ok,
+        BadArgs,
+        Error,
+        Malloc
+    }
 
     static Ckzg() => AssemblyLoadContext.Default.ResolvingUnmanagedDll += (assembly, path) => NativeLibrary.Load($"runtimes/{(
             RuntimeInformation.IsOSPlatform(OSPlatform.Linux) ? "linux" :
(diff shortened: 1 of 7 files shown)


bindings/csharp/Ckzg.Bindings/Ckzg.cs
```diff
@@ -28,58 +51,65 @@ public class Ckzg
...
+    /// <param name="blobs">Blobs as a flattened byte array</param>
+    /// <param name="commitments">Commitments as a flattened byte array</param>
+    /// <param name="proofs">Proofs as a flattened byte array</param>
+    /// <param name="count">The number of blobs/commitments/proofs</param>
     /// <param name="ts">Trusted setup settings</param>
-    [DllImport("ckzg", EntryPoint = "free_trusted_setup_wrap", CallingConvention = CallingConvention.Cdecl)]
-    public static extern void FreeTrustedSetup(IntPtr ts);
+    /// <returns></returns>
+    [DllImport("ckzg", EntryPoint = "verify_blob_kzg_proof_batch", CallingConvention = CallingConvention.Cdecl)]
+    public unsafe static extern Ret VerifyBlobKzgProofBatch(bool* result, byte* blobs, byte* commitments_bytes, byte* proofs_bytes, int count, IntPtr ts);
```
asn-d6, line 113: I'm surprised by how thin this binding is compared to other similar bindings. For example [the Rust bindings](https://github.com/ethereum/c-kzg-4844/blob/15101c3dd55026eef7f026b0fb24f216e0381db3/bindings/rust/src/lib.rs#L256) which do explicit length checks (or the Java bindings which do even more checks).

Am I missing the place where the length checks happen? Or are we inconsistent between the two bindings? If so, we should figure out which one is right.

jtraglia: I think we should make these changes in a subsequent PR, this one has already changed a lot. And I'd prefer if I weren't the one to do this, but I will if no one wants to.

jtraglia:
@flcl42 when you get a chance, could you review these changes?

bindings/csharp/Makefile
```diff
@@ -1,54 +1,54 @@
...
+    ifeq ($(UNAME_S),Linux)
+        ifeq ($(UNAME_M),x86_64)
+            CSHARP_PLATFORM ?= linux-x64
+        else
+            CSHARP_PLATFORM ?= linux-arm64
+        endif
+    endif
+    ifeq ($(UNAME_S),Darwin)
+        ifeq ($(UNAME_M),arm64)
+            CSHARP_PLATFORM ?= osx-arm64
```
flcl42, line 23: Does it really work on arm macos? I'm not sure that github provides arm macos agent at all.

jtraglia: Yes, that's what platform I made these changes on. Also, I believe it does. It's in the CI here:

https://github.com/ethereum/c-kzg-4844/blob/5c33098d6f401899f60c8cc7a6bafcba959188f6/.github/workflows/csharp-bindings-build.yml#L22-L27

flcl42: It's a clang target not a host os, so you can make arm build on x64 mac (macos-latest) setting target as I do here in my draft PR https://github.com/ethereum/c-kzg-4844/pull/46/files#diff-6fc44ed49a172f2e0114b9eac31616d29aac3c679e88877b9b675b884e1b8364R4 here
The current binding does not support macos arm actually. That's is why there is CLANG_PLATFORM, it should be used to specify the target.

flcl42: I'll add those changes after your PR

jtraglia: Ah I understand. Okay thanks 👍

bindings/csharp/Ckzg.Bindings/Ckzg.cs
flcl42: It's was delegated to a consumer library, I can bring that checks from there, diffidently a good idea to have them here.

asn-d6: Yes please! :)

PR #147, opened 2023-02-18
Update nodejs bindings

jtraglia:
This PR updates the nodejs bindings to the new interface.

Notable changes:

* Add new interface functions, remove old.
* Group test functions by interface function name.


bindings/node.js/kzg.cxx
```diff
@@ -152,190 +152,215 @@ Napi::Value BlobToKzgCommitment(const Napi::CallbackInfo& info) {
...
+  auto commitments_count = commitments_param.Length();
+  auto proofs_count = proofs_param.Length();
+
+  if (blobs_count != commitments_count || blobs_count != proofs_count) {
+    Napi::Error::New(env, "verifyBlobKzgProofBatch requires equal number of blobs/commitments/proofs")
+      .ThrowAsJavaScriptException();
     return env.Null();
   }
 
+  auto blobs = (Blob*)calloc(blobs_count, sizeof(Blob));
```
asn-d6, line 314: Wow this required much more C code than all the other functions and actually most other bindings. Isn't there a simpler way to do things with Napi?

jtraglia: Yeah that section is pretty ugly. I'd love for a nodejs person to improve it.

asn-d6: Hello @dgcoffman. Can you please take a look at this section please? :)

jtraglia:
@dapplion when you get a chance, could you review these changes?

bindings/node.js/test.ts
```diff
@@ -46,99 +79,232 @@ describe("C-KZG", () => {
...
   });
 
-  describe("computing commitment from blobs", () => {
-    it("throws as expected when given an argument of invalid type", () => {
-      // @ts-expect-error
-      expect(() => blobToKzgCommitment("wrong type")).toThrowError(
-        "Expected blob to be UInt8Array of 131072 bytes",
-      );
+  describe("verifyBlobKzgProofBatch", () => {
+    it("zero blobs/commitments/proofs", () => {
```
dgcoffman, line 282: These tests previously read as sentences describing the expected behavior. They should remain doing so.

bindings/node.js/kzg.cxx
dgcoffman: This is what the code to extract C-style byte arrays from Napi wrapped C++ data structures actually looks like. It's fine.

bindings/node.js/test.ts
jtraglia: I understand. Will fix.

bindings/node.js/kzg.cxx
```diff
@@ -152,190 +152,215 @@ Napi::Value BlobToKzgCommitment(const Napi::CallbackInfo& info) {
...
   if (argument_count != expected_argument_count) {
     return throw_invalid_arguments_count(expected_argument_count, argument_count, env);
   }
 
-  auto commitment_bytes = extract_byte_array_from_param(info, 0, "commitmentBytes");
-  auto z_bytes = extract_byte_array_from_param(info, 1, "zBytes");
-  auto y_bytes = extract_byte_array_from_param(info, 2, "yBytes");
-  auto proof_bytes = extract_byte_array_from_param(info, 3, "proofBytes");
-  auto kzg_settings = info[4].As<Napi::External<KZGSettings>>().Data();
+  auto blobs_param = info[0].As<Napi::Array>();
```
asn-d6, line 299: Do we need any checks for these casts to `Napi::Array` here? Looking at extract_byte_array_from_param() we are doing a bunch of checks that are not done here.

bindings/node.js/kzg.cxx
```diff
@@ -152,190 +152,215 @@ Napi::Value BlobToKzgCommitment(const Napi::CallbackInfo& info) {
...
     return env.Null();
   }
 
+  auto blobs = (Blob*)calloc(blobs_count, sizeof(Blob));
+   if (blobs == NULL) {
+     Napi::Error::New(env, "Error while allocating memory for blobs").ThrowAsJavaScriptException();
+     return env.Null();
+   };
+
+   auto commitments = (Bytes48*)calloc(commitments_count, sizeof(Bytes48));
```
asn-d6, line 320: I think we are leaking `blobs`, `commitments` and `proofs` here under the happy-case.

I wonder if we can slightly refactor this function to use the `goto err` style to do the frees in the end.

bindings/node.js/kzg.cxx
```diff
@@ -152,190 +152,215 @@ Napi::Value BlobToKzgCommitment(const Napi::CallbackInfo& info) {
...
+     free(blobs);
+     free(commitments);
+     Napi::Error::New(env, "Error while allocating memory for proofs").ThrowAsJavaScriptException();
+     return env.Null();
+   };
+
+  for (uint32_t index = 0; index < blobs_count; index++) {
+    // Extract blob bytes from parameter
+    Napi::Value blob = blobs_param[index];
+    auto blob_bytes = blob.As<Napi::Uint8Array>().Data();
```
asn-d6, line 338: Do we need to do any checks here similar to what we are doing in `extract_byte_array_from_param()`?
I wonder if we can abstract this entire logic into a function instead of doing it three times explicitly (once for proofs, once for blobs, once for commitments).

bindings/node.js/kzg.cxx
```diff
@@ -152,190 +152,215 @@ Napi::Value BlobToKzgCommitment(const Napi::CallbackInfo& info) {
...
-  };
-
-  for (uint32_t blob_index = 0; blob_index < blobs_count; blob_index++) {
-    Napi::Value blob = blobs_param[blob_index];
-    auto blob_bytes = blob.As<Napi::Uint8Array>().Data();
-    memcpy(blobs[blob_index].bytes, blob_bytes, BYTES_PER_BLOB);
-  }
+  auto blob = extract_byte_array_from_param(info, 0, "blob");
+  auto z_bytes = extract_byte_array_from_param(info, 1, "zBytes");
+  auto kzg_settings = info[2].As<Napi::External<KZGSettings>>().Data();
```
asn-d6, line 168: I think we need a:

```
  if (env.IsExceptionPending()) {
    return env.Null();
  }
```
block here to protect against `extract_byte_array_from_param()` failing.

bindings/node.js/kzg.cxx
```diff
@@ -152,190 +152,215 @@ Napi::Value BlobToKzgCommitment(const Napi::CallbackInfo& info) {
...
-  size_t expected_argument_count = 3;
+  size_t expected_argument_count = 2;
   if (argument_count != expected_argument_count) {
     return throw_invalid_arguments_count(expected_argument_count, argument_count, env);
   }
 
   auto blob = extract_byte_array_from_param(info, 0, "blob");
-  auto z_bytes = extract_byte_array_from_param(info, 1, "zBytes");
-  auto kzg_settings = info[2].As<Napi::External<KZGSettings>>().Data();
+  auto kzg_settings = info[1].As<Napi::External<KZGSettings>>().Data();
```
asn-d6, line 198: Same as in https://github.com/ethereum/c-kzg-4844/pull/147/files#r1114171554

bindings/node.js/kzg.cxx
```diff
@@ -152,190 +152,215 @@ Napi::Value BlobToKzgCommitment(const Napi::CallbackInfo& info) {
...
+   };
+
+   auto commitments = (Bytes48*)calloc(commitments_count, sizeof(Bytes48));
+   if (commitments == NULL) {
+     free(blobs);
+     Napi::Error::New(env, "Error while allocating memory for commitments").ThrowAsJavaScriptException();
+     return env.Null();
+   };
+
+   auto proofs = (Bytes48*)calloc(proofs_count, sizeof(Bytes48));
```
asn-d6, line 327: Hmm, given that this requires manual memory allocation, I wonder if we are gonna have problems if we ever enter with `proofs_count == 0` and how to best handle this.

bindings/node.js/kzg.cxx
asn-d6: Hmmm, isn't there a way to represent the arrays as byte arrays from the start (e.g. an [`ArrayBuffer`](https://github.com/nodejs/node-addon-api/blob/main/doc/array_buffer.md)), so that we don't need to do this entire `Array -> Uint8Array` conversion? It's quite error prone (as you can see in the various comments I left) and more complicated than other bindings.

bindings/node.js/kzg.cxx
```diff
@@ -152,190 +152,215 @@ Napi::Value BlobToKzgCommitment(const Napi::CallbackInfo& info) {
...
-  auto blobs_param = info[0].As<Napi::Array>();
-  auto commitments_param = info[1].As<Napi::Array>();
-  auto proof_param = info[2].As<Napi::TypedArray>();
-  auto kzg_settings = info[3].As<Napi::External<KZGSettings>>().Data();
-
-  auto proof_bytes = proof_param.As<Napi::Uint8Array>().Data();
-  auto blobs_count = blobs_param.Length();
-  auto commitments_count = commitments_param.Length();
+  auto commitment_bytes = extract_byte_array_from_param(info, 0, "commitmentBytes");
+  auto z_bytes = extract_byte_array_from_param(info, 1, "zBytes");
```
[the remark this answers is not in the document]
jtraglia: These length checks are done in typescript:

https://github.com/ethereum/c-kzg-4844/blob/b04c35437f45b447d12e02f8ae7b08be6c4e34dc/bindings/node.js/kzg.ts#L170-L174

I tried to do it here a while ago, but couldn't figure out how to get the byte array length.

bindings/node.js/kzg.cxx
jtraglia: Yes, I believe you're right. Good spot. Will fix.

bindings/node.js/kzg.cxx
jtraglia: Hmm I don't think it will fail, but we could short circuit it. There's a test here:

https://github.com/ethereum/c-kzg-4844/blob/d5bd326877659b0a79a89cd48fb4770d7ad387ae/bindings/node.js/test.ts#L284-L286

bindings/node.js/kzg.cxx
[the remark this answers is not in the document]
asn-d6: > These length checks are done in typescript:
> 
> https://github.com/ethereum/c-kzg-4844/blob/b04c35437f45b447d12e02f8ae7b08be6c4e34dc/bindings/node.js/kzg.ts#L170-L174
> 
> I tried to do it here a while ago, but couldn't figure out how to get the byte array length.

Maybe we can use `ByteLength()`?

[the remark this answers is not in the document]
jtraglia: I definitely remember trying to use that function. There's some gotcha that I don't remember.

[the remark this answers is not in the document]
jtraglia: Let's look into this more later.

bindings/node.js/kzg.cxx
jtraglia: Let's look into this more later.

bindings/node.js/kzg.cxx
jtraglia: Fixed.

bindings/node.js/kzg.cxx
jtraglia: Fixed.

bindings/node.js/kzg.cxx
jtraglia: I don't think so, these are checked in typescript. Let's look into it more later.

bindings/node.js/kzg.cxx
jtraglia: Let's look into this more later.

dapplion:
Thank you so much @jtraglia @asn-d6 for the polish effort :heart: we don't have a stable owner of this yet. Hope to sort it out soon

PR #155, opened 2023-02-22
Update bindings to work with JSON reference tests

jtraglia:
This PR will:

* Convert the success tests to JSON.
* Add a new failure test case for each function. Just a placeholder.
* Update each binding to accept JSON tests.


jtraglia:
Going to scrap this & support yaml tests instead. I thought both would be available.

Issue #138, opened 2023-02-17
Make sure that `g1_lincomb()` handles input zero points correctly


asn-d6:
Right now `g1_lincomb()` will attempt to convert all points to affine in batch using `blst_p1s_to_affine()`. However `blst_p1s_to_affine()` [might have undefined behavior](https://github.com/supranational/blst/blob/master/src/multi_scalar.c#L11) if one of the input points is zero (and IIRC returns zero for the result).

This can cause issues in our batch verification routine by zeroing out the pairing, so we should make sure we are protected against it.

Perhaps one way to do it is to check if one of the inputs are zero, and if it is, pass the entire computation through the [basic non-pippenger MSM path](https://github.com/ethereum/c-kzg-4844/blob/9a764de61900c913aa2e6b13db590a52c87370bb/src/c_kzg_4844.c#L818). Another approach would be to try to weed out the bad points from the list of points but this might be more troublesome to implement.

xrchz:
Weeding out the zeros does look tricky, though maybe it's worth benchmarking if it's worthwhile? One way to do it would be to make a copy of `p` (and `coeffs`), then make a single pass over it swapping any zeros (and the corresponding coeff) with the last element.

asn-d6:
Wrote a unittest that demonstrates the bug: https://github.com/ethereum/c-kzg-4844/pull/140/commits/8b7ef3646676f4f1bf7840ec3e95107fe85c59ba

And a demontration of the simple fix: https://github.com/ethereum/c-kzg-4844/pull/140/commits/84c9ef9889c7d58e06fcd1d66447dad81ac5a4d8

Marked the PR as draft for now. Will give a shot at the "weed out zeroes" approach later.

asn-d6:
> Weeding out the zeros does look tricky, though maybe it's worth benchmarking if it's worthwhile? One way to do it would be to make a copy of `p` (and `coeffs`), then make a single pass over it swapping any zeros (and the corresponding coeff) with the last element.

I did not feel smart enough to do the swapping strategy, so I implemented a more naive weeding out strategy here: https://github.com/asn-d6/c-kzg-4844/commit/a39bff7a1244b363f32354d23f6274f1a8a9797a

The benefits of the weeding out strategy start to be seen in batch verify with n=16 where we get an 8% speed boost, and at n=32 we get a 10% speed boost. For smaller n the performance is almost identical (after all we always do the naive MSM for n < 8).

[...]

PR #156, opened 2023-02-23
Fix g1_lincomb() to work with zero points as input

asn-d6:
Implements the weed out strategy from https://github.com/ethereum/c-kzg-4844/issues/138#issue-1588962093

### DIFF
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -769,21 +769,66 @@ static void compute_challenge(
  * We do the second of these to save memory here.
  */
 static C_KZG_RET g1_lincomb(
-    g1_t *out, const g1_t *p, const fr_t *coeffs, uint64_t len
+    g1_t *point_out, const g1_t *p, const fr_t *coeffs, uint64_t len
 ) {
-    C_KZG_RET ret;
+    C_KZG_RET ret = C_KZG_MALLOC;
     void *scratch = NULL;
+    g1_t *new_p = NULL;
+    fr_t *new_coeffs = NULL;
     blst_p1_affine *p_affine = NULL;
     blst_scalar *scalars = NULL;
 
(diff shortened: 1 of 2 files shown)


xrchz:
Why rename to `point_out`? Don't think this is necessary but if you want to do it also update the doc comment.

asn-d6:
> Why rename to `point_out`? Don't think this is necessary but if you want to do it also update the doc comment.

Ah I renamed so that it's not called the same as the goto `out` label.

asn-d6:
Updated doc in 3e0dfd8

dankrad:
I think this is fine and not super complicated.

However, we should check if the blst MSM has ever been audited. If it hasn't been, it might be useless to implement this strategy:
 - for the batch verification code, we should then probably always use plain multiplication
 - for the commitment/proof construction code, we don't have to use this as the trusted setup will never have identity elements.

dankrad:
BLST Pippenger has not been audited. I suggest therefore that we change our strategy to what I wrote above.

xrchz:
> > Why rename to `point_out`? Don't think this is necessary but if you want to do it also update the doc comment.
> 
> Ah I renamed so that it's not called the same as the goto `out` label.

This is inconsistent with the rest of the file: I think we usually use `out` for both return argument and label.

asn-d6:
> I think this is fine and not super complicated.
> 
> However, we should check if the blst MSM has ever been audited. If it hasn't been, it might be useless to implement this strategy:
> 
>     * for the batch verification code, we should then probably always use plain multiplication
> 
>     * for the commitment/proof construction code, we don't have to use this as the trusted setup will never have identity elements.

OK that makes sense. Closing this PR for now, and will open a new one with this strategy.

Thanks all!

PR #161, opened 2023-02-28
Update python bindings to support YAML ref tests

jtraglia:
This PR updates the python bindings to support YAML reference tests.

To use, generate the KZG 4844 tests & replace the `c-kzg-4844/tests` contents with `consensus-spec-tests/tests/general/deneb/kzg` like:

image


ppopth:
lgtm!

ppopth (approved):

PR #163, opened 2023-02-28
Update go bindings to support YAML ref tests

jtraglia:
This updates the go bindings. Not expected to work unless the old tests are replaced with the new yaml ones.


asn-d6:
Great work!

PR #165, opened 2023-02-28
Update csharp bindings to support YAML ref tests

jtraglia:


jtraglia:
I'm planning to merge #46 which already includes these changes.

PR #166, opened 2023-03-01
Update java bindings to support YAML ref tests

jtraglia:
* Update tests to use YAML inputs.
* Add input length checks that were missing.
  * My in-progress tests found that.
* Fix a minor nit in proof length check.


jtraglia:
We're missing yaml tests, but could you give this a preliminary review, @StefanBratanov?

StefanBratanov:
> We're missing yaml tests, but could you give this a preliminary review, @StefanBratanov?

LGTM. I may just put all the yaml POJOs in a separate package to make browsing the code easier. `serializer` maybe? Not sure about the naming.

jtraglia:
> LGTM. I may just put all the yaml POJOs in a separate package to make browsing the code easier. `serializer` maybe? Not sure about the naming.

Good suggestion. Moved them to a new `test_formats` directory, which matches the rust bindings.

PR #168, opened 2023-03-02
Fix little-endian serialization in compute_challenge()

asn-d6:
That code was supposed to be doing:
```
    degree_poly = int.to_bytes(FIELD_ELEMENTS_PER_BLOB, 16, 'little')
    data = FIAT_SHAMIR_PROTOCOL_DOMAIN + degree_poly
```

but it was not doing the endianness right.


src/c_kzg_4844.c
```diff
@@ -722,12 +722,11 @@ static void compute_challenge(
     /* Copy domain separator */
     memcpy(offset, FIAT_SHAMIR_PROTOCOL_DOMAIN, 16);
     offset += 16;
-    bytes_from_uint64(
-        offset, 0
-    ); /* need to fill 16 bytes with the degree in little-endian */
-    offset += 8;
     bytes_from_uint64(offset, FIELD_ELEMENTS_PER_BLOB);
     offset += 8;
+    /* leave the rest 8 bytes as zeroes since little endian */
```
jtraglia, line 727:
```suggestion
    /* Set the other bytes of the 16-byte (little-endian) polynomial degree to zero */
```

PR #160, opened 2023-02-27
Rebuild of Nodejs Bindings

matthewkeil:
Follow-up to PR #147 and related to Issue #153, Issue #149 and Issue #118.

## Inclusions
- #153 - Polishes the node bindings
- Unit tests for arg and return types
- `tools` folder with utils, dev, publish scripts
- #149 - Updated files/filestructure in `bindings/node.js`
- #118 - Add `-fPIC` to linux build flag

**Notes:**
- I am adding some additional unit tests for the bindings today but those should not affect the code as I was running the code on `watch` while I worked and saw the errors being thrown

Issue #153, opened 2023-02-22
Polish nodejs bindings


asn-d6:
We merged the nodejs bindings in https://github.com/ethereum/c-kzg-4844/pull/147 but there is still polishing work to be done.
Please look at https://github.com/ethereum/c-kzg-4844/pull/147 for unresolved issues.

Most importantly:
- the `VerifyBlobKzgProofBatch()` function is quite messy. How can we simplify it? (See #147 for some ideas)
- verification checks are split between the C code and the typescript code. How can we unify them?

/cc @dapplion @dgcoffman

dapplion:
CC @matthewkeil to check if you have avail to look into these pending items

matthewkeil:
My pleasure @dapplion.  Hello @asn-d6! My name is Matt and I would love to help out with this issue.  I reviewed the code and PR #147.  I have a couple questions to ask, some code samples to show, and some API options you may select from to dial in the final product.

In regards to the first two questions above I think your intuition is correct.
- The manually memory management and memcpy in `VerifyBlobKzgProofBatch` may not be the best approach while on the node thread.  There are ways to obtain the native pointer from the byte array that comes in from JS, and only hold a reference to them while the native library is using the data.
- For the verification check are you referring to prop validation happening across contexts like in [this comment?](https://github.com/ethereum/c-kzg-4844/pull/147#discussion_r1114368382).  If so, yes there is a way to do that natively so that code context is preserved for maintainers.  It will also, arguably, run marginally faster when done natively.

A couple of notes about #147 and the API
- @asn-d6 you are correct in [this comment](https://github.com/ethereum/c-kzg-4844/pull/147#discussion_r1114182798). It is possible.  Its also probably a very good idea, to refactor `napi_typed_array_from_bytes` to avoid the  `uint8_t* -> std::vector<uint_8t> -> Napi::ArrayBuffer -> Napi::Uint8Array` both from both an error/memory safety and from a performance perspective.
- [this may be an issue](https://github.com/ethereum/c-kzg-4844/pull/147#discussion_r1114171554) and will segfault if the `setupHandle` gets overwritten somehow.  It's instantiated with `let`.  It possible to store the `KZGSettings` as a native pointer,  [SetInstanceData](https://github.com/nodejs/node-addon-api/blob/main/doc/env.md#setinstancedata-1), so it doesn't have to be passed with execution calls to guard against this.
[...]

PR #170, opened 2023-03-03
Node binding argument and return type updates

matthewkeil:
## Notes
Relates to #153 and is follow-up to #147.  Originally proposed in #160, this PR is one of the pieces that #160 was broken into.

## Inclusions
- move argument validation from TS to C
- additional unit tests for argument validation
- replace `napi_typed_array_from_bytes` with `Buffer::Copy`


bindings/node.js/kzg.cxx
```diff
@@ -32,42 +32,28 @@ Napi::Value throw_invalid_argument_type(const Napi::Env env, std::string name, s
...
-
-  for (size_t i = 0; i < length; ++i) {
-    (*vector)[i] = array[i];
+inline uint8_t *get_bytes(
+    const Napi::Env &env,
+    Napi::Value val,
+    size_t length,
+    std::string &&name)
+{
+  if (!(val.IsTypedArray() && val.As<Napi::TypedArray>().TypedArrayType() == napi_uint8_array)) {
```
asn-d6, line 41: No strong opinion but IMO the previous approach of condition checking is more standard and easier to grok:
`  if (!param.IsTypedArray() || param.TypedArrayType() != napi_uint8_array) {`

bindings/node.js/kzg.cxx
```diff
@@ -32,42 +32,28 @@ Napi::Value throw_invalid_argument_type(const Napi::Env env, std::string name, s
...
-
-  return Napi::Uint8Array::New(env, length, buffer, 0);
-}
-
-const uint8_t * extract_byte_array_from_param(const Napi::CallbackInfo& info, const int index, const std::string name) {
-  auto param = info[index].As<Napi::TypedArray>();
-  if (!param.IsTypedArray() || param.TypedArrayType() != napi_uint8_array) {
-    throw_invalid_argument_type(info.Env(), name, "UInt8Array");
+  Napi::Uint8Array array = val.As<Napi::Uint8Array>();
+  if (array.ByteLength() != length) {
```
asn-d6, line 48: That's nice  and simple!

bindings/node.js/kzg.cxx
```diff
@@ -32,42 +32,28 @@ Napi::Value throw_invalid_argument_type(const Napi::Env env, std::string name, s
...
 
-Napi::TypedArrayOf<uint8_t> napi_typed_array_from_bytes(uint8_t* array, size_t length, Napi::Env env) {
-  // Create std::vector<uint8_t> out of array.
-  // We allocate it on the heap to allow wrapping it up into ArrayBuffer.
-  std::unique_ptr<std::vector<uint8_t>> vector =
-      std::make_unique<std::vector<uint8_t>>(length, 0);
-
-  for (size_t i = 0; i < length; ++i) {
-    (*vector)[i] = array[i];
+inline uint8_t *get_bytes(
```
asn-d6, line 35: Maybe a basic small docstring here would be helpful for future rereference.

bindings/node.js/kzg.cxx
```diff
@@ -126,19 +112,11 @@ Napi::Value FreeTrustedSetup(const Napi::CallbackInfo& info) {
...
-  size_t argument_count = info.Length();
-  size_t expected_argument_count = 2;
-  if (argument_count != expected_argument_count) {
-    return throw_invalid_arguments_count(expected_argument_count, argument_count, env);
-  }
-
-  Blob *blob = (Blob *)extract_byte_array_from_param(info, 0, "blob");
-  if (env.IsExceptionPending()) {
-    return env.Null();
+  Napi::Env env = info.Env();
```
asn-d6, line 115: Hmm.. How come we don't need the argument count checking anymore? Was it useless before?

bindings/node.js/kzg.cxx
```diff
@@ -126,19 +112,11 @@ Napi::Value FreeTrustedSetup(const Napi::CallbackInfo& info) {
...
+  Blob *blob = reinterpret_cast<Blob *>(get_bytes(env, info[0], BYTES_PER_BLOB, "blob"));
```
asn-d6, line 116: Maybe instead of this reinterpret_cast thing around the codebase, we can introduce four wrapper functions to `get_bytes` that do: `get_blob()`, `get_commitment()`, `get_proof()`, `get_field_element()` and only use those. Then `get_bytes` becomes an internal helper function to them.

bindings/node.js/kzg.cxx
```diff
@@ -149,21 +127,20 @@ Napi::Value BlobToKzgCommitment(const Napi::CallbackInfo& info) {
     return env.Undefined();
   };
 
-  return napi_typed_array_from_bytes((uint8_t *)(&commitment), BYTES_PER_COMMITMENT, env);
+  return Napi::Buffer<uint8_t>::Copy(env, reinterpret_cast<uint8_t *>(&commitment), BYTES_PER_COMMITMENT);
```
asn-d6, line 130: That's much simpler! Is there any weird way that Buffer::Copy might fail that we should be aware of?

bindings/node.js/kzg.cxx
```diff
@@ -296,73 +274,58 @@ Napi::Value VerifyBlobKzgProof(const Napi::CallbackInfo& info) {
...
-    Napi::Error::New(env, "verifyBlobKzgProofBatch requires equal number of blobs/commitments/proofs")
-      .ThrowAsJavaScriptException();
-    result = env.Null();
-    goto out;
+  uint32_t count = blobs_param.Length();
+  if (count != commitments_param.Length() || count != proofs_param.Length()) {
+    Napi::Error::New(env, "requires equal number of blobs/commitments/proofs").ThrowAsJavaScriptException();
+    return result;
   }
+  // allocate all three at once. if any fails goto out; will call free on all three so all must exist
```
asn-d6, line 293: I think the previous approach of initializing these pointers on the top as NULL:
```
  Blob *blobs = NULL;
  Bytes48 *commitments = NULL;
  Bytes48 *proofs = NULL;
```

is a more standard C approach, since it ensures that even if something fails (or the code changes) and we go to `out` sooner, they will always be NULL and `free` will handle them correctly.

I also think that allocating a single array and then checking it, instead of allocating them all at once and then checking in the end is more standard, but no strong opinion here.

bindings/node.js/kzg.cxx
```diff
@@ -260,17 +237,19 @@ Napi::Value VerifyKzgProof(const Napi::CallbackInfo& info) {
 
 // verifyBlobKzgProof: (blob: Blob, commitmentBytes: Bytes48, proofBytes: Bytes48, setupHandle: SetupHandle) => boolean;
```
asn-d6, line 238: Can you please add some comments on the expected return values (and exceptions) of this function, and all other verification functions (they are the most important functions here so let's take good care of them).

bindings/node.js/kzg.cxx
```diff
@@ -371,13 +334,12 @@ Napi::Value VerifyBlobKzgProofBatch(const Napi::CallbackInfo& info) {
...
     commitments,
     proofs,
-    blobs_count,
+    count,
     kzg_settings
   );
 
   if (ret != C_KZG_OK) {
     Napi::TypeError::New(env, "Error in verifyBlobKzgProofBatch").ThrowAsJavaScriptException();
-    result = env.Null();
```
asn-d6, line 380: Hmm, I'm confused a bit. Why did we change `result` to default to `Undefined`? Shouldn't we always return `Null` when there is a bad verification? In the other verify functions we return `Null` but here we return `Undefined`.

bindings/node.js/kzg.cxx
```diff
@@ -296,73 +274,58 @@ Napi::Value VerifyBlobKzgProof(const Napi::CallbackInfo& info) {
...
-    Napi::Value proof = proofs_param[index];
-    auto proof_bytes = proof.As<Napi::Uint8Array>().Data();
-    memcpy(&proofs[index], proof_bytes, BYTES_PER_PROOF);
+  for (uint32_t index = 0; index < count; index++) {
+    // add HandleScope here to release reference to temp values
+    // after each iteration since data is being memcpy
+    Napi::HandleScope scope{env};
+    uint8_t *blob = get_bytes(env, blobs_param[index], BYTES_PER_BLOB, "blob");
+    if (blob == nullptr) {
+      goto out;
```
asn-d6, line 316: Should we be throwing an exception if this happens, like in all the other failure cases (except from bad verification equation)? 

(Same for all the cases above)

bindings/node.js/kzg.cxx
```diff
@@ -260,17 +237,19 @@ Napi::Value VerifyKzgProof(const Napi::CallbackInfo& info) {
...
 Napi::Value VerifyBlobKzgProof(const Napi::CallbackInfo& info) {
-  auto env = info.Env();
-
-  size_t argument_count = info.Length();
-  size_t expected_argument_count = 4;
-  if (argument_count != expected_argument_count) {
-    return throw_invalid_arguments_count(expected_argument_count, argument_count, env);
+  Napi::Env env = info.Env();
+  Blob *blob_bytes = reinterpret_cast<Blob *>(get_bytes(env, info[0], BYTES_PER_BLOB, "blob"));
+  if (blob_bytes == nullptr) {
```
asn-d6, line 242: Why return `Undefined` and not `Null`?

This way the verification functions can return four different things: `Null`, `Undefined`, `true`, `false`.

Wouldn't it make sense to return `Null` in all those cases to keep things simpler?
(Same for all occurences above)

asn-d6 (changes requested):
This is really nice work! Left a bunch of comments and a bunch of nits. Thanks!

bindings/node.js/kzg.cxx
matthewkeil: thanks!

bindings/node.js/kzg.cxx
matthewkeil: sure thing.  updated

bindings/node.js/kzg.cxx
matthewkeil: great idea! updated

bindings/node.js/kzg.cxx
matthewkeil: updated

bindings/node.js/kzg.cxx
matthewkeil: It was definitely useful before to make sure the js functions passed the `KzgSettings *` back to the native side.  If for some reason one of the `requireSetupHandle` calls don't return the settings, [like on this line](https://github.com/matthewkeil/c-kzg-4844/blob/94198b5c18f8b82c2e56d79d7741ff583399f9e2/bindings/node.js/kzg.ts#L183), there will be a segfault [on this line](https://github.com/matthewkeil/c-kzg-4844/blob/94198b5c18f8b82c2e56d79d7741ff583399f9e2/bindings/node.js/kzg.cxx#L142) when trying to unwrap the `External` value.

There is no way to "validate" and external value other than the `IsExternal` method or by checking the number of arguments passed, although the latter technically doesn't protect against some other value getting passed in that argument position.  Which will throw a hard C error and not a JS exception.  Those are non-recoverable generally.

This is one of the main reasons for the [context-aware-addon](https://github.com/matthewkeil/c-kzg-4844/tree/context-aware-addon) branch that I mentioned yesterday and will work to get finalized PR for today.  It will accomplish a couple of things but storing that pointer in C is a biggie.  It wouldn't need to be [passed back](https://github.com/matthewkeil/c-kzg-4844/blob/94198b5c18f8b82c2e56d79d7741ff583399f9e2/bindings/node.js/kzg.cxx#L108) to JS, stored [in JS](https://github.com/matthewkeil/c-kzg-4844/blob/94198b5c18f8b82c2e56d79d7741ff583399f9e2/bindings/node.js/kzg.ts#L75) mutably and then passed back to the native code for [calling the functions](https://github.com/matthewkeil/c-kzg-4844/blob/94198b5c18f8b82c2e56d79d7741ff583399f9e2/bindings/node.js/kzg.ts#L212).

It would be possible to just store it in a static/global variable in C but that will crash a worker thread if someone implements the library using workers.  The safe "node" way to store that pointer is by attaching it to the node process and there are a few ways that Node provides to do that.

For this PR though, would you like me to add the argument length check back in and then also add the `IsExternal` call I mentioned in case that PR does not get merged?  When I broke this first piece out I did so assuming they would all get merged but thinking about that a bit more, its not a great plan, just in-case.  Having the `IsExternal` with a unit test to make sure the check is in all functions might be a really good idea...  What are your thoughts?

As a note, if it does get merged I will be able to pull those added length checks back out because we will be fully vetting all arguments to meet the function signature and the standard practice in JS is to allow for duck typing (if it walks like a duck... let it fly). Ie extra stuff that is passed to functions (arguments or even props/fields of objects) that are unused are left alone and do not raise an error so function chaining, or props getting passed through multiple functions work as expected in all of them.  While not necessarily applicable to this library it will be an option that I would like to put forward for the team to select.

bindings/node.js/kzg.cxx
matthewkeil: The JS exception is being thrown in `get_bytes` as (`Napi::Error::New(...).ThrowAsJsException()`) [here](https://github.com/matthewkeil/c-kzg-4844/blob/f10317547860753dddb2633f22121d3412136cc2/bindings/node.js/kzg.cxx#L72). The check for `nullptr` is looking to see if a JS exception was raised to resolve the c stack cleanly.  JS exceptions are handled as application logic and will unwind the application stack, and not an actual c exception (`thow Napi::Error::New(...)` which inherits from `std::exception`) to prevent the c stack from unwinding violently.

matthewkeil: I added a note about this in the comments that I added to `get_bytes` [here](https://github.com/matthewkeil/c-kzg-4844/blob/f10317547860753dddb2633f22121d3412136cc2/bindings/node.js/kzg.cxx#L52).  Please let me know if I should add some additional comments elsewhere or if you think that paradigm is confusing

bindings/node.js/kzg.cxx
matthewkeil: Yep.  It's the exact purpose that function was built.  There are not a lot of difference between the `Buffer<uint8_t>` and the underlying `TypedArray<uint8_t>` aside from the `Copy` method.  On the JS side the Buffer object has a couple other features that are nice to haves (like being able to choose encodings when calling `toString('hex')` vs `toString()`). Both have the method but `Buffer` allows passing the encoding, in particular for serialization as a `hex` encoded string.

As far as failure methods, that I am not sure about honestly.  The node repo is one of the most heavily tested and it has huge visibility so stability is key but that is most definitely the answer that you would want to hear.  I did as much googling as I could to try and find search term combinations that would provide some insight but it was not very fruitful.  I also looked through the old node and node-addon-api issues and didnt find much either.

[here](https://github.com/nodejs/node-addon-api/blob/0b53d885f50c83cd7cbcf505d3dcc89cf8c3d59f/test/buffer.cc#L108) is the unit test for that function but i didnt see any failure tests

bindings/node.js/kzg.cxx
matthewkeil: AHHHH... That was my one for the day!  I updated that so it follows standard C practice.

bindings/node.js/kzg.cxx
matthewkeil: Sure, updated.   I actually misread the comment when I was making some changes and intuitively and inadvertently switched them all to `Undefined` due the nature of `null` in JS but reverted and then made everything `Null`.  Sorry for the extra commit.

As a note `null` can pose challenges in JS.  In this instance they value gets buried because of the pending exception but figured it was worth mentioning.  Among other things `typeof null === 'object'` for no logical reason

bindings/node.js/kzg.cxx
matthewkeil: updated.

bindings/node.js/test.ts
```diff
@@ -255,6 +277,40 @@ describe("C-KZG", () => {
...
     });
+    it("throws as expected when given an argument of invalid length", () => {
+      expect(() =>
+        verifyKzgProof(
+          commitmentBadLength,
+          fieldElementValidLength,
+          fieldElementValidLength,
+          proofValidLength,
+        ),
+      ).toThrowError("Expected commitment to be 48 bytes");
```
jtraglia, line 288:
```suggestion
      ).toThrowError("Expected commitmentBytes to be 48 bytes");
```

bindings/node.js/test.ts
```diff
@@ -255,6 +277,40 @@ describe("C-KZG", () => {
...
+        ),
+      ).toThrowError("Expected yBytes to be 32 bytes");
+      expect(() =>
+        verifyKzgProof(
+          commitmentValidLength,
+          fieldElementValidLength,
+          fieldElementValidLength,
+          proofBadLength,
+        ),
+      ).toThrowError("Expected proof to be 48 bytes");
```
jtraglia, line 312:
```suggestion
      ).toThrowError("Expected proofBytes to be 48 bytes");
```

bindings/node.js/test.ts
```diff
@@ -278,14 +334,72 @@ describe("C-KZG", () => {
...
+          proofValidLength,
+        ),
+      ).toThrowError("Expected blob to be 131072 bytes");
+      expect(() =>
+        verifyBlobKzgProof(
+          blobValidLength,
+          commitmentBadLength,
+          proofValidLength,
+        ),
+      ).toThrowError("Expected commitment to be 48 bytes");
```
jtraglia, line 351:
```suggestion
      ).toThrowError("Expected commitmentBytes to be 48 bytes");
```

bindings/node.js/test.ts
```diff
@@ -278,14 +334,72 @@ describe("C-KZG", () => {
...
+      expect(() =>
+        verifyBlobKzgProof(
+          blobValidLength,
+          commitmentValidLength,
+          proofBadLength,
+        ),
+      ).toThrowError("Expected proof to be 48 bytes");
```
jtraglia, line 358:
```suggestion
      ).toThrowError("Expected proofBytes to be 48 bytes");
```

bindings/node.js/kzg.cxx
```diff
@@ -32,41 +33,66 @@ Napi::Value throw_invalid_argument_type(const Napi::Env env, std::string name, s
...
+    return nullptr;
   }
-  return param.As<Napi::Uint8Array>().Data();
+  return array.Data();
+}
+inline Blob *get_blob(const Napi::Env &env, const Napi::Value &val) {
+  return reinterpret_cast<Blob *>(get_bytes(env, val, BYTES_PER_BLOB, "blob"));
+}
+inline KZGCommitment *get_commitment(const Napi::Env &env, const Napi::Value &val) {
+  return reinterpret_cast<KZGCommitment *>(get_bytes(env, val, BYTES_PER_COMMITMENT, "commitment"));
```
jtraglia, line 88:
```suggestion
  return reinterpret_cast<KZGCommitment *>(get_bytes(env, val, BYTES_PER_COMMITMENT, "commitmentBytes"));
```

bindings/node.js/kzg.cxx
```diff
@@ -32,41 +33,66 @@ Napi::Value throw_invalid_argument_type(const Napi::Env env, std::string name, s
...
+}
+inline KZGProof *get_proof(const Napi::Env &env, const Napi::Value &val) {
+  return reinterpret_cast<KZGProof *>(get_bytes(env, val, BYTES_PER_PROOF, "proof"));
```
jtraglia, line 91:
```suggestion
  return reinterpret_cast<KZGProof *>(get_bytes(env, val, BYTES_PER_PROOF, "proofBytes"));
```

bindings/node.js/test.ts
```diff
@@ -278,14 +334,72 @@ describe("C-KZG", () => {
...
+          [proofValidLength, proofValidLength],
+        ),
+      ).toThrowError("Expected blob to be 131072 bytes");
+      expect(() =>
+        verifyBlobKzgProofBatch(
+          [blobValidLength, blobValidLength],
+          [commitmentBadLength, commitmentValidLength],
+          [proofValidLength, proofValidLength],
+        ),
+      ).toThrowError("Expected commitment to be 48 bytes");
```
jtraglia, line 395:
```suggestion
      ).toThrowError("Expected commitmentBytes to be 48 bytes");
```

bindings/node.js/test.ts
```diff
@@ -278,14 +334,72 @@ describe("C-KZG", () => {
...
+      expect(() =>
+        verifyBlobKzgProofBatch(
+          [blobValidLength, blobValidLength],
+          [commitmentValidLength, commitmentValidLength],
+          [proofValidLength, proofBadLength],
+        ),
+      ).toThrowError("Expected proof to be 48 bytes");
```
jtraglia, line 402:
```suggestion
      ).toThrowError("Expected proofBytes to be 48 bytes");
```

jtraglia (changes requested):
This is great, thank you! Just some nits:

bindings/node.js/test.ts
matthewkeil: updated

bindings/node.js/test.ts
matthewkeil: updated

bindings/node.js/test.ts
matthewkeil: updated

bindings/node.js/kzg.cxx
matthewkeil: updated

bindings/node.js/kzg.cxx
matthewkeil: updated

bindings/node.js/test.ts
matthewkeil: updated

bindings/node.js/test.ts
matthewkeil: updated

bindings/node.js/test.ts
matthewkeil: updated

bindings/node.js/kzg.cxx
matthewkeil: added.  I also put the comments in the JS file so they show up with intellisense.

bindings/node.js/kzg.cxx
matthewkeil: UPDATE: I put all length checks back as they were and added `IsExternal` checks before all `As<Napi:External>` calls

bindings/node.js/kzg.cxx
```diff
@@ -32,41 +33,67 @@ Napi::Value throw_invalid_argument_type(const Napi::Env env, std::string name, s
...
+ * 
+ * Designed to raise the correct javascript exception and return a
+ * valid pointer to the calling context to avoid native stack-frame
+ * unwinds.  Calling context can check for `nullptr` to see if an
+ * exception was raised or a valid pointer was returned from V8
+ * 
+ * @param[in] env - napi_env passed from calling context
+ * @param[in] val - napi_value
+ * @param[in] length - size_t to validate ArrayBuffer against
+ * @param[in] name - name of prop being validated for error reporting
```
jtraglia, line 60: Nit, these docstrings don't match the style of the others. Could we update?

bindings/node.js/kzg.cxx
```diff
@@ -173,31 +224,45 @@ Napi::Value ComputeKzgProof(const Napi::CallbackInfo& info) {
...
-Napi::Value ComputeBlobKzgProof(const Napi::CallbackInfo& info) {
-  auto env = info.Env();
 
+/**
+ * Given a blob, return the KZG proof that is used to verify it against the
+ * commitment.
+ * 
+ * @param {Blob} blob - The blob (polynomial) to generate a proof for
+ * 
+ * @returns {KZGProof} - The resulting proof
```
jtraglia, line 248: Nit, noticed mixed usage between `returns` and `return` docs. I think we should stick to one style. I prefer `returns` but not a strong opinion, would be fine with either.

bindings/node.js/kzg.cxx
```diff
@@ -173,31 +224,45 @@ Napi::Value ComputeKzgProof(const Napi::CallbackInfo& info) {
...
 
-// computeBlobKzgProof: (blob: Blob, setupHandle: SetupHandle) => KZGProof;
-Napi::Value ComputeBlobKzgProof(const Napi::CallbackInfo& info) {
-  auto env = info.Env();
 
+/**
+ * Given a blob, return the KZG proof that is used to verify it against the
+ * commitment.
+ * 
+ * @param {Blob} blob - The blob (polynomial) to generate a proof for
```
jtraglia, line 246: In the C file, I think we should document all of the parameters with `[in]` or `[out]`. Most of them are already like this, but some are missing it.

bindings/node.js/kzg.cxx
```diff
@@ -124,46 +151,70 @@ Napi::Value FreeTrustedSetup(const Napi::CallbackInfo& info) {
   return env.Undefined();
 }
 
-// blobToKzgCommitment: (blob: Blob, setupHandle: SetupHandle) => KZGCommitment;
+/**
+ * Convert a blob to a KZG commitment.
+ *
+ * @param {Blob} blob - The blob representing the polynomial to be committed to
+ *
+ * @returns {KZGCommitment} -  The resulting commitment
```
jtraglia, line 159:
```suggestion
 * @returns {KZGCommitment} - The resulting commitment
```

bindings/node.js/kzg.cxx
```diff
@@ -207,46 +272,69 @@ Napi::Value ComputeBlobKzgProof(const Napi::CallbackInfo& info) {
...
+    return env.Null();
   };
 
-  return napi_typed_array_from_bytes((uint8_t *)(&proof), BYTES_PER_PROOF, env);
+  return Napi::Buffer<uint8_t>::Copy(env, reinterpret_cast<uint8_t *>(&proof), BYTES_PER_PROOF);
 }
 
-// verifyKzgProof: (commitmentBytes: Bytes48, zBytes: Bytes32, yBytes: Bytes32, proofBytes: Bytes48, setupHandle: SetupHandle) => boolean;
+/**
+ * Verify a KZG poof claiming that `p(z) == y`
```
jtraglia, line 289:
```suggestion
 * Verify a KZG poof claiming that `p(z) == y`.
```

bindings/node.js/kzg.ts
```diff
@@ -178,32 +122,65 @@ export function freeTrustedSetup(): void {
...
+ *
+ * @throws {TypeError} - for invalid arguments or failure of the native library
+ */
 export function computeBlobKzgProof(blob: Blob): KZGProof {
-  checkBlob(blob);
   return kzg.computeBlobKzgProof(blob, requireSetupHandle());
 }
 
+/**
+ * Verify a KZG poof claiming that `p(z) == y`
```
jtraglia, line 167:
```suggestion
 * Verify a KZG poof claiming that `p(z) == y`.
```

bindings/node.js/kzg.cxx
```diff
@@ -32,41 +33,67 @@ Napi::Value throw_invalid_argument_type(const Napi::Env env, std::string name, s
...
+ * Internal function for argument validation. Prefer to use
+ * the helpers below that already have the reinterpreted casts:
+ * - get_blob
+ * - get_commitment
+ * - get_proof
+ * - get_bytes_32
+ * 
+ * Built to pass in a raw Napi::Value so it can be used like
+ * `get_bytes(env, info[0])` or can also be used to pull props from
+ * arrays like `get_bytes(env, passed_napi_array[2])`
```
jtraglia, line 50:
```suggestion
 * arrays like `get_bytes(env, passed_napi_array[2])`.
```

bindings/node.js/kzg.cxx
```diff
@@ -32,41 +33,67 @@ Napi::Value throw_invalid_argument_type(const Napi::Env env, std::string name, s
...
+ * 
+ * Designed to raise the correct javascript exception and return a
+ * valid pointer to the calling context to avoid native stack-frame
+ * unwinds.  Calling context can check for `nullptr` to see if an
+ * exception was raised or a valid pointer was returned from V8
```
jtraglia, line 55:
```suggestion
 * exception was raised or a valid pointer was returned from V8.
```

jtraglia (approved):
This is like 99% there, just a few more nits.

bindings/node.js/kzg.cxx
```diff
@@ -32,41 +33,67 @@ Napi::Value throw_invalid_argument_type(const Napi::Env env, std::string name, s
...
+  return array.Data();
+}
+inline Blob *get_blob(const Napi::Env &env, const Napi::Value &val) {
+  return reinterpret_cast<Blob *>(get_bytes(env, val, BYTES_PER_BLOB, "blob"));
+}
+inline KZGCommitment *get_commitment(const Napi::Env &env, const Napi::Value &val) {
+  return reinterpret_cast<KZGCommitment *>(get_bytes(env, val, BYTES_PER_COMMITMENT, "commitmentBytes"));
+}
+inline KZGProof *get_proof(const Napi::Env &env, const Napi::Value &val) {
+  return reinterpret_cast<KZGProof *>(get_bytes(env, val, BYTES_PER_PROOF, "proofBytes"));
```
asn-d6, line 92: This was a bad suggestion from my side, but I think it's better if instead of `get_proof()` and `get_commitment()` we just have `get_bytes_48()` for parsing.

There is a convention we are using in the code and spec, where untrusted bytes (before they get into the KZG library) are called `Bytes32` or `Bytes48` and they only get turned to `KZGProof` and `KZGCommitment` after they've been validated and normalized by the library.

For this reason, and because the bindings are receiving untrusted bytes as input (from the perspective of the KZG library) I think we should handle them `Bytes48` when they are passed into the lib. OTOH, it's fine for the KZG library to return a `KZGProof`.

bindings/node.js/kzg.cxx
```diff
@@ -124,46 +151,70 @@ Napi::Value FreeTrustedSetup(const Napi::CallbackInfo& info) {
...
-  Blob *blob = (Blob *)extract_byte_array_from_param(info, 0, "blob");
-  if (env.IsExceptionPending()) {
+  Blob *blob = get_blob(env, info[0]);
+  if (blob == nullptr) {
     return env.Null();
   }
-
+  if (!info[1].IsExternal()) {
+     Napi::TypeError::New(env, "Must pass setupHandle as the last function argument").ThrowAsJavaScriptException();
+  }
```
asn-d6, line 176: I'm not totally sure what this `IsExternal` thing is doing but maybe we forgot it in `freeTrustedSetup()`?

Perhaps we can just put this boilerplate code into a `get_kzg_settings()` function that does:
```
if (!info.isExternal()) {
  blabla
}
auto kzg_settings = info[1].As<Napi::External<KZGSettings>>().Data();
```
?

bindings/node.js/kzg.cxx
```diff
@@ -173,31 +224,45 @@ Napi::Value ComputeKzgProof(const Napi::CallbackInfo& info) {
...
+    z_bytes,
     kzg_settings
   );
 
   if (ret != C_KZG_OK) {
      Napi::Error::New(env, "Failed to compute proof")
       .ThrowAsJavaScriptException();
-    return env.Undefined();
+    return env.Null();
   };
```
asn-d6, line 236: Are these semicolons after the brackets actually necessary in C++?

asn-d6 (changes requested):
I think this is almost there.

asn-d6:
Finally, there are some conflicts because of the newly merged https://github.com/ethereum/c-kzg-4844/pull/167 .

bindings/node.js/kzg.cxx
asn-d6: Makes sense. Thanks!

bindings/node.js/kzg.cxx
jtraglia: Yeah good call, I agree. Also please do `get_bytes48` (without the second underscore).

bindings/node.js/kzg.cxx
```diff
@@ -124,46 +151,70 @@ Napi::Value FreeTrustedSetup(const Napi::CallbackInfo& info) {
...
   C_KZG_RET ret = blob_to_kzg_commitment(&commitment, blob, kzg_settings);
   if (ret != C_KZG_OK) {
      Napi::Error::New(env, "Failed to convert blob to commitment")
       .ThrowAsJavaScriptException();
-    return env.Undefined();
+    return env.Null();
   };
 
-  return napi_typed_array_from_bytes((uint8_t *)(&commitment), BYTES_PER_COMMITMENT, env);
+  return Napi::Buffer<uint8_t>::Copy(env, reinterpret_cast<uint8_t *>(&commitment), BYTES_PER_COMMITMENT);
```
jtraglia, line 187: Fixed the conflicts after merging in main. But some of the tests fail now because the return type has been changed from `Uint8Array` to `Buffer`. It's pretty simple to fix if you modify the tests, but `KZGCommitment` & `KZGProof` are defined as `Uint8Array` at the top of the `kzg.ts` file, so that's what I would expect.

bindings/node.js/kzg.cxx
asn-d6: IIUC this is a thing that will not be needed anymore if we move KZGSettings to be owned by C, right?

bindings/node.js/kzg.cxx
matthewkeil: Updated.

bindings/node.js/kzg.cxx
matthewkeil: Updated.

bindings/node.js/kzg.cxx
matthewkeil: Sure! Updated.

matthewkeil: Sure!  Updated.

bindings/node.js/kzg.cxx
matthewkeil: Updated.

bindings/node.js/kzg.cxx
matthewkeil: Nope.  Was me and my predecessor semicoloning all the things!!!

I made another full pass and removed a couple other extras I found hanging out

bindings/node.js/kzg.cxx
matthewkeil: Updated.

bindings/node.js/kzg.cxx
matthewkeil: I found `return` in the `c_kzg_4844.c` and matched that

bindings/node.js/kzg.cxx
matthewkeil: Updated.

bindings/node.js/kzg.ts
matthewkeil: Updated.

bindings/node.js/kzg.cxx
matthewkeil: Ok.  Updated so that the only places that `KZGCommitment` and `KZGProof` exist now are as returns of `BlobToKzgCommitment`, `ComputeKzgProof` and `ComputeBlobKzgProof`.  Everywhere else the code refers to `Bytes48`.  

I also updated both names to not have the underscore preceding the number

bindings/node.js/kzg.cxx
matthewkeil: Yes, exactly.  If we store the pointer to in C there will be no need to wrap it as External [here](https://github.com/ethereum/c-kzg-4844/blob/7c0bd867d5748e0420c130c91dc5d5033ebb7db0/bindings/node.js/kzg.cxx#L108) and then check every time it gets passed back in because it will be a strongly typed C object stored like [this](https://github.com/matthewkeil/c-kzg-4844/blob/99997c2cd692f2859fb69d188f48369d177bea60/bindings/node.js/kzg.cxx#L27)

`External` values in node are values that are not generated and managed by node.  Ie they are external to v8.  They allow native objects to get passed into JS and then passed back to native code (like the setup handle).

The `IsExternal` check is validating that the `napi_value` is an external value.   In JS there is no such thing as "data types" per-say.  That is what TypeScript was built for.  Instead all variables hold a "javascript value" and that value can be anything (string/number/function/etc).  All of the `Is*` functions allow for reading the meta on the value to know what actual c type the value represents.  All of the `As<Napi::*>()` functions unwrap those values and allow access to the typed JS value.

If a number value is passed into that parameter slot and `As<Napi::External>()` is called on a `Napi::Number` it will throw.

bindings/node.js/kzg.cxx
matthewkeil: Thanks for merging that in @jtraglia.  Its updated, and all tests are passing again.

matthewkeil:
@asn-d6  and @jtraglia I think I got everything above.  There is one last outstanding question that we started in telegram and I copied below.  It related to returning a mix of `undefined` and `null` for error states.  @dgcoffman we were curious if you wanted to weigh in?

------ begin copy from chat ------
"... have a question for you regarding returning `undefined` and `null` for error states.  In node, once the exception is raised, the value returned from the native function will never manifest in the calling JS scope.  The verification functions will act as if they return `true`/`false` or they will throw.   In these instances the return value is more to keep the compiler from hailing a `void return` error.  ....  Was curious of the impetus behind using the two different return types so I can make sure I do it correctly in the draft PR I would like to propose for the KzgSettings point storage."
------ end copy from chat ------

jtraglia:
Awesome work 🥇 Let's deal with that other thing in a different PR. Going to merge.

PR #171, opened 2023-03-04
Allow invalid length inputs in rust tests

jtraglia:
This PR will check that each input value is ok, if not it will assert that the output is null.

* I ran into issues with overflowing the stack, so `Blob::from_bytes` returns a boxed value now.
  * `thread 'tests::test_verify_blob_kzg_proof_batch' has overflowed its stack`
  * If there's a better way of handling this, please let me know.
* The `test.input.get_*` functions return a result now, instead of always unwrapping.


asn-d6:
Looks great! Merging!

PR #174, opened 2023-03-06
Change on compute_[blob_]kzg_proof function signatures

ppopth:
This PR implements the spec change specified in:

* https://github.com/ethereum/consensus-specs/pull/3280

In addition, this PR also updates the tests in `/tests` and implements the Python binding change with respect to the spec change.


src/test_c_kzg_4844.c
```diff
@@ -1089,17 +1089,27 @@ static void test_compute_kzg_proof__succeeds_expected_proof(void) {
     /* Compare the computed proof to the expected proof */
     diff = memcmp(proof.bytes, expected_proof.bytes, sizeof(Bytes48));
     ASSERT_EQUALS(diff, 0);
+
+    bytes32_from_hex(
```
asn-d6, line 1093: Perhaps another approach here would be to evaluate the blob at `input_value` using `evaluate_polynomial_in_evaluation_form` and make sure the output matches what compute_kzg_proof returned.

src/c_kzg_4844.c
```diff
@@ -1049,55 +1050,60 @@ static C_KZG_RET compute_kzg_proof_impl(
...
-    KZGProof *out,
+    KZGProof *proof_out,
+    fr_t *y_out,
     const Polynomial *polynomial,
     const fr_t *z,
     const KZGSettings *s
 ) {
     C_KZG_RET ret;
-    fr_t y;
+    fr_t *y = y_out;
```
asn-d6, line 1102: Hmm, this seems a bit non-standard. Why do we need both `y` and `y_out`?

If we need both, I think it's more standard to use `y` throughout the function and then in the end do `*y_out = y`.

bindings/python/ckzg.c
```diff
@@ -80,22 +86,26 @@ static PyObject* compute_kzg_proof_wrap(PyObject *self, PyObject *args) {
 }
 
 static PyObject* compute_blob_kzg_proof_wrap(PyObject *self, PyObject *args) {
-  PyObject *b, *s;
+  PyObject *b, *c, *s;
 
-  if (!PyArg_UnpackTuple(args, "compute_kzg_proof_wrap", 2, 2, &b, &s) ||
+  if (!PyArg_UnpackTuple(args, "compute_blob_kzg_proof_wrap", 3, 3, &b, &c, &s) ||
```
jtraglia, line 91: Nice catch with the name 🥇

src/c_kzg_4844.c
```diff
@@ -1049,55 +1050,60 @@ static C_KZG_RET compute_kzg_proof_impl(
...
+    bytes_from_bls_field(y_out, &fry);
 
 out:
     return ret;
 }
 
 /**
  * Helper function for compute_kzg_proof() and
- * compute_aggregate_kzg_proof().
+ * compute_blob_kzg_proof().
```
jtraglia, line 1085: Another nice catch. These lines may be able to be joined into one now btw.

tests/verify_kzg_proof/small/verify_kzg_proof_case_commitment_not_in_G1/data.yaml
```diff
@@ -1,4 +1,4 @@
 input: {commitment: '0x8123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef',
-  z: '0x623ce31cf9759a5c8daf3a357992f9f3dd7f9339d8998bc8e68373e54f00b75e', y: '0x0200000000000000000000000000000000000000000000000000000000000000',
+  z: '0x623ce31cf9759a5c8daf3a357992f9f3dd7f9339d8998bc8e68373e54f00b75e', y: '0x00000000fffffffffe5bfeff02a4bd5305d8a10908d83933487d9d2953a7ed73',
```
asn-d6, line 2: Were these changes done by running the new test gen from https://github.com/ethereum/consensus-specs/pull/3280/files ? Just wanted to make sure.

jtraglia (changes requested):
Very nice work. Looks great, just one nit.

Also, please run `make format` to fix the formatting complaints. This will actually require a little manual editing afterwards. Sadly, clang-format doesn't work that great with comments. You will need to indent the doxygen parameter doc that's too long & manually join those two lines I mention in the comment.

src/c_kzg_4844.c
jtraglia: I agree this a little weird. You could also delete the `y` variable and use `y_out` throughout the function.

ppopth: I just thought `y_out` is too long. However, I'm fine with either way. Note that `*y_out = y` may not work. I thought about it previously, but it will not work if `fr_t` is not copyable, so I avoided using it.

tests/verify_kzg_proof/small/verify_kzg_proof_case_commitment_not_in_G1/data.yaml
ppopth: Yes, I got them from re-runing the test gen from that branch. I also don't know why I got different tests.

jtraglia (approved):
Nice. LGTM 👍 We need to wait until the spec gets updated before merging though.

src/c_kzg_4844.c
asn-d6: > I just thought `y_out` is too long. However, I'm fine with either way. Note that `*y_out = y` may not work. I thought about it previously, but it will not work if `fr_t` is not copyable, so I avoided using it.

Hmm. That's true, but we've been assuming that `fr_t` is copyable in a bunch of places already (check for places where we are copying `FR_ZERO`).

asn-d6 (approved):
LGTM! Great work!

Let's wait for the spec PR to get merged now.

jtraglia:
The spec has been updated. Merging. Then I'm going to start work on updating the bindings.

PR #177, opened 2023-03-07
Move setup handle to C in node bindings

matthewkeil:


bindings/node.js/kzg.cxx
```diff
@@ -118,62 +153,26 @@ inline Bytes48 *get_bytes48(const Napi::Env &env, const Napi::Value &val, std::s
...
-  if (kzg_settings == nullptr) {
-    return env.Null();
-  }
-
-  free_trusted_setup(kzg_settings);
-  free(kzg_settings);
+  Napi::Env env = info.Env();
+  KzgBindings *bindings = static_cast<KzgBindings *>(info.Data());
+  if (bindings->IsSetup()) {
+      // QUESTION: Should this throw for re-setup or just ignore like it is?
```
asn-d6, line 160: Seems like something I would throw an exception for, but no strong opinion.

bindings/node.js/kzg.cxx
```diff
@@ -225,11 +219,6 @@ Napi::Value BlobToKzgCommitment(const Napi::CallbackInfo& info) {
  */
 Napi::Value ComputeKzgProof(const Napi::CallbackInfo& info) {
   Napi::Env env = info.Env();
-  size_t argument_count = info.Length();
```
asn-d6, line 228: I don't fully understand what allows us to ignore all these arg length checks now. What changed?

bindings/node.js/kzg.cxx
```diff
@@ -9,55 +9,90 @@
...
+    exports["loadTrustedSetup"] = Napi::Function::New(env, LoadTrustedSetup, "setup", this);
+    exports["blobToKzgCommitment"] = Napi::Function::New(env, BlobToKzgCommitment, "blobToKzgCommitment", this);
+    exports["computeKzgProof"] = Napi::Function::New(env, ComputeKzgProof, "computeKzgProof", this);
+    exports["computeBlobKzgProof"] = Napi::Function::New(env, ComputeBlobKzgProof, "computeBlobKzgProof", this);
+    exports["verifyKzgProof"] = Napi::Function::New(env, VerifyKzgProof, "verifyKzgProof", this);
+    exports["verifyBlobKzgProof"] = Napi::Function::New(env, VerifyBlobKzgProof, "verifyBlobKzgProof", this);
+    exports["verifyBlobKzgProofBatch"] = Napi::Function::New(env, VerifyBlobKzgProofBatch, "verifyBlobKzgProofBatch", this);
+  }
+
+  ~KzgBindings() {
```
asn-d6, line 54: What is this little ~ function? Some sort of destructor? Maybe we can document it as such?

bindings/node.js/kzg.cxx
```diff
@@ -9,55 +9,90 @@
...
 
-Napi::Value throw_invalid_argument_type(const Napi::Env env, std::string name, std::string expectedType) {
-  Napi::TypeError::New(
-    env,
-    "Invalid argument type: " + name + ". Expected " + expectedType
-  ).ThrowAsJavaScriptException();
+  KzgBindings(KzgBindings &&source) = delete;
+  KzgBindings(const KzgBindings &source) = delete;
+  KzgBindings &operator=(KzgBindings &&source) = delete;
+  KzgBindings &operator=(const KzgBindings &source) = delete;
```
asn-d6, line 64: What is this interesting section with the &&s? Can we add some small docs?

asn-d6 (commented):
I really feel like I'm lacking huge amounts of knowledge to be able to review this piece of code. I still left some comments trying to be useful.

Any chance you can find one of your team members to take a look as well? This has more JS-related tidbits than I can be useful with.

bindings/node.js/kzg.cxx
jtraglia: Yeah I would throw an exception here.

jtraglia:
I think this looks pretty good. I think @asn-d6 covered all of the questions/feedback I had.

bindings/node.js/kzg.cxx
matthewkeil: Yep, exactly.

Node provides a stable api for plugging in native code to the execution environment called [N-Api](https://nodejs.org/docs/latest-v12.x/api/n-api.html#n_api_napi_set_instance_data). The `Napi` namespace is from a header only c++ implementation library called `node-addon-api` [doclink](https://github.com/nodejs/node-addon-api/blob/main/doc/env.md#setinstancedata).

I did some research and am refactoring to use the underlying C api and a functional paradigm.  It will avoid the c++ destructor and the move/copy constructors and operators that you brought up.  It uses a struct instead of a class and will look and feel much more familiar.

tl/dr;
In reference to the syntax you mentioned.  It will go away but you are 100% correct, that is a c++ destructor.  When a stack allocated instance goes out of scope, or a heap allocated one is `delete`d that function is called (not called by free though). They also can get sneakily called when temporary objects get deleted, if the location is not eligible for copy elison (like return value optimization).

`&&` is a `right-hand` reference (temporary, unnamed variable like the result of an expression between `()`) and the two signatures with it are the move constructor and move-assignment operator-overload.

The other two signatures with `const &` are similar but for copy construction and copy assignment.

By marking all 4 as `delete` it makes the object immovable and non-copiable.

matthewkeil: @asn-d6 and @jtraglia the first few refactor steps went smoothly and will request a re-review once all the test are green again.

asn-d6: ACK. So I guess we are waiting for the refactoring now. No reason to further review the current code.

bindings/node.js/kzg.cxx
matthewkeil: Updated.

bindings/node.js/kzg.cxx
matthewkeil: Updated.

bindings/node.js/kzg.cxx
matthewkeil: This is not something I feel strongly about but its a general best-practice for JS.  It relates to "duck typing" and when applied to this circumstance for function arguments.   Generally, if an argument is unused it is ignored.  In our case the arguments that come after the `argument_count` are unused, and thus could just be ignored instead of throwing the error.  I will be happy to put it back in though if there is concern that the API will be different for the node bindings

matthewkeil:
Here is a screenshot showing the cleanup function getting run for reference

Screen Shot 2023-03-08 at 11 23 16 AM

bindings/node.js/kzg.cxx
asn-d6: I'm more concerned about providing less arguments than expected. For example, if 2 arguments are provided when 3 are expected, we will dig into `info[2]` which is out-of-bounds.

Are we protecting against this?

matthewkeil: Yep @asn-d6.  We fully vet each argument that gets used.   As an example if a `Bytes32` is expected in the second parameter position, that argument position, (`info[1]`), is checked for existence, for the expected type, `Uint8Array`, and then `.ByteLength()` is called to check it matches 32 bytes.  A distinct error gets thrown in the JS environment for each of those circumstances.  All of the above holds true for every value that gets passed from JS to C for use by the library.

Its only for extra arguments (out-of-bounds) that I was thinking about, like:
```typescript
const isUnused = true;
const result = verifyKzgProof(commitment, z_bytes, y_bytes, proof, isUnused);
```
TypeScript would hail an error for that but in regular JS it would be a valid call.

asn-d6:
Thanks a lot for this @matthewkeil ! Let's get this merged!

PR #183, opened 2023-03-08
Update Compute*KZGProof in rust bindings

asn-d6:
This is still WIP because even though everything works with `cargo test --release`, there is a stack overflow in the end-to-end test when run with `cargo test`.

I might need to box something like @jtraglia did for variable length tests.

### DIFF
--- a/bindings/rust/src/lib.rs
+++ b/bindings/rust/src/lib.rs
@@ -614,7 +614,6 @@ mod tests {
                 assert!(test.get_output().is_none());
                 continue;
             };
-            println!("new");
 
             match KZGProof::verify_blob_kzg_proof_batch(
                 &blobs,
--- a/bindings/rust/src/lib.rs
+++ b/bindings/rust/src/lib.rs
@@ -186,7 +186,13 @@ impl KZGProof {
         let mut kzg_proof = MaybeUninit::<KZGProof>::uninit();
         let mut y_out = MaybeUninit::<Bytes32>::uninit();
         unsafe {
-            let res = compute_kzg_proof(kzg_proof.as_mut_ptr(), y_out.as_mut_ptr(), &blob, &z_bytes, kzg_settings);
+            let res = compute_kzg_proof(
+                kzg_proof.as_mut_ptr(),
+                y_out.as_mut_ptr(),
+                &blob,
+                &z_bytes,
+                kzg_settings,
+            );
             if let C_KZG_RET::C_KZG_OK = res {
                 Ok((kzg_proof.assume_init(), y_out.assume_init()))
             } else {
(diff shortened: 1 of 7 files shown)


jtraglia (commented):
Looks pretty good! I only had one nit & I think `cargo fmt` needs to be run.

bindings/rust/src/lib.rs
```diff
@@ -591,15 +597,12 @@ mod tests {
                 assert!(test.get_output().is_none());
                 continue;
             };
+            println!("new");
```
jtraglia, line 600: This should be deleted.

jtraglia (approved):
LGTM 👍

asn-d6:
@pawanjay176 if you can take a look at the changes it would be great!

bindings/rust/src/lib.rs
```diff
@@ -397,16 +413,17 @@ mod tests {
...
+            .map(|blob| KZGCommitment::blob_to_kzg_commitment(*blob, &kzg_settings).unwrap())
             .map(|commitment| commitment.to_bytes())
             .collect();
 
         let proofs: Vec<Bytes48> = blobs
-            .clone()
-            .into_iter()
-            .map(|blob| KZGProof::compute_blob_kzg_proof(blob, &kzg_settings).unwrap())
+            .iter()
+            .zip(commitments.clone())
```
pawanjay176, line 423:
```suggestion
            .zip(commitments.iter())
```

bindings/rust/src/test_formats/compute_blob_kzg_proof.rs
```diff
@@ -6,14 +6,21 @@ use serde::Deserialize;
...
 pub struct Input<'a> {
     blob: &'a str,
+    commitment: &'a str,
 }
 
 impl Input<'_> {
-    pub fn get_blob(&self) -> Result<Box<Blob>, Error> {
+    pub fn get_blob(&self) -> Result<Blob, Error> {
         let hex_str = self.blob.replace("0x", "");
         let bytes = hex::decode(hex_str).unwrap();
```
pawanjay176, line 15: If the test format specifies a 0x prefixed string, then we could also do 
`hex::decode(self.blob[2..])` to avoid creating a new string.

pawanjay176 (approved):
LGTM. Just a minor nit.
I didn't understand if the stack overflow issue was fixed somehow by other changes since @asn-d6 mentioned that we might need to box the blobs.

asn-d6:
> LGTM. Just a minor nit. I didn't understand if the stack overflow issue was fixed somehow by other changes since @asn-d6 mentioned that we might need to box the blobs.

The stack overflow issue was caused because on a bunch of places we had `Vec<Blob>` (allocated on the heap) which was turned into an iterator with `into_iter()` (iiuc can be allocated on the stack) or used with map which was moving blobs on the heap.

Then, when `cargo test` is run without `--release`, the memory allocation is very verbose and that was causing stack overflow issues.

It has been addressed with a bit more conservative coding when handling blobs in tests.

PR #182, opened 2023-03-07
Update Compute*KZGProof in java bindings

jtraglia:
This PR updates the Java bindings to be in accordance with:

* https://github.com/ethereum/c-kzg-4844/pull/174

PR #184, opened 2023-03-08
Make verify_*() functions pessimistic by default

asn-d6:
Avoids bugs like
https://github.com/ethereum/c-kzg-4844/pull/182/files/ad2c68d4458b9ec649b874a18b0ae3c07434fefe#r1129556029 where the bindings return a wrong retval from the verify functions.

### DIFF
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -932,7 +932,7 @@ static C_KZG_RET verify_kzg_proof_impl(
 /**
  * Verify a KZG proof claiming that `p(z) == y`.
  *
- * @param[out] ok        `true` if the proof is valid, `false` if not
+ * @param[out] ok         `true` if the proof is valid, `false` if not
  * @param[in]  commitment The KZG commitment corresponding to polynomial
  *                        p(x)
  * @param[in]  z          The evaluation point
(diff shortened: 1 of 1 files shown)


src/c_kzg_4844.c
```diff
@@ -932,7 +932,7 @@ static C_KZG_RET verify_kzg_proof_impl(
 /**
  * Verify a KZG proof claiming that `p(z) == y`.
  *
- * @param[out] out        `true` if the proof is valid, `false` if not
+ * @param[out] ok        `true` if the proof is valid, `false` if not
  * @param[in]  commitment The KZG commitment corresponding to polynomial
```
jtraglia, line 936: A little spacing issue here.

jtraglia (commented):
Besides that one little nit, LGTM. Do the other bindings (like C# and Node.js) need to be updated too?

jtraglia:
Oh, I guess there's a build issue. That needs to be fixed too.

src/c_kzg_4844.c
jtraglia, line 936:
```suggestion
 * @param[out] ok         `true` if the proof is valid, `false` if not
 * @param[in]  commitment The KZG commitment corresponding to polynomial
```

PR #185, opened 2023-03-08
detect aliasing in multi-inversion

GottfriedHerold:
fr_batch_inv might (from the doc) reasonably be expected to work if a == out, batch-inverting in-place.
However, the current implementation actually silently fails while giving wrong results for this case.
This PR just makes this fail explicitly. 

Additionally / alternatively, the documentation of this (internal) function should probably reflect this.

### DIFF
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -301,6 +301,9 @@ static void fr_from_uint64(fr_t *out, uint64_t n) {
 /**
  * Montgomery batch inversion in finite field.
  *
+ * @remark This function does not support in-place computation (i.e. `a` MUST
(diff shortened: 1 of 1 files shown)


asn-d6:
Hey @GottfriedHerold . Thanks for digging in here!

Given that this doesn't actually happen in the current codebase (and we don't plan to use it this way) I wonder if we can just add an assertion and a remark in the documentation. @jtraglia opinions?

jtraglia:
Interesting, where would this be an issue? I would also prefer an assert here, like @asn-d6 suggested. Also, I would like for it to go after the function variable declarations.

jtraglia (approved):
LGTM 👍

PR #190, opened 2023-03-09
Allow invalid length inputs in nim tests

jangko:
Description of this PR is mostly similar with:
* #175

And an anticipation for upcoming tests:
* #186


jtraglia (approved):
Nice, this looks great. Relatively simple compared to the other bindings 👍

PR #192, opened 2023-03-09
Make verify_blob_kzg_proof_batch() work with multithreading

asn-d6:
This PR is a PoC replacing `verify_blob_kzg_proof_batch()` with multithreaded calls to `verify_blob_kzg_proof()`. This is made really easy by Rust's rayon crate which is why I only did it for Rust.

The code is quite messy; I just made it as a PoC and I can improve it if we want to go with it.

Two big benefits here:
- This is significantly faster than the status quo (see below)
- We don't touch the more complicated batch verification logic of c-kzg

Two drawbacks:
- This is not gonna work in single threaded machines (but even rpi4s have multiple cores these days iirc)
- An extra dependency for the rust bindings

---

Here are some benchmarks:

* Parallel batch (this PR):
```
verify_blob_kzg_proof_batch/1      time:   [4.6296 ms 4.6458 ms 4.6643 ms]
verify_blob_kzg_proof_batch/2      time:   [5.0340 ms 5.1749 ms 5.3304 ms]
verify_blob_kzg_proof_batch/4      time:   [8.2233 ms 8.2545 ms 8.2890 ms]
verify_blob_kzg_proof_batch/8      time:   [13.109 ms 13.194 ms 13.279 ms]
verify_blob_kzg_proof_batch/16     time:   [19.952 ms 20.080 ms 20.225 ms]
verify_blob_kzg_proof_batch/32     time:   [36.771 ms 36.972 ms 37.187 ms]
verify_blob_kzg_proof_batch/64     time:   [67.669 ms 68.022 ms 68.448 ms]
```

* Regular `verify_blob_kzg_proof_batch()`
```
verify_blob_kzg_proof_batch/1      time:   [4.7198 ms 4.7386 ms 4.7630 ms]
verify_blob_kzg_proof_batch/2      time:   [7.2811 ms 7.3248 ms 7.3990 ms]
verify_blob_kzg_proof_batch/4      time:   [12.417 ms 12.436 ms 12.459 ms]
verify_blob_kzg_proof_batch/8      time:   [22.761 ms 22.772 ms 22.785 ms]
verify_blob_kzg_proof_batch/16     time:   [43.231 ms 43.302 ms 43.398 ms]
verify_blob_kzg_proof_batch/32     time:   [84.329 ms 84.544 ms 84.800 ms]
verify_blob_kzg_proof_batch/64     time:   [170.50 ms 170.72 ms 170.98 ms]
```

(My laptop is slow, so just pay attention to the relative changes between the two approaches)

### DIFF
--- a/bindings/rust/Cargo.toml
+++ b/bindings/rust/Cargo.toml
@@ -13,6 +13,7 @@ minimal-spec = []
 [dependencies]
 hex = "0.4.2"
 libc = "0.2"
+rayon = "1.7.0"
 serde = { version = "1.0", features = ["derive"] }
 
 [dev-dependencies]
--- a/bindings/rust/benches/kzg_benches.rs
+++ b/bindings/rust/benches/kzg_benches.rs
@@ -23,20 +23,6 @@ fn generate_random_blob(rng: &mut ThreadRng) -> Blob {
     arr.into()
 }
 
-fn generate_random_commitment(rng: &mut ThreadRng, s: &KZGSettings) -> Bytes48 {
-    let blob = generate_random_blob(rng);
-    KZGCommitment::blob_to_kzg_commitment(blob, s)
-        .unwrap()
-        .to_bytes()
-}
-
-fn generate_random_proof(rng: &mut ThreadRng, s: &KZGSettings) -> Bytes48 {
-    let blob = generate_random_blob(rng);
-    KZGProof::compute_blob_kzg_proof(blob, s)
-        .unwrap()
-        .to_bytes()
-}
-
 pub fn criterion_benchmark(c: &mut Criterion) {
     let max_count: usize = 64;
     let mut rng = rand::thread_rng();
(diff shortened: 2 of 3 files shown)


dankrad:
> This is significantly faster than the status quo (see below)

It is lower latency, but the total amount of resources consumed is higher.

I'm not sure about this, because I don't see latency as the limiting factor here. Batching of up to 16 seems fine in terms of latency with batch as well, because to verify the validity of the block, execution has to run as well (which is more like a few 100 ms). However, if you hog all the cores for blob verification, then it will slow down other things that could run in parallel (signature verification, execution, etc.)

dankrad:
(Also of course wanted to mention that it is possible to parallelize the batch verification as well)

pawanjay176 (commented):
If we want to make this change, it might make sense to feature gate the parallel implementation behind a feature flag. 
We can benchmark block processing with and without the parallelization to check if this is potentially taking away resources from other block processing tasks and ends up being slower in practice.

asn-d6:
> > This is significantly faster than the status quo (see below)
> 
> It is lower latency, but the total amount of resources consumed is higher.
> 
> I'm not sure about this, because I don't see latency as the limiting factor here. Batching of up to 16 seems fine in terms of latency with batch as well, because to verify the validity of the block, execution has to run as well (which is more like a few 100 ms). However, if you hog all the cores for blob verification, then it will slow down other things that could run in parallel (signature verification, execution, etc.)

Hmm, that's a valid point.

That said, is it the case that clients do operations like verifying blobs and verifying execution in parallel? It's great if they do, but sometimes implementations do operations serially (for whatever internal code architecture reason) and block on each other.

sauliusgrigaitis:
As per discussion on TG, I'm not sure such parallelization at KZG level is very useful for the current "Free the blobs" approach. The optimal way in terms of speed (not CPU cycles) would be to spin a verification task at the client level immediately when blob is received (such a pipeline would be tricky to implement at KZG lib level) and keep it in the verified pool until all the blobs are received and verified. This likely will not hog all the cores as blobs should not come at once. A similar approach we are already using for previous optimizations we implemented in Grandine.

asn-d6:
Closing this PR as it doesn't seem like it's of current interest. Thanks all!

PR #193, opened 2023-03-09
Documentation fix for roots of unity

GottfriedHerold:
The previous statement was slightly wrong. Being a primite root does not mean r^( (q-1)/2) != 1. If q=1 == 2^s * t, the latter is satisfied by e.g. t'th roots of unity, but those are not primitive roots by the usual definition. Fortunately, the actual computation of 2^i'th roots of unity does not require a primitve root to start with, only a non-square.


asn-d6:
Thanks for this!

PR #196, opened 2023-03-09
Run clang sanitizers in CI + workflow cleanup

### LINKED ISSUE #157: Run integration tests with ASan
We should tweak our Makefile to run our unittests with [ASan](https://github.com/google/sanitizers/wiki/AddressSanitizer) at some point, to detect various C memory issues that can happen.

jtraglia:
* Run clang's sanitizers (address, undefined, leak, ...) in the CI tests.
* Fix an undefined behavior in a test.
* Rename workflow files to `<lang>-tests.yml` for simplicity.
* Consolidate C formatting tests into `c-tests.yml`.
* Try to shorten workflow names so all info is visible.
* Run go benchmarks in CI because why not.

Closes #157.


asn-d6:
Just out of curiosity, what happened in https://github.com/ethereum/c-kzg-4844/pull/196/commits/f4c2749e410c68479f83d78c9a1780efe191c7fe ?

jtraglia:
> Just out of curiosity, what happened in [f4c2749](https://github.com/ethereum/c-kzg-4844/commit/f4c2749e410c68479f83d78c9a1780efe191c7fe) ?

That was a finding from `-fsantitize=memory` before it crashed with this error:

```
$ make sanitize_memory
Running memory sanitizer
FATAL: Code 0xaaab4cc9c1e0 is out of application range. Non-PIE build?
FATAL: MemorySanitizer can not mmap the shadow memory.
FATAL: Make sure to compile with -fPIE and to link with -pie.
FATAL: Disabling ASLR is known to cause this error.
FATAL: If running under GDB, try 'set disable-randomization off'.
```

I felt like because of the error, it was showing an incomplete list of findings. I plan to investigate why this is happening more later, but I suspect it's because `blst` isn't built with `-fsanitize=memory`. It seemed wrong to fix one instance when there are other instances of the same thing elsewhere, for example the original + one more:

https://github.com/ethereum/c-kzg-4844/blob/da83e45e9cefdd14bef049ccce9f602b9f7987df/src/c_kzg_4844.c#L228-L232

https://github.com/ethereum/c-kzg-4844/blob/da83e45e9cefdd14bef049ccce9f602b9f7987df/src/c_kzg_4844.c#L243-L248

I may make this change again later. I don't think they are crucial though.

asn-d6:
LGTM! Merging this, and opened https://github.com/ethereum/c-kzg-4844/issues/204 just so we don't forget about the discussion above (which FWIW I think is innocuous).

PR #197, opened 2023-03-09
Reduce boilerplate in nim tests

jangko:
I should have realize this repetitive pattern earlier. 
This PR will replace for-loop, test invocation and try-except with template. 
The result is compact test code.


jtraglia:
Neat. LGTM 👍

PR #202, opened 2023-03-10
Small cleanups

StefanBratanov:
- Made the `.gitignore` file in `/inc` folder same as in `/lib` folder and removed the `inc/` folder ignores in the base `.gitignore`
- Use snake case in the java C binding for consistency
- Parameterize a common library version


jtraglia (approved):
LGTM 👍 that's better, thanks!

PR #206, opened 2023-03-13
Add an into_inner method for Bytes48

pawanjay176:
This is useful for converting between c-kzg's internal representation of bytes and the application's internal representation.


asn-d6:
LGTM!

PR #207, opened 2023-03-13
Automate publishing for C#

flcl42:
The C# package is available at https://www.nuget.org/packages/Ckzg.Bindings/
Previously, it was placed manually there, by uploading the build artifact.
It can be done automatically now.


jtraglia (approved):
LGTM 👍 We've added your API key to the repo secrets.

PR #209, opened 2023-03-14
Null out pointers when failing to construct FFT or KZG Settings

GottfriedHerold:
This aims to prevent a double-free footgun if a user (wrongly) calls the free_* function after construction of the setup fails.

Notably, if there is a (say) out-of-memory error at the second malloc call when creating a KZG / FFT setup and the user (wrongly) then frees the setup, we get a double-free. This PR aims to make that bogus free call a no-op.
Of course, the user is not actually expected to free the setup, but it's a likely error.

### DIFF
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -1672,6 +1672,9 @@ out_error:
     free(fs->expanded_roots_of_unity);
     free(fs->reverse_roots_of_unity);
     free(fs->roots_of_unity);
+    fs->expanded_roots_of_unity = NULL;
+    fs->reverse_roots_of_unity = NULL;
+    fs->roots_of_unity = NULL;
 out_success:
     return ret;
 }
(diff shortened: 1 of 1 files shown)


asn-d6:
Thanks for this PR.

What about instead of handling those frees specially, we define a macro like this:
```
#define c_kzg_free(p) ({ \
    free(p)              \
    (p)=NULL;            \
  })
```

and using this everywhere instead of `free()`? IIUC there are no downsides and we are protecting against double-frees everywhere.

asn-d6:
Please see https://github.com/ethereum/c-kzg-4844/pull/210 for an approach with the macro here.

jtraglia:
@GottfriedHerold thanks for bringing this up. I think I prefer #210.

GottfriedHerold:
I would also prefer #210, voting to close #209.

asn-d6:
Closing this one. Thanks all.

PR #211, opened 2023-03-14
Add ref test count checks for all bindings

jtraglia:
We discovered that the testdir path for one of the bindings was configured incorrectly and the reference tests were not being executed. This was a risk of globing files. Thanks to @asanso for pointing this out.

* Collect test files into a list & ensure there are a non-zero number of them.


asn-d6 (approved):
LGTM! Nice!

PR #212, opened 2023-03-14
Simplify and cleanup C code

jtraglia:
This is a work in progress. This simplifies & cleans up the C library.

* Define static constants for sizes.
* ~Convert macros to consts where possible.~
* Remove backticks from functions in documentation.
* Group g1/g2 multiplications and g1/g2 subtraction funcs together.
* Replace `reverse_bits` with simpler version without macros.
  * Benchmarks show no difference in performance.
* Convert `//` style comments to `/* */` style comments everywhere.
* Replace integers with constants where possible.
* For `verify_kzg_proof_impl` use variable names from spec.
* Remove unnecessary casts now that `FFTSettings` isn't const.

### DIFF
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -1268,13 +1268,13 @@ C_KZG_RET verify_blob_kzg_proof(
 }
 
 /**
+ * Compute random linear combination challenge scalars for batch verification.
  *
- *
- * @param[out]  r_powers_out
- * @param[in]   commitments_g1
- * @param[in]   zs_fr
- * @param[in]   ys_fr
- * @param[in]   proofs_g1
+ * @param[out]  r_powers_out   The output challenges
+ * @param[in]   commitments_g1 The input commitments
+ * @param[in]   zs_fr          The input evaluation points
+ * @param[in]   ys_fr          The input evaluation results
+ * @param[in]   proofs_g1      The input proofs
  */
 static C_KZG_RET compute_r_powers(
     fr_t *r_powers_out,
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -55,11 +55,6 @@ static const char *RANDOM_CHALLENGE_KZG_BATCH_DOMAIN = "RCKZGBATCH___V1_";
 /** Length of the domain strings above. */
 static const size_t DOMAIN_STR_LENGTH = 16;
 
-/** Length of the "metadata" before the data */
-static const size_t DOMAIN_SEPARATOR_LENGTH = DOMAIN_STR_LENGTH +
-                                              sizeof(uint64_t) +
-                                              sizeof(uint64_t);
-
 /** The number of bytes in a g1 point. */
 static const size_t BYTES_PER_G1 = 48;
 
@@ -702,7 +697,8 @@ static C_KZG_RET blob_to_polynomial(Polynomial *p, const Blob *blob) {
 }
 
 /* Input size to the Fiat-Shamir challenge computation. */
-static const size_t CHALLENGE_INPUT_SIZE = DOMAIN_SEPARATOR_LENGTH +
+static const size_t CHALLENGE_INPUT_SIZE = DOMAIN_STR_LENGTH +
+                                           sizeof(uint64_t) + sizeof(uint64_t) +
                                            BYTES_PER_BLOB +
                                            BYTES_PER_COMMITMENT;
 
@@ -1291,7 +1287,8 @@ static C_KZG_RET compute_r_powers(
     Bytes32 r_bytes;
     fr_t r;
 
-    size_t input_size = DOMAIN_SEPARATOR_LENGTH +
+    size_t input_size = DOMAIN_STR_LENGTH + sizeof(uint64_t) +
+                        sizeof(uint64_t) +
                         (n * (BYTES_PER_COMMITMENT +
                               2 * BYTES_PER_FIELD_ELEMENT + BYTES_PER_PROOF));
     ret = c_kzg_malloc((void **)&bytes, input_size);
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -30,6 +30,7 @@
 // Macros
 ///////////////////////////////////////////////////////////////////////////////
 
+/** Returns C_KZG_BADARGS if the condition is not met. */
 #define CHECK(cond) \
     if (!(cond)) return C_KZG_BADARGS
 
@@ -52,6 +53,21 @@
 /** The Fiat-Shamir protocol domains. */
 static const char *FIAT_SHAMIR_PROTOCOL_DOMAIN = "FSBLOBVERIFY_V1_";
 static const char *RANDOM_CHALLENGE_KZG_BATCH_DOMAIN = "RCKZGBATCH___V1_";
+
+/** The number of bytes in a protocol domain. */
+static const size_t BYTES_PER_DOMAIN = 16;
+
+/* Input size to the Fiat-Shamir challenge computation. */
+static const size_t CHALLENGE_INPUT_SIZE = 32 + BYTES_PER_BLOB + 48;
+
+/** The number of bytes in a g1 point. */
+static const size_t BYTES_PER_G1 = 48;
+
+/** The number of bytes in a g2 point. */
+static const size_t BYTES_PER_G2 = 96;
+
+/** The number of g2 points in a trusted setup. */
+static const size_t NUM_G2_POINTS = 65;
 
 /** Deserialized form of the G1 identity/infinity point. */
 static const g1_t G1_IDENTITY = {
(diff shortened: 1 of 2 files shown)


src/c_kzg_4844.c
```diff
@@ -53,6 +54,21 @@
...
+static const size_t CHALLENGE_INPUT_SIZE = 32 + BYTES_PER_BLOB + 48;
+
+/** The number of bytes in a g1 point. */
+static const size_t BYTES_PER_G1 = 48;
+
+/** The number of bytes in a g2 point. */
+static const size_t BYTES_PER_G2 = 96;
+
+/** The number of g2 points in a trusted setup. */
+static const size_t NUM_G2_POINTS = 65;
```
asn-d6, line 70: Perhaps `TRUSTED_SETUP_G2_POINTS_LEN` or something to make it more specific? Or perhaps move this constant down in the trusted setup section?

(I'm not doing a github code suggestion because I can't change both places i think)

src/c_kzg_4844.c
```diff
@@ -719,13 +716,14 @@ static void compute_challenge(
     uint8_t *offset = bytes;
 
     /* Copy domain separator */
-    memcpy(offset, FIAT_SHAMIR_PROTOCOL_DOMAIN, 16);
-    offset += 16;
+    memcpy(offset, FIAT_SHAMIR_PROTOCOL_DOMAIN, BYTES_PER_DOMAIN);
```
asn-d6, line 719: Perhaps instead of BYTES_PER_DOMAIN we can use strlen() here?



Also, usually "domain separator" is the name of the entire header before the actual data, whereas here `BYTES_PER_DOMAIN` only refers to that string.

src/c_kzg_4844.c
```diff
@@ -719,13 +716,14 @@ static void compute_challenge(
...
+    offset += BYTES_PER_DOMAIN;
+
+    /* Copy polynomial degree (16-bytes, little-endian) */
     bytes_from_uint64(offset, FIELD_ELEMENTS_PER_BLOB);
-    offset += 8;
-    /* Set all other bytes of this 16-byte (little-endian) field to zero */
+    offset += sizeof(uint64_t);
```
asn-d6, line 724: Sounds good! Although I think this change is not that useful if we don't include `sizeof(uint64_t)` in the definition of `CHALLENGE_INPUT_SIZE` as well.

src/c_kzg_4844.c
```diff
@@ -1388,29 +1385,26 @@ static C_KZG_RET verify_kzg_proof_batch(
...
-        r_powers, commitments_g1, evaluation_challenges_fr, ys_fr, proofs_g1, n
+        r_powers, commitments_g1, zs_fr, ys_fr, proofs_g1, n
     );
     if (ret != C_KZG_OK) goto out;
 
     /* Compute \sum r^i * Proof_i */
     g1_lincomb_naive(&proof_lincomb, proofs_g1, r_powers, n);
 
     for (size_t i = 0; i < n; i++) {
-        g1_t ys_encrypted;
```
asn-d6, line 1399: Why not have this variable in reduced scope? I think it's better to have it in here to signify to the reader that it's not used anywhere else.

src/c_kzg_4844.c
```diff
@@ -1344,32 +1347,26 @@ static C_KZG_RET compute_r_powers(
...
     bool *ok,
     const g1_t *commitments_g1,
-    const fr_t *evaluation_challenges_fr,
+    const fr_t *zs_fr,
     const fr_t *ys_fr,
     const g1_t *proofs_g1,
     size_t n,
     const KZGSettings *s
 ) {
     C_KZG_RET ret;
```
asn-d6, line 1367: How about we are more explicit about `ret` being set in this function and starting with a fail-closed value in all verification functions?

```suggestion
    C_KZG_RET ret = C_KZG_ERROR;
```

src/c_kzg_4844.c
```diff
@@ -1018,27 +1016,29 @@ C_KZG_RET verify_kzg_proof(
...
+    /* Calculate: P_minus_y */
     g1_mul(&y_g1, &G1_GENERATOR, y);
-    g1_sub(&commitment_minus_y, commitment, &y_g1);
+    g1_sub(&P_minus_y, commitment, &y_g1);
 
-    *out = pairings_verify(
-        &commitment_minus_y, &G2_GENERATOR, proof, &s_minus_x
-    );
+    /* Verify: P - y = Q * (X - z) */
+    *out = pairings_verify(&P_minus_y, &G2_GENERATOR, proof, &X_minus_z);
```
asn-d6, line 1041: changes here lgtm

asn-d6 (changes requested):
Great work! Here is an initial review!

src/c_kzg_4844.c
```diff
@@ -53,6 +54,21 @@
 static const char *FIAT_SHAMIR_PROTOCOL_DOMAIN = "FSBLOBVERIFY_V1_";
 static const char *RANDOM_CHALLENGE_KZG_BATCH_DOMAIN = "RCKZGBATCH___V1_";
 
+/** The number of bytes in a protocol domain. */
+static const size_t BYTES_PER_DOMAIN = 16;
+
+/* Input size to the Fiat-Shamir challenge computation. */
+static const size_t CHALLENGE_INPUT_SIZE = 32 + BYTES_PER_BLOB + 48;
```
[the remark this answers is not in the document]
jtraglia: Defined a new `DOMAIN_SEPARATOR_LENGTH` constant for this.

src/c_kzg_4844.c
jtraglia: So I renamed this to `DOMAIN_STR_LENGTH`. I tried out using `strlen` but wasn't really a fan. One concrete reason, I would like to be able to use this value (the str length) in the new `DOMAIN_SEPARATOR_LENGTH` constant. If we use `strlen` at runtime this isn't available.

src/c_kzg_4844.c
jtraglia: I don't think that's really necessary. That value will be overwritten every time.

jtraglia:
@asn-d6 Feel free to make commits to this PR. Also, there's missing documentation for `compute_r_powers` that should probably be added. I would appreciate it if you did that.

src/c_kzg_4844.c
asn-d6: Another alternative just for completeness is to turn the string into a constant array, and then use `sizeof(str) - 1` at compile time. IIRC that works as well.

jtraglia: Hmm maybe. Not really a fan of the `-1` part though.

src/c_kzg_4844.c
jtraglia: Included `sizeof(uint64_t)` in the definition, part of `DOMAIN_SEPARATOR_LENGTH`.

jtraglia:
I guess the CI is building with an old version of gcc or something.

```
./c_kzg_4844.h:79:13: error: variable length array folded to constant array as an extension [-Werror,-Wgnu-folding-constant]
[11](https://github.com/ethereum/c-kzg-4844/actions/runs/4429177546/jobs/7769280514#step:4:12)
    uint8_t bytes[BYTES_PER_BLOB];
[12](https://github.com/ethereum/c-kzg-4844/actions/runs/4429177546/jobs/7769280514#step:4:13)
            ^
[13](https://github.com/ethereum/c-kzg-4844/actions/runs/4429177546/jobs/7769280514#step:4:14)
1 error generated.
```

This wasn't happening for me locally, or with `CC=gcc make`. Reverted the changes.

src/c_kzg_4844.c
```diff
@@ -697,9 +706,6 @@ static C_KZG_RET blob_to_polynomial(Polynomial *p, const Blob *blob) {
     return C_KZG_OK;
 }
 
-/* Input size to the Fiat-Shamir challenge computation */
```
asn-d6, line 700: I'm not sure how I feel about moving and grouping all these constants on the top. Especially for constants that are not globally relevant or globally needed, I think it makes sense to keep them close to the place where they are used. Both so that it's easier to see which part of the code they are used in, but also because IMO having them all grouped up on the top doesn't offer any major advantages to readability.

jtraglia: No problem. I understand. I'll move `CHALLENGE_INPUT_SIZE` back to be above it's function.

jtraglia:
I think it could use some more comments throughout, but I think this PR has done enough. Marking is ready.

asn-d6:
Pushed a commit that removes `DOMAIN_SEPARATOR_LENGTH`.

We were always serializing using `DOMAIN_STR_LENGTH + uint64_t + uint64_t` anyway, so DOMAIN_SEPARATOR_LENGTH was not helpful. You always had to check that DOMAIN_SEPARATOR_LENGTH is correct. Let's just use the basic elements now.

asn-d6 (approved):
LGTM! If you also like 844a9cfc9a91d40010d34a5592a07cf8123a1227, let's get this merged.

jtraglia:
LGTM thanks!

PR #213, opened 2023-03-15
[Java bindings] Improve javadoc + update dependencies

StefanBratanov:
We create a javadoc jar when we package the java bindings in https://github.com/ConsenSys/jc-kzg-4844 , so it would be useful to run the `javadoc` task as part of the `check`.

Added some small comments to few fields + a rename in the binding to be consistent

Updated dependencies (gradle + plugins)


bindings/java/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.6-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.0.2-bin.zip
+networkTimeout=10000
```
jtraglia, line 4: I'm guessing this will help the intermittent CI failures? If so, great 👍

bindings/java/src/main/java/ethereum/ckzg4844/Tuple.java
```diff
@@ -1,5 +1,6 @@
 package ethereum.ckzg4844;
 
+/** A tuple holding 2 byte arrays. */
 public class Tuple {
```
jtraglia, line 4: Do you think it's worth renaming this class? Something like `ByteArrayTuple`?

Initially, this class used generics but I ran into an issue in JNI and switched to hardcoded `byte[]` types. It may be worth revisiting this to see if you get it working with generics.

jtraglia (approved):
LGTM 👍 thanks!

bindings/java/gradle/wrapper/gradle-wrapper.properties
StefanBratanov: Don't know actually, It's just the default generated when I ran `gradle wrapper --gradle-version 8.0.2`

jtraglia:
Thanks, looks good.

PR #214, opened 2023-03-15
Replace all "Kzg" with "KZG" in java bindings

jtraglia:
In Teku, I believe it's most common for KZG to be all uppercase. Find/replace all instances in the java bindings.

For example in Teku:

https://github.com/ConsenSys/teku/blob/9d0696dff30965a3f210cf3822e04218f9a0e756/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/MiscHelpers.java#L272

@StefanBratanov how does this look?


jtraglia:
Actually, nevermind. Usage in Teku seems pretty mixed & all of the other bindings (except for Go) are like this too.

StefanBratanov:
Yeah, no strong preference for this. Will try to be more consistent in Teku.

jtraglia:
Sounds good. If you guys decide to go the "KZG" route, we can re-open this PR.

Issue #205, opened 2023-03-13
Be intentional about zero inputs in `fr_batch_inv()`


asn-d6:
`fr_batch_inv()` is used in c-kzg to implement the `bls_modular_inverse()` of [polynomial-commitments.md](https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/polynomial-commitments.md#bls_modular_inverse).

The documentation of the latter says `return 0 for x == 0`, but `fr_batch_inv()` would not have this behavior. In particular, due to using the batch inversion algorithm, if an input is zero it would also zero the entire output array and not just that index.

This should not be exploitable with the current protocol (since we are only inverting known values or RO outputs) but we should figure out how to handle this in a less suprirsing manner either way IMO.

Two approaches:
- Return `C_KZG_BADARGS` from `fr_batch_inv()` if an input is zero (not the spec's behavior, but at least it's not a surprising behavior if we document it)
- Skip zeroes during the batch inversion [like kev's gnark library is doing](https://github.com/ConsenSys/gnark-crypto/blob/master/ecc/bls12-381/fr/element.go#L687). This would satisfy the spec behavior IIUC, but it's slightly more complicated to implement. This is [also what arkworks is doing](https://github.com/arkworks-rs/algebra/blob/master/ff/src/fields/mod.rs#L340) fwiw.

(while we are at it, we should also see if we can ditch the `prod` array and only do it with a single field element like gnark is doing. or i suspect we might be doing an extra field mul in the second loop)
(while we are at it, we should also document that `fr_batch_inv` does not handle len==0)

ppopth:
What is the advantage of doing the first approach over the second? Following the spec is probably a good idea.

If we think returning an error is a good idea in c-kzg-4844, it's probably also a good idea to do it in the spec?

asn-d6:
Advantage of the first approach is simplicity but you are right that it doesn't make sense to diverge from the spec.
Let's go with second approach indeed.

asn-d6:
FWIW pushed a PoC here: https://github.com/asn-d6/c-kzg-4844/tree/batch_inv_zeroes_dev

It's using the gnark's approach. The performance seems to be about the same as the status quo, but it might be possible to optimize `fr_is_zero()` further by removing the use of `blst_uint64_from_fr()` and using the montgommery form directly.

dankrad:
I think I agree that failing on 0 inputs is the simpler and safest approach.

asn-d6:
[...]

PR #215, opened 2023-03-16
Simplify `fr_batch_inv` and reject zero inputs

asn-d6:
This implements the first approach from https://github.com/ethereum/c-kzg-4844/issues/205

### DIFF
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -255,6 +255,20 @@ static bool fr_is_one(const fr_t *p) {
 }
 
 /**
+ * Test whether the operand is zero in the finite field.
+ *
+ * @param[in] p The field element to be checked
+ *
+ * @retval true  The element is zero
+ * @retval false The element is not zero
+ */
+static bool fr_is_zero(const fr_t *p) {
+    uint64_t a[4];
+    blst_uint64_from_fr(a, p);
+    return a[0] == 0 && a[1] == 0 && a[2] == 0 && a[3] == 0;
+}
+
+/**
  * Test whether two field elements are equal.
  *
  * @param[in] aa The first element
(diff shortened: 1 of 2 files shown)


jtraglia (commented):
Nice, this is cleaner 👍

src/c_kzg_4844.c
```diff
@@ -332,42 +346,43 @@ static void fr_from_uint64(fr_t *out, uint64_t n) {
...
 
     assert(len > 0);
     assert(a != out);
 
-    ret = new_fr_array(&prod, len);
-    if (ret != C_KZG_OK) goto out;
+    fr_t accumulator = FR_ONE;
 
-    prod[0] = a[0];
+    for (i = 0; i < (int)len; i++) {
```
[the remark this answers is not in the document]
asn-d6: Addressed in 3dd65d7e3c0c994460eb9fde46dc01caad1181d7

ppopth:
Please add a remark that if `C_KZG_BADARGS` is returned, we did mutate `out`.

src/c_kzg_4844.c
```diff
@@ -254,6 +254,20 @@ static bool fr_is_one(const fr_t *p) {
...
+ *
+ * @param[in] p The field element to be checked
+ *
+ * @retval true  The element is zero
+ * @retval false The element is not zero
+ */
+static bool fr_is_zero(const fr_t *p) {
+    uint64_t a[4];
+    blst_uint64_from_fr(a, p);
+    return a[0] == 0 && a[1] == 0 && a[2] == 0 && a[3] == 0;
```
ppopth, line 268: Why not just compare it with `FR_ZERO`? like
```c
return fr_equal(p, &FR_ZERO);
```

ppopth: From another point of view, this function is not necessary at all, since you can use `fr_equal` with `FR_ZERO`, but I'm fine with either approach.

asn-d6: The idea was to avoid a second `blst_uint64_from_fr()` which is a moderately expensive operation.

asn-d6:
> Please add a remark that if `C_KZG_BADARGS` is returned, we did mutate `out`.

Good point.  Will do.

src/c_kzg_4844.c
ppopth: Thank makes sense

PR #218, opened 2023-03-16
Fix a mistake in python bindings makefile

jtraglia:
The expected file is `c_kzg_4844.o` (with underscores) not `c-kzg-4844.o` (with dashes).

This was causing the build/test to always happen. No biggie but should fix.


asn-d6:
lgtm!

PR #219, opened 2023-03-16
Fix some nits with nodejs

jtraglia:
* When I ran `make` it would always fail in the `clean` rule.
  * This is because there's a circular dependency.
  * I need to run `yarn install` to run `yarn node-gyp clean`.
  * Just remove that part since it's causing issues.
* Remove `node_modules` in `clean` rule.
* Update `.gitignore` with json trusted setup.
  * Every time I ran `make` it would reformat this file & show up in the status.  
* Reformat json trusted setup that nodejs wants done.

Build error I mentioned:

```
$ make
yarn node-gyp clean
yarn run v1.22.19
error Command "node-gyp" not found.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
make: *** [clean] Error 1
```


asn-d6:
lgtm!

PR #220, opened 2023-03-16
Speed up rust build just a little

jtraglia:
In the rust bindings build, it will `make all` which builds `c_kzg_4844.o` and runs all of the C tests.

This PR will call `make c_kzg_4844.o` instead, which should speed things up just a little.


asn-d6:
lgtm!

PR #223, opened 2023-03-17
changed VLA to an explicit c_kzg_malloc - call.

GottfriedHerold:
Variable length arrays are prone to cause (silent!) stack overflows and a potential cause of security vulns. While the usage pattern is probably safe in our context, changing to a more robust solution seems prudent.


asn-d6:
LGTM! Thanks!

I'll let @jtraglia take a look as well just in case.

jtraglia (approved):
I agree, this is better. Thanks 🥇 LGTM.

PR #224, opened 2023-03-17
Document that bit_reversal_permutation does not work for n==1.

GottfriedHerold:
Document that bit_reversal_permutation does not work with length 1. There is even a test to check this behaviour. Note that the length-1 case actually would work the given implementation; I do not know why this case is (intentionally) excluded.

Also, explicitly check that n != 0. This does not change anything, really, but previously, it was caught as a side-effect of implementation quirks of log2_pow2 and is_power_of_two, which is not good style.

### DIFF
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -590,12 +590,13 @@ static uint32_t reverse_bits(uint32_t n) {
  * @param[in,out] values The array, which is re-ordered in-place
  * @param[in]     size   The size in bytes of an element of the array
  * @param[in]     n      The length of the array, must be a power of two
- *                       less that 2^32
+ *                       less that 2^32 and unequal to 1.
  */
 static C_KZG_RET bit_reversal_permutation(
     void *values, size_t size, uint64_t n
 ) {
     CHECK(n >> 32 == 0);
+    CHECK(n != 0);
     CHECK(is_power_of_two(n));
     CHECK(log2_pow2(n) != 0);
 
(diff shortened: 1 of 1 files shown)


src/c_kzg_4844.c
```diff
@@ -590,12 +590,13 @@ static uint32_t reverse_bits(uint32_t n) {
  * @param[in,out] values The array, which is re-ordered in-place
  * @param[in]     size   The size in bytes of an element of the array
  * @param[in]     n      The length of the array, must be a power of two
- *                       less that 2^32
+ *                       less that 2^32 and unequal to 1.
```
jtraglia, line 593: I think it would be clearer to say this, since returns BADARGS for 0 too. Also fixes previous typo (that -> than).
```suggestion
 *                       greater than 1 and less than 2^32.
```

Or maybe even "between 2 and 2^32-1" but I'll leave that up to you.

src/c_kzg_4844.c
```diff
@@ -590,12 +590,13 @@ static uint32_t reverse_bits(uint32_t n) {
...
  */
 static C_KZG_RET bit_reversal_permutation(
     void *values, size_t size, uint64_t n
 ) {
     CHECK(n >> 32 == 0);
+    CHECK(n != 0);
```
jtraglia, line 599: Nit, could we swap the order of these? I prefer checks be in order of fast/simple -> slow/complex.

```suggestion
    CHECK(n != 0);
    CHECK(n >> 32 == 0);
```

jtraglia (changes requested):
I like these changes, I just have some nits.

GottfriedHerold:
accepted changes, but wrote strictly greater rather than greater for emphasis (as this is unexpected behaviour, so emphasis is better)

jtraglia (approved):
LGTM 👍

PR #231, opened 2023-03-20
nim bindings: free global KzgSettings before creating new one

jangko:


jtraglia:
This would allow `loadTrustedSetup` to be called twice, right? Personally, I feel like it should throw an error/exception instead, like the other bindings with a global context do:

https://github.com/ethereum/c-kzg-4844/blob/b7902b2300a9d23f947720f2467ec9a6afb17c9e/bindings/go/main.go#L60-L62

https://github.com/ethereum/c-kzg-4844/blob/b7902b2300a9d23f947720f2467ec9a6afb17c9e/bindings/java/c_kzg_4844_jni.c#L68-L72

https://github.com/ethereum/c-kzg-4844/blob/b7902b2300a9d23f947720f2467ec9a6afb17c9e/bindings/node.js/src/kzg.cxx#L136-L139

jangko:
I've decided to follow java bindings. Let user free loaded trusted setup and error if it's already loaded.

jtraglia (approved):
Looks great, thanks for making that change.

PR #232, opened 2023-03-21
nim bindings: add support for nimble package

jangko:
Nimble is Nim standard package manager, this PR aims to let user using Nimble to manage this package installation.
Only `kzg4844.nimble` file is added to root directory. The rest is contained in `bindings/nim` and the installation script in `kzg4844.nimble` will handle the rest to meet Nimble requirements.

### DIFF
--- a/kzg4844.nimble
+++ b/kzg4844.nimble
@@ -1,0 +1,45 @@
+mode = ScriptMode.Verbose
+
+##################################################
+# Package definition
+##################################################
+
+packageName   = "kzg4844"
+version       = "0.1.0"
+author        = "Andri Lim"
+description   = "Nim wrapper of c-kzg-4844"
+license       = "Apache License 2.0"
+skipDirs      = @[
+  "tests", "lib", "inc", "fuzz",
+  "bindings/csharp",
+  "bindings/go",
+  "bindings/java",
+  "bindings/node.js",
+  "bindings/python",
+  "bindings/rust"
+  ]
+installDirs   = @[
+  "blst",
+  "src",
+  "bindings/nim"
+  ]
+
+requires "nim >= 1.2.0",
+         "stew"
+
+##################################################
+# Test code
+##################################################
+
+import "bindings/nim/config.nims"
+
+task test, "Run all tests":
+  runAllTest()
+
+##################################################
+# Package installation code
+##################################################
+
+after install:
+  mvDir("bindings/nim/nimble", ".")
+  rmDir("bindings/nim/tests")
--- a/bindings/nim/README.md
+++ b/bindings/nim/README.md
@@ -14,7 +14,8 @@ nimble install stew
 
 ## Tests
 
-Currently tests only support Nim compiler version 1.4, and 1.6 because of yaml library limitations.
+Currently reference tests only support Nim compiler version 1.4, and 1.6 because of yaml library limitations.
+But other tests that is not using yaml can be run by Nim 1.2 - devel.
 
 Dependencies:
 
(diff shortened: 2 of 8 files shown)


kzg4844.nimble
```diff
@@ -0,0 +1,45 @@
...
+  "tests", "lib", "inc", "fuzz",
+  "bindings/csharp",
+  "bindings/go",
+  "bindings/java",
+  "bindings/node.js",
+  "bindings/python",
+  "bindings/rust"
+  ]
+installDirs   = @[
+  "blst",
```
jtraglia, line 22: With nimble, will this directory (submodule) be initialized? For Go, I had to copy the blst headers to here:

* https://github.com/ethereum/c-kzg-4844/tree/main/bindings/go/blst_headers

jangko: Yes, nimble can handle submodule directory. I've tested this package configuration, and it's works.
But other nimbus team member prefer that the `.nimble` file be put into `bindings/nim`, and those ugly proxy modules with import and reexport can be removed.
But sadly, putting `.nimble` file into `bindings/nim` doesn't work for the moment, because nimble can't install `blst` and `src` from `bindings/nim`.
So we have to ask nimble team to take a look at this issue.

jtraglia: That's really cool that nimble can handle that. And ah, I understand. Sounds good 👍

arnetheduck:
since projects are likely to use both "general" bls and c-kzg at the same time, I think it would make sense to export the raw blst wrapper to a separate `blst_abi` package which would be used by this wrapper and https://github.com/status-im/nim-blscurve - else we potentially end up with two competing versions of blst in the same project which is difficult to manage.

The `blst_abi` package can either be upstreamed to blst itself or live as a separate package similar to how we deal with https://github.com/status-im/nim-secp256k1/

jtraglia:
Hmm the `nimble install` command didn't work for me:

```
[jtraglia@laptop ~]$ nimble install https://github.com/ethereum/c-kzg-4844
Downloading https://github.com/ethereum/c-kzg-4844 using git
     Error: Could not find a file with a .nimble extension inside the specified directory: /tmp/nimble_17099/githubcom_ethereumckzg4844
```

jtraglia (approved):
LGTM 🥇

jtraglia:
Will wait to merge until you see Jacek's comment.

Also, when running the tests with `nimble test` from the project's root directory, I get this warning:

>    Warning: Package 'kzg4844' has an incorrect structure. It should contain a single directory hierarchy for source files, named 'kzg4844', but file 'test_all.nim' is in a directory named 'bindings/nim/tests' instead. This will be an error in the future.

I don't think its "solution" is good either, since we want to run the tests during installation & delete them afterwards. Its solution, to add the tests dir to the `skipDirs`, didn't make a difference when I tried it.
 
> Hint: If 'bindings/nim/tests' contains source files for building 'kzg4844', rename it to 'kzg4844'. Otherwise, prevent its installation by adding `skipDirs = @["bindings/nim/tests"]` to the .nimble file.

Is there anything that can be done about this?

jangko:
| Is there anything that can be done about this?

Nothing of I know. Nimble start with a good idea, the nim script is indeed flexible and powerful powering nimble, but nimble also made too much assumption and make the powerful nim script become less flexible.

About blst, I have no strong opinion, nim bindings is not directly using it, and Nim is flexible enough to choose between blst package.

arnetheduck:
> Nim is flexible enough to choose between blst package.

We've had the problem in the past where applications had two copies of the same library and were accidentally using the "older" one (with its associated security bugs) due to there being no clear strategy of which one is used when, when two are present.

PR #233, opened 2023-03-21
[Java binding] Rename ByteArrayTuple to ProofAndY

StefanBratanov:


StefanBratanov:
@jtraglia I think the interface is cleaner and more concise this way. What do you think?

asn-d6:
I like it. Cleaner!

jtraglia (approved):
I like this better too 🙂

PR #234, opened 2023-03-21
Short circuit batch verification if n==1

asn-d6:
Stole this idea from @kevaundray's golang codebase: https://github.com/crate-crypto/go-proto-danksharding-crypto/blob/15ed9c4701955ea47ad787fdccb9baba4ffc1687/internal/kzg/kzg_verify.go#L125

No strong opinion about whether want to do this. It gives an 8% or so speedup when batch is called with a single blob (because we don't do any mem allocation, or the compute powers fiat shamir stuff), which is something that might happen a lot if clients decide to just use batch for everything (better UX).

(/cc @jtraglia @ppopth @GottfriedHerold )

### DIFF
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -1494,6 +1494,13 @@ C_KZG_RET verify_blob_kzg_proof_batch(
         return C_KZG_OK;
     }
 
+    /* For a single blob, just do a regular single verification */
+    if (n == 1) {
+        return verify_blob_kzg_proof(
+            ok, &blobs[0], &commitments_bytes[0], &proofs_bytes[0], s
+        );
+    }
+
     /* We will need a bunch of arrays to store our objects... */
     ret = new_g1_array(&commitments_g1, n);
     if (ret != C_KZG_OK) goto out;


asn-d6:
> Seems fine. While I'm also in favor of this change (for clarity), note that a significant part of the speedup can also be gained by improving the batch functions. Notably:

Thanks for the feedback Gotti. It's really useful.

I'm personally in favor of most suggested changes, but given the proximity to the code freeze, I'm hesitant to commit to all of them. Here is some more detailed thoughts:

> 
>     * The memory allocation for the polynomial is unnecessary. We can reuse the same poly in the loop.
> 

Ugh. Good point. Let's do this for sure.

>     * The memory allocations can be merged from 5 (or 4 without the poly) into 1 for an array-of-struct containing commitment+proof+ev_challenge+ys.
>       We should then pass an array-of-structs rather than 4 same-sized arrays to verify_kzg_proof_batch. The same struct should be passed to verify_kzg_proof_impl.
>       This gives a better memory footprint and also allows code-reuse and to deduplicate the joint code of verify_blob_kzg_proof and verify_blob_kzg_proof_batch, so it actually _simplifies_ the code.
> 

I agree that this gives a better memory footprint and it's likely to allow code simplification. I'm hesitant to commit to this change because I feel like it will be a substantial refactoring (for this phase) and it might not give us enough time for review/testing before the freeze. Another reason, I'm slightly hesitant is because the current code structure looks much closer to the spec (which is nice for the auditor IMO).

That said, I would be happy to look at a PR for this.

>     * compute_powers can be optimized to only do n-1 multiplication rather than n.
> 

Agreed, altho one field mul is a few nanoseconds of time. Let's not do this for the sake of less changes.

>     * compute_r_powers does not need to compute hashes for n==1 or n==0.
>

`compute_r_powers()` should current never be called for `n==0` but agreed that it can be greatly simplified for `n==1`. I'm not totally sure how the refactoring is gonna look like for this one.
 
> 
> I can prepare a bunch of PRs if there is interest in these general improvements.
>

I think all of the above suggestions (apart from the last one)  are generally good ideas and orthogonal of this PR. How about we merge this PR (since we all seem to like it) and if we get to merge all the above PRs and we think that it's fine dropping this PR, we can revert it then?

Apart from the performance improvement of this PR, it's also nice for clarity so that's an independent factor to the above suggestions by Gotti.

ppopth:
LGTM

ppopth (approved):

ppopth:
> > ```
> > * The memory allocations can be merged from 5 (or 4 without the poly) into 1 for an array-of-struct containing commitment+proof+ev_challenge+ys.
> >   We should then pass an array-of-structs rather than 4 same-sized arrays to verify_kzg_proof_batch. The same struct should be passed to verify_kzg_proof_impl.
> >   This gives a better memory footprint and also allows code-reuse and to deduplicate the joint code of verify_blob_kzg_proof and verify_blob_kzg_proof_batch, so it actually _simplifies_ the code.
> > ```
> 
> 
> I agree that this gives a better memory footprint and it's likely to allow code simplification.
> 
> I'm hesitant to commit to this change because I feel like it will require substantial refactoring (for this late project phase) and it might not give us enough time for review/testing before the freeze.
> 
> Another reason, I'm slightly hesitant is because the current code structure looks much closer to the spec (which is nice for the auditor IMO). Perhaps the struct approach could also be made so that the function interfaces looks close to the spec as well.

I agree with @asn-d6 on this point. If it makes sense to do the struct approach in c-kzg-4844, it probably makes sense to do it in consensus-specs as well. I think if we really want to do this, we should change consensus-specs first and then change the c-kzg-4844 code.

ppopth:
> compute_powers can be optimized to only do n-1 multiplication rather than n.

As @jtraglia said, this requires spec change. I think it's a good idea to change the spec first before changing c-kzg-4844.

> compute_r_powers does not need to compute hashes for n==1 or n==0.

This not only changes the internal logic, but also change the expected output (test vectors), so let's change this in consensus-specs first.

PR #235, opened 2023-03-21
c_kzg_4844.c: In verify_blog_kzg_proof_batch, reuse temporary variable.

GottfriedHerold:
Polynomial is only used as a local temporary inside the loop.

There is no need to allocate an array of them.

See #234 

Note that this PR declares the variable inside the loop for clarity.

### DIFF
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -1486,7 +1486,7 @@ C_KZG_RET verify_blob_kzg_proof_batch(
     g1_t *proofs_g1 = NULL;
     fr_t *evaluation_challenges_fr = NULL;
     fr_t *ys_fr = NULL;
-    Polynomial *polynomials = NULL;
+
 
     /* Exit early if we are given zero blobs */
     if (n == 0) {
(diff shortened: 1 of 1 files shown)


src/c_kzg_4844.c
```diff
@@ -1513,16 +1510,18 @@ C_KZG_RET verify_blob_kzg_proof_batch(
         );
         if (ret != C_KZG_OK) goto out;
 
+        Polynomial polynomial;
```
asn-d6, line 1513: Nitpick: I think we tend to declare variables in the beginning of each scope (that is, right after the `for` line in this case). No strong opinion from my side.

asn-d6:
Looks great! Left a small nit that can also be ignored.

src/c_kzg_4844.c
jtraglia: Yeah, I agree with @asn-d6 here. Let's put that at the top of the loop, before `bytes_to_kzg_commitment`.

GottfriedHerold: No strong opinion here, did this mostly to convince myself that (re-)using the variable was safe (this way, further using it after the loop would not compile).

GottfriedHerold: Fixed.

asn-d6 (approved):
LGTM!

PR #236, opened 2023-03-21
Consolidate allocs and conversions

GottfriedHerold:
This PR draft is a reply to 

@asn-d6 "That said, I would be curious to see how a PR for this one would look like." from the discussion in #234.

Note that this PR-draft is based on #235 just to avoid conflicts, it completely removes the variable from the code anyway.
It may be helpful for review to look at the 2 commits 3ba09df and 9e5de9a separately.

Personally, I find the code actually cleaner that way (less duplication, clearer semantics), but I have no strong feelings about this PR draft.

### DIFF
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -1486,7 +1486,7 @@ C_KZG_RET verify_blob_kzg_proof_batch(
     g1_t *proofs_g1 = NULL;
     fr_t *evaluation_challenges_fr = NULL;
     fr_t *ys_fr = NULL;
-    Polynomial *polynomials = NULL;
+
 
     /* Exit early if we are given zero blobs */
     if (n == 0) {
@@ -1503,8 +1503,6 @@ C_KZG_RET verify_blob_kzg_proof_batch(
     if (ret != C_KZG_OK) goto out;
     ret = new_fr_array(&ys_fr, n);
     if (ret != C_KZG_OK) goto out;
-    ret = c_kzg_calloc((void **)&polynomials, n, sizeof(Polynomial));
-    if (ret != C_KZG_OK) goto out;
 
     for (size_t i = 0; i < n; i++) {
         /* Convert each commitment to a g1 point */
@@ -1513,8 +1511,10 @@ C_KZG_RET verify_blob_kzg_proof_batch(
         );
         if (ret != C_KZG_OK) goto out;
 
+        Polynomial polynomial;
+
         /* Convert each blob from bytes to a poly */
-        ret = blob_to_polynomial(&polynomials[i], &blobs[i]);
+        ret = blob_to_polynomial(&polynomial, &blobs[i]);
         if (ret != C_KZG_OK) goto out;
 
         compute_challenge(
@@ -1522,7 +1522,7 @@ C_KZG_RET verify_blob_kzg_proof_batch(
         );
 
         ret = evaluate_polynomial_in_evaluation_form(
-            &ys_fr[i], &polynomials[i], &evaluation_challenges_fr[i], s
+            &ys_fr[i], &polynomial, &evaluation_challenges_fr[i], s
         );
         if (ret != C_KZG_OK) goto out;
 
@@ -1539,7 +1539,6 @@ out:
     c_kzg_free(proofs_g1);
     c_kzg_free(evaluation_challenges_fr);
     c_kzg_free(ys_fr);
-    c_kzg_free(polynomials);
     return ret;
 }
 
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -1330,18 +1330,12 @@ C_KZG_RET verify_blob_kzg_proof(
  * Compute random linear combination challenge scalars for batch verification.
  *
  * @param[out]  r_powers_out   The output challenges
- * @param[in]   commitments_g1 The input commitments
- * @param[in]   zs_fr          The input evaluation points
- * @param[in]   ys_fr          The input evaluation results
- * @param[in]   proofs_g1      The input proofs
+ * @param[in]   claims         The input commitments, points, results and
+ * proofs.
+ * @param[in]   n              The length of the claims vector
  */
 static C_KZG_RET compute_r_powers(
-    fr_t *r_powers_out,
-    const g1_t *commitments_g1,
-    const fr_t *zs_fr,
-    const fr_t *ys_fr,
-    const g1_t *proofs_g1,
-    size_t n
+    fr_t *r_powers_out, const kzg_evaluation_claim *claims, size_t n
 ) {
     C_KZG_RET ret;
     uint8_t *bytes = NULL;
@@ -1372,19 +1366,19 @@ static C_KZG_RET compute_r_powers(
 
     for (size_t i = 0; i < n; i++) {
         /* Copy commitment */
-        bytes_from_g1((Bytes48 *)offset, &commitments_g1[i]);
+        bytes_from_g1((Bytes48 *)offset, &claims[i].commitment_g1);
         offset += BYTES_PER_COMMITMENT;
 
         /* Copy z */
-        bytes_from_bls_field((Bytes32 *)offset, &zs_fr[i]);
+        bytes_from_bls_field((Bytes32 *)offset, &claims[i].evaluation_point);
         offset += BYTES_PER_FIELD_ELEMENT;
 
         /* Copy y */
-        bytes_from_bls_field((Bytes32 *)offset, &ys_fr[i]);
+        bytes_from_bls_field((Bytes32 *)offset, &claims[i].y_fr);
         offset += BYTES_PER_FIELD_ELEMENT;
 
         /* Copy proof */
-        bytes_from_g1((Bytes48 *)offset, &proofs_g1[i]);
+        bytes_from_g1((Bytes48 *)offset, &claims[i].proof_g1);
         offset += BYTES_PER_PROOF;
     }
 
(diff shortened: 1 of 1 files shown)


GottfriedHerold:
Note that this does not actually touch the public API.

src/c_kzg_4844.c
```diff
@@ -1253,66 +1258,86 @@ C_KZG_RET compute_blob_kzg_proof(
...
 
 /**
- * Given a blob and its proof, verify that it corresponds to the provided
- * commitment.
- *
- * @param[out] ok               True if the proofs are valid, otherwise false
+ * Given a blob and its proof and commitment as bytes, compute a
+ * kzg_evaluation_claim by parsing these into group elements and computing the
+ * Fiat-Shamir challenge for verification. The resulting claim consists of a
+ * commitment to a polynomial p, an evaluation point, a puported evaluation
```
asn-d6, line 1264: puported -> claimed?

asn-d6:
Did an initial review here. Thanks for the code!

I'm pretty sure that the code is correct, but I'm debating whether the changes are worth incorporating or not.

I think while we are evaluating this PR, we can ignore the performance benefits of this approach as memory allocation is generally fast, and the speed difference between five allocs of total size N, versus one alloc of total size N is _probably_ on the order of nanoseconds, whereas we are dealing with millisecond ops here.

Focusing on the simplifications of this PR, I can definitely see how `make_claim_from_blob_kzg_proof()` allows us to avoid code duplication in the two functions where it's used. However that comes at the cost of introducing a new custom struct (not in the spec) and a new non-trivial function (not in the spec).

All in all, I think I don't have a strong opinion about these changes. I would opt to not merge this PR due to being close to the freeze but I can certainly be persuaded towards the other direction.

src/c_kzg_4844.c
```diff
@@ -990,14 +990,25 @@ C_KZG_RET blob_to_kzg_commitment(
...
+ * An instance of this struct models a claim (to be verified) that:
+ * proof_g1 is a proof for p(evaluation_point) == y_fr,
+ * where commitment_g1 is a commitment to the polynomial p.
+ */
+typedef struct {
+    g1_t commitment_g1;    /** Commitment to polynomial p*/
+    g1_t proof_g1;         /** Proof that p(evaluation_point) == y_fr*/
+    fr_t evaluation_point; /** point at which p is checked, typically determined
+                              as a hash output via Fiat-Shamir */
+    fr_t y_fr;             /** supposed value of p at the evaluation point */
```
jtraglia, line 1006: Some minor nits with the doc:
```suggestion
    g1_t commitment_g1;    /**< Commitment to polynomial p */
    g1_t proof_g1;         /**< Proof that p(evaluation_point) == y_fr */
    fr_t evaluation_point; /**< Point at which p is checked, typically determined
                              as a hash output via Fiat-Shamir */
    fr_t y_fr;             /**< Supposed value of p at the evaluation point */
```

src/c_kzg_4844.c
```diff
@@ -990,14 +990,25 @@ C_KZG_RET blob_to_kzg_commitment(
     return C_KZG_OK;
 }
 
+/**
+ * Struct holding the argument to verify_kzg_proof_impl.
+ * Having a struct holding these allows some code simpliciations
```
jtraglia, line 995: Minor nit, punctuation & typo.
```suggestion
 * Having a struct holding these allows some code simplifications.
```

jtraglia (commented):
My initial review.

jtraglia:
After thinking about it some more, I'm not really a fan of this idea. I made an alternative version, but I don't like it either. My main problem is that it differs from the spec too much. Unless there's a big performance improvement, I think I'd rather keep what we already have.

jtraglia:
And I don't see any noticeable performance improvements. The differences are within the margin of error.

#### Current
```
Benchmark/VerifyKZGProof-20              	    1174	   1020153 ns/op
Benchmark/VerifyBlobKZGProof-20          	     740	   1612213 ns/op
Benchmark/VerifyBlobKZGProofBatch(count=1)-20         	     714	   1677503 ns/op
Benchmark/VerifyBlobKZGProofBatch(count=2)-20         	     452	   2652104 ns/op
Benchmark/VerifyBlobKZGProofBatch(count=4)-20         	     261	   4589978 ns/op
Benchmark/VerifyBlobKZGProofBatch(count=8)-20         	     141	   8442185 ns/op
Benchmark/VerifyBlobKZGProofBatch(count=16)-20        	      72	  16187426 ns/op
Benchmark/VerifyBlobKZGProofBatch(count=32)-20        	      37	  31674580 ns/op
Benchmark/VerifyBlobKZGProofBatch(count=64)-20        	      18	  62640058 ns/op
```

#### Consolidated Allocations
```
Benchmark/VerifyKZGProof-20              	    1167	   1023479 ns/op
Benchmark/VerifyBlobKZGProof-20          	     741	   1607667 ns/op
Benchmark/VerifyBlobKZGProofBatch(count=1)-20         	     714	   1677561 ns/op
Benchmark/VerifyBlobKZGProofBatch(count=2)-20         	     452	   2647706 ns/op
Benchmark/VerifyBlobKZGProofBatch(count=4)-20         	     261	   4584247 ns/op
Benchmark/VerifyBlobKZGProofBatch(count=8)-20         	     141	   8461970 ns/op
Benchmark/VerifyBlobKZGProofBatch(count=16)-20        	      73	  16197958 ns/op
Benchmark/VerifyBlobKZGProofBatch(count=32)-20        	      37	  31807864 ns/op
Benchmark/VerifyBlobKZGProofBatch(count=64)-20        	      18	  62744236 ns/op
```

src/c_kzg_4844.c
```diff
@@ -990,14 +990,25 @@ C_KZG_RET blob_to_kzg_commitment(
...
+ *
+ * An instance of this struct models a claim (to be verified) that:
+ * proof_g1 is a proof for p(evaluation_point) == y_fr,
+ * where commitment_g1 is a commitment to the polynomial p.
+ */
+typedef struct {
```
ppopth, line 1001: It's better to provide the name of the struct here. Please look at https://stackoverflow.com/questions/1675351/typedef-struct-vs-struct-definitions

src/c_kzg_4844.c
```diff
@@ -990,14 +990,25 @@ C_KZG_RET blob_to_kzg_commitment(
...
+    g1_t commitment_g1;    /** Commitment to polynomial p*/
+    g1_t proof_g1;         /** Proof that p(evaluation_point) == y_fr*/
+    fr_t evaluation_point; /** point at which p is checked, typically determined
+                              as a hash output via Fiat-Shamir */
+    fr_t y_fr;             /** supposed value of p at the evaluation point */
+} kzg_evaluation_claim;
```
ppopth, line 1007: When I see `kzg_evaluation_claim`, I don't know if it's a variable name or a type name.

It's idiomatic to make all the type names have the suffix of `*_t`, like fr_t, g1_t. So it's better to have the type name of `kzg_evaluation_claim_t` (if we think `kzg_evaluation_claim` is a good name)

src/c_kzg_4844.c
```diff
@@ -990,14 +990,25 @@ C_KZG_RET blob_to_kzg_commitment(
     return C_KZG_OK;
 }
 
+/**
+ * Struct holding the argument to verify_kzg_proof_impl.
+ * Having a struct holding these allows some code simpliciations
+ *
+ * An instance of this struct models a claim (to be verified) that:
+ * proof_g1 is a proof for p(evaluation_point) == y_fr,
```
ppopth, line 998: If we want to change the name of `evaluation_point` to `z_fr` as @asn-d6 suggested, this has to be updated as well.

src/c_kzg_4844.c
```diff
@@ -1417,24 +1435,30 @@ static C_KZG_RET verify_kzg_proof_batch(
...
+    if (ret != C_KZG_OK) goto out;
 
     /* Compute the random lincomb challenges */
-    ret = compute_r_powers(
-        r_powers, commitments_g1, zs_fr, ys_fr, proofs_g1, n
-    );
+    ret = compute_r_powers(r_powers, claims, n);
     if (ret != C_KZG_OK) goto out;
 
+    /* copy the proofs into a single array, because g1_lincomb_naive requires
```
ppopth, line 1445: Let's capitalize the first letter of every comment as we are doing everywhere.
```suggestion
    /* Copy the proofs into a single array, because g1_lincomb_naive requires
```

src/c_kzg_4844.c
```diff
@@ -1253,66 +1258,86 @@ C_KZG_RET compute_blob_kzg_proof(
...
+ * @param[out] claim            Corresponding claim & proof about a polynomial
+ * identity
  * @param[in]  blob             Blob to verify
  * @param[in]  commitment_bytes Commitment to verify
  * @param[in]  proof_bytes      Proof used for verification
  * @param[in]  s                The trusted setup
  */
-C_KZG_RET verify_blob_kzg_proof(
-    bool *ok,
+C_KZG_RET make_claim_from_blob_kzg_proof(
```
ppopth, line 1274: 1. You don't need to tell what you make the claim from. People can tell from the parameter names
2. Reusing the verb from other functions also help make the code base conformant

So I think naming this function `compute_claim` makes more sense.

src/c_kzg_4844.c
ppopth, line 1264: I think it's better not to repeat what a claim consists of here because if we add a new field into the claim struct, we need to update the doc everywhere.

src/c_kzg_4844.c
```diff
@@ -1377,34 +1402,27 @@ static C_KZG_RET compute_r_powers(
...
  *     read off a length field in the protocol.
  *
  * @remark This function only works for `n > 0`.
  *
  * @param[out] ok             True if the proofs are valid, otherwise false
- * @param[in]  commitments_g1 Array of commitments to verify
- * @param[in]  zs_fr          Array of evaluation points for the KZG proofs
- * @param[in]  ys_fr          Array of evaluation results for the KZG proofs
- * @param[in]  proofs_g1      Array of proofs used for verification
+ * @param[in]  claims         Array of claims, each consisting of a commitment,
```
ppopth, line 1412: You don't need to repeat what a claim consists of here for the same reason I mentioned in the other comment.

PR #240, opened 2023-03-22
[Go] Rename CKzgRet to CKZGRet for consistency

jtraglia:
Rename this type to match the other types, which use all uppercase "KZG".

Also, make the type unsigned because there shouldn't be any negative values.


asn-d6:
LGTM!

PR #241, opened 2023-03-23
Fix minor memleak in Go bindings

jtraglia:
The allocated C strings that were being passed to `fopen` were never being freed. I thought these were casts.

* Call `defer free` for the two strings right after allocating them.
* Change mode from "rb" to "r" because the trusted setup shouldn't be binary.

### DIFF
--- a/bindings/go/main.go
+++ b/bindings/go/main.go
@@ -99,8 +99,8 @@ func LoadTrustedSetupFile(trustedSetupFile string) CKZGRet {
 	if fp == nil {
 		panic("error reading trusted setup")
 	}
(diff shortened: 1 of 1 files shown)


asn-d6:
Nice find. Maybe we suffer from a similar file descriptor leak in the nodejs bindings: https://github.com/ethereum/c-kzg-4844/blob/main/bindings/node.js/src/kzg.cxx#L142

Should we fix it as part of this PR or open a new one?

asn-d6:
Merged this one. Opened https://github.com/ethereum/c-kzg-4844/issues/246 for the nodejs issue.

PR #242, opened 2023-03-24
Multi-platform nodejs bindings

### LINKED ISSUE #118: npm package error
Is this repo the src for npm package? 

Error log: 
http://pastie.org/p/25xkd6j82tGM4x8gjQ4bMl

Cant get it installed on Ubuntu18, but I think this one flag it complains about -fPIC might solve the issue. on gcc11 seems to work fine.

jtraglia:
This PR is kind of a lot, but I think it's a major improvement. Should fix #118.

This was a joint effort with @matthewkeil. I took a lot of inspiration from his other PR.

* Update nodejs CI workflow to test all platforms.
* Merge `build` and `bundle` rules into one.
  * I think it makes sense to do these together.
* Merge `bindings.gyp` and `bindings.dist.gyp` into one file. 
  * To do this, the paths for all of the files must match.
  * Created temporary `dep` directory which is used for local builds.
    * This gets copied to the `dist` directory afterwards.
* Re-write `bindings.gyp` to not manually compile all of the deps.
  * This configures the input files & lets `node-gyp` assemble/compile.
  * Got this idea from the nim bindings & [blst](https://github.com/supranational/blst/blob/master/bindings/node.js/binding.gyp).
  * I was forced to do it this way, because of Windows.
      * It was complaining about blst being built with a different compiler.
* Modify C code to fix some (dumb) compiler warnings from MSVS.
  * It's not smart enough know multiplying two consts is still a const.
  * This also required I patch `blst_aux.h` to add a dummy field.
      * This is only done for nodejs & the dist.
  * These changes are alright though, better portability.
* Update the `Dockerfile` to run the reference tests.
  * The tests have to be temporarily copied to the bindings dir.
    * Docker cannot copy files from outside of this directory.
  * Simplified the files that get copied. We just need a few things.
* Remove `join` from the reference test paths.
  * The backslash in the middle was confusing `glob`.
* My IDE updated the CONTRIBUTING file automatically.
  * I do agree with these changes though.


bindings/node.js/binding.dist.gyp
```diff
@@ -22,30 +15,78 @@
...
+                "action_name": "build_ckzg",
+                "inputs": [
+                  "<(module_root_dir)/deps/c-kzg/c_kzg_4844.c",
+                  "<(module_root_dir)/blst.lib"
+                ],
+                "outputs": ["<(module_root_dir)/c_kzg_4844.o"],
+                "action": [
+                  "cc",
+                  "-I<(module_root_dir)/deps/blst/bindings",
+                  "-DFIELD_ELEMENTS_PER_BLOB=<!(echo ${FIELD_ELEMENTS_PER_BLOB:-4096})",
```
jtraglia, line 81: Oops I need to update this to be windows-style.

bindings/node.js/binding.gyp
```diff
@@ -2,27 +2,44 @@
...
+          ]
+        }],
+        ["OS=='mac'", {
+          "xcode_settings": {
+            "CLANG_CXX_LIBRARY": "libc++",
+            "MACOSX_DEPLOYMENT_TARGET": "13.0"
+          }
+        }],
+        ["OS=='win'", {
+          "defines": ["_CRT_SECURE_NO_WARNINGS"],
```
matthewkeil, line 34: Does this still need to be here if its in `src/Makefile` now?

jtraglia: Yes, I believe so. The makefile isn't being used in this process. I'll experiment with removing stuff later.

asn-d6:
Let's get a review from @matthewkeil here before merging?

bindings/node.js/Makefile
```diff
@@ -1,40 +1,49 @@
...
+	rm -rf deps
 	rm -f *.node
 	rm -f *.a
 	rm -f *.o
 	rm -rf node_modules
+	rm -rf ref-tests
 
-build: src/kzg.cxx lib/kzg.ts package.json binding.gyp Makefile
-	cd ../../src && make c_kzg_4844.o && cp c_kzg_4844.o ../bindings/node.js
+build: clean src/kzg.cxx lib/kzg.ts package.json binding.gyp Makefile
```
matthewkeil, line 13: I am not sure that I agree with merging the `build` and `bundle` from a JS/TS devEx perspective.

Generally in JS/TS repos there are a few distinct phases for development 
- install
- build/test
- bundle/publish

The install process for node dependencies (`node_modules` folder) is long running and mostly designed to be run once.  The build process is the one that gets iteratively run on file changes (usually there is a watcher monitoring for file changes and rerunning build/test).  Bundle/publish happens infrequently and generally by CI workflows triggered by git tags.

The idea behind structuring them that way is the "build" step is non-existent for JS and the code gets run through the tests with each file save.  For TS, the code is transpiled (in-memory) and then run through the tests similarly.  Because the process happens constantly, to minimize regression, it's ideally as fast as possible.

Native code is a bit different as there is a distinct "build" phase but it's relatively quick for this library. In practice, during active development, I have `chokidar` installed globally, run via a watch script that is not committed, so that I can see when things break with each keystroke.

We are using the publish command on CI to validate cross compatibility.  This is ideal for CI validation before merging but not great during active development.  It is much faster to just rerun node-gyp than to wait for all the node dependencies to be installed on file save.

As a workflow I would love to be able to have `build` running on a loop and then when I'm done with wholesale changes to run the container with `make publish; cd dist; yarn` once to make sure any changes are cross-compatible

As a suggestion, I have copied a Makefile below but please feel free to ignore or use only parts that you like 
```
all: install build format test bundle

# cleans typescript dependencies
clean-install:
	rm -rf node_modules

# installs typescript dependencies
install: clean-install
	yarn install --ignore-scripts

# Cleans native dependency, bindings and typescript artifacts
clean:
	rm -rf build
	rm -rf dist
	rm -rf deps
	rm -f *.node
	rm -f *.a
	rm -f *.o
	rm -rf ref-tests

# Cleans and rebuilds native dependencies, bindings and typescript wrapper
build: clean src/kzg.cxx lib/kzg.ts package.json tsconfig.build.json binding.gyp Makefile
	# Prepare the dependencies directory
	mkdir -p deps/c-kzg
	cp -r ../../blst deps
	cp ../../src/c_kzg_4844.c deps/c-kzg
	cp ../../src/c_kzg_4844.h deps/c-kzg
	# Patch the blst_aux.h to fix a compiler error
	awk '{gsub(/typedef struct \{\} blst_uniq;/, "typedef struct { int dummy; } blst_uniq;")}1' \
		deps/blst/bindings/blst_aux.h > blst_aux_temp.h
	mv blst_aux_temp.h deps/blst/bindings/blst_aux.h
	# Build the bindings
	yarn node-gyp configure
	yarn node-gyp build
	yarn tsc -p tsconfig.build.json

# Bundle the distribution, also helpful for cross compatibility checks
bundle:
	mv deps dist
	cp -r src dist/src
	cp README.md dist/README.md
	cp package.json dist/package.json
	cp binding.gyp dist/binding.gyp

# run unit tests and ref-tests
test:
	yarn jest

# lint js/ts code
format:
	yarn prettier --write .

# publish package to npm. requires an auth token
publish: build
	cd dist
	npm publish

# run ref-tests in linux environment for cross-compatability check
linux-test: build
	# Docker cannot copy from outside this dir
	cp -r ../../tests ref-tests
	docker build -t "linux-test" .
	docker logs --follow `docker run -d linux-test`
	rm -rf ref-tests
```

jtraglia: Sounds good, I will split those up again. Your reasoning makes sense.

.github/workflows/nodejs-tests.yml
```diff
@@ -8,13 +8,28 @@ on:
...
+  test:
+    runs-on: ${{matrix.os}}
+    strategy:
+      matrix:
+        os:
+          - ubuntu-latest
+          - macos-latest
+          - windows-latest
+        node:
+          - 16
```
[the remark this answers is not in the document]
jtraglia: Hmm. I'm not sure it's worth it considering there's only one month of (security) maintenance left. I'm fairly surprised that nodejs 16 is almost in the same boat; only 5 months left of security maintenance.

Useful link for reference: https://endoflife.date/nodejs

bindings/node.js/Makefile
jtraglia: Made these changes, with some alterations.

src/c_kzg_4844.h
```diff
@@ -37,15 +37,15 @@ extern "C" {
...
 #endif // FIELD_ELEMENTS_PER_BLOB
 /**
  * There are only 1<<32 2-adic roots of unity in the field, limiting the
  * possible values of FIELD_ELEMENTS_PER_BLOB. The restriction to 1<<31 is a
  * current implementation limitation. Notably, the size of the FFT setup would
- * overflow uint32_t, which would casues issues.
+ * overflow uint32_t, which would cause issues.
  */
-#if ((FIELD_ELEMENTS_PER_BLOB) <= 0) || ((FIELD_ELEMENTS_PER_BLOB) > (1 << 31))
+#if (FIELD_ELEMENTS_PER_BLOB <= 0) || (FIELD_ELEMENTS_PER_BLOB > (1UL << 31))
```
jtraglia, line 48: I made this change because `(1 << 31)` is undefined behavior. This could be done in a different PR too.

matthewkeil: Very very good catch!

bindings/node.js/Makefile
```diff
@@ -1,40 +1,78 @@
-all: clean build format test bundle
+# For a less verbose yarn.
+YARN = yarn --silent
 
+.PHONY: all
+all: format build test bundle
```
matthewkeil, line 5: Random question, and also something I do not feel strongly about as a note.  @jtraglia, why not run `install` here instead of in the `build` and `test` commands?  If it is in `all`, running `make` runs through the whole process once for ease of high-level/CI review (run one command and it all "just works") and then does not run the install again on iterative builds and tests during active development to speed up the process (`make test` running on a watch process that monitors file changes)

bindings/node.js/Makefile
```diff
@@ -1,40 +1,78 @@
...
+	@mv blst_aux_temp.h deps/blst/bindings/blst_aux.h
+	@# Build the bindings
+	@$(YARN) node-gyp --loglevel=warn configure
+	@$(YARN) node-gyp --loglevel=warn build
+	@$(YARN) tsc -p tsconfig.build.json
+
+# Bundle the distribution, also helpful for cross compatibility checks
+.PHONY: bundle
+bundle: build
+	@mv deps dist
```
matthewkeil, line 48: Would it be ok to `cp` here instead of `mv` so that subsequent runs of `make test` when updating the test file don't require `build` to be run again?

jtraglia: So `build` is phony (meaning it's not backed by a file & will run every time) and `build` will run `clean` first anyway, so those files won't exist. But a `cp` would work here too. It should have the same result.

bindings/node.js/Makefile
jtraglia: So `install` is a dependency of anything that requires `yarn`. Running `test` will run `install` first. It should "just work" as describe. See here, `install` is listed after `test`.

https://github.com/ethereum/c-kzg-4844/blob/c7b5e32c56524ddcfba106225f2cd098bb4c5e6e/bindings/node.js/Makefile#L55-L56

As for not including `install` in `all`, it will be called anyway so it's just a little simpler without.

matthewkeil (approved):

jtraglia:
@matthewkeil, thanks for the review 😄 merged.

PR #243, opened 2023-03-24
Reproducible rust bindings

divagant-martian:
Makes bindings generated for rust use binding to be able to update them easily

PR description TBD


bindings/rust/src/bindings/generated.rs
```diff
@@ -1,72 +1,73 @@
...
-type g2_t = blst_p2;
-type fr_t = blst_fr;
+pub type g1_t = blst_p1;
+pub type g2_t = blst_p2;
+pub type fr_t = blst_fr;
+#[doc = " Internal representation of a polynomial."]
 #[repr(C)]
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct Polynomial {
```
asn-d6: Hmm, `Polynomial` is not used by any of the public methods. I guess bindgen caught it because it's in `.h`, but I think it should be moved to `.c`.

(/cc @jtraglia )

bindings/rust/src/bindings/generated.rs
```diff
@@ -84,32 +85,32 @@ pub enum C_KZG_RET {
...
+    fs: *mut FFTSettings,
+    #[doc = " G1 group elements from the trusted setup,\n in Lagrange form bit-reversal permutation."]
+    g1_values: *mut g1_t,
+    #[doc = " G2 group elements from the trusted setup;\n both arrays have `FIELD_ELEMENTS_PER_BLOB` elements."]
+    g2_values: *mut g2_t,
 }
 extern "C" {
     pub fn load_trusted_setup(
         out: *mut KZGSettings,
-        g1_bytes: *const u8, /* n1 * 48 bytes */
```
asn-d6: I guess these were helper comments that will no longer be around. I guess we are fine with this.

bindings/rust/src/bindings/mod.rs
```diff
@@ -19,6 +19,20 @@ pub const BYTES_PER_G2_POINT: usize = 96;
 /// 65 is fixed and is used for providing multiproofs up to 64 field elements.
 const NUM_G2_POINTS: usize = 65;
 
+/// A trusted (valid) KZG commitment.
```
asn-d6, line 22: Hmm, can you explain a bit more how this works? We have `KZGCommitment` (in `bindings.rs`) and `KzgCommitment` here? When to use one and when the other? How to avoid confusion?

bindings/rust/src/bindings/mod.rs
```diff
@@ -184,10 +198,17 @@ impl KZGProof {
...
-        let mut bytes = [0; 48];
-        bytes.copy_from_slice(&self.bytes);
-        Bytes48 { bytes }
+    pub fn to_binding_bytes(&self) -> Bytes48 {
+        Bytes48 { bytes: self.bytes }
+    }
+
+    // NOTE: this function is not public to avoid easy conversions between KzgCommitment and
+    //       KzgProof that go via the [`Bytes48`] type. For the same reason, there is no `From`
+    //       implementation between the types.
```
asn-d6, line 207: Nice doc.

Can you also please document what these two functions do? I'm comfortable with `to_bytes()` functions having no docs, but `to_binding_bytes()` is adding additional context that can be confusing.

asn-d6:
Hello. I really like the general direction of this PR; that is, generating the bindings directly using bindgen. I left some review comments on the PR.

---

That said, I feel like this PR is also doing some other stuff? It's adding another type `KzgCommitment` (along with `KZGCommitment`) and using that as its public type? Or something along those lines?

Should this be done as part of this PR? Or should we first merge the bindgen stuff, and then open a new PR for the new type? Or the new type is necessary for bindgen to work?

I guess some more documentation would be helpful to dispell these questions.

---

Finally, I guess a review from @pawanjay176 would also be useful here.

bindings/rust/src/bindings/mod.rs
```diff
@@ -210,7 +231,8 @@ impl KZGProof {
                 kzg_settings,
             );
             if let C_KZG_RET::C_KZG_OK = res {
-                Ok((kzg_proof.assume_init(), y_out.assume_init()))
+                let kzg_proof = KzgProof::from_binding_bytes(kzg_proof.assume_init());
```
asn-d6, line 234: Hmm, I was surprised to see changes to the core bindings code. Also, these changes were completely hidden because of the directory restructuring in later commits.

divagant-martian:
Thanks for the fast review. I opened this PR too late in my TZ and too tired to provide a more complete explanation. 

So first thing to notice is that the current  bindings are a heavily edited version of the actually generated bindings. The first bindings version actually looked like this: [commit within original pr](https://github.com/ethereum/c-kzg-4844/pull/4/commits/bf5c1da96777fa0e0ca5c264d032abb3c08ac46f). The difference between the current bindings and the ones in this PR are better viewed commit by commit in this pr (this is the way I intended this PR to be read, sorry for the lack of instructions there). The important differences are highlighted in [this commit](https://github.com/ethereum/c-kzg-4844/pull/243/commits/7460410bc4be9018c178356bf470e2def23f9896#diff-26022be6c44ddc3ad438e417d1d7514f79a62b76614d9ed7e4c00dce470d5d6a) 

As you can see, the differences are
- types were private, now public. This change was made in the current bindings by hand in [this commit](https://github.com/ethereum/c-kzg-4844/pull/62/files#diff-78831f18c3692b90d974705618bd9c68ea68d06073ece757e9f3d40c4b253373)
- `const` vs `mut` pointer were changed by hand [here](https://github.com/ethereum/c-kzg-4844/pull/4/commits/5c767affe4ed7f6f683bb98869d43fb77c7c42da)
- constants used in definitions vs integer literals. This change was made by hand [here](https://github.com/ethereum/c-kzg-4844/pull/4/commits/f703be41c2e2952cddcea08da01e5f499c2ea9c0)
- KZGCommitment and KZGProof as structs, not type aliases: This change was made by hand in [this PR](https://github.com/ethereum/c-kzg-4844/pull/62/files#diff-78831f18c3692b90d974705618bd9c68ea68d06073ece757e9f3d40c4b253373)
This one might be hard to spot so I'm pasting a screenshot
image

Now this last change is the only one that I could not make using bindgen. The rest of the by-hand cleanup are now simple bindgen instructions. I agree that the change is necessary to help with type safety. This is the reason why I introduced the types `KzgProof ` and `KzgCommitment`. These are the publicly exposed types and are the only notable difference between the current and proposed bindings. There is no way to mistake them outside the crate simply because they are not exposed, and there is no way to mistake them inside the crate because the compiler will complain loudly.

I'll address the rest of your comments individually, but I hope this help to better explain and understand the proposed changes

bindings/rust/src/bindings/mod.rs
divagant-martian: Late for this suggestion, but this is PR is intended to be read commit by commit since, as you can tell, the final diff is unhelpful. The generated bindings differ mostly because the current ones are heavily edited by hand

bindings/rust/src/bindings/mod.rs
divagant-martian: yeah this is a weird state. In principle I would like to not expose the `Bytes48` type but the facade API already does. So I changed this function's name and made it tests only but then realized it would be needed to satisfy the public API. Happy to revert this change and discuss the possibility of removing `Bytes48` from the public API in a different PR

bindings/rust/src/bindings/mod.rs
divagant-martian: `KzgCommitment`  will be used in the public API, and `KZGCommitment` in the FFI boundary. Fortunately the compiler will complain loudly is used one instead of the other and only us need to deal with the existence of both types, not the final user, since the `KZG` one is not exposed

bindings/rust/src/bindings/generated.rs
divagant-martian: I think this is due to some change in bindgen that now they do not get generated. I was also surprised by this, but seems an acceptable change to me

bindings/rust/src/bindings/generated.rs
divagant-martian: I noticed this too. I had originally instructed bindgen to blocklist the type but left it here to highlight what you just noticed

jtraglia: Ah, yes. The `Polynomial` type should be moved back to the `.c` file. Good idea.

bindings/rust/src/bindings/generated.rs
jtraglia: Yeah this is fine with me too.

bindings/rust/src/bindings/generated.rs
jtraglia: Made a PR for this here:
* https://github.com/ethereum/c-kzg-4844/pull/244

bindings/rust/Cargo.toml
```diff
@@ -21,6 +21,9 @@ glob = "0.3.1"
 rand = "0.8.5"
 serde_yaml = "0.9.17"
 
+[build-dependencies]
+bindgen = { git = "https://github.com/rust-lang/rust-bindgen" , rev = "0de11f0a521611ac8738b7b01d19dddaf3899e66" }
```
pawanjay176, line 25: Any specific features we are using from this particular commit?

bindings/rust/src/bindings/mod.rs
pawanjay176: I'm also strongly in favour of using specific types for commitments and proofs(`KzgProof` and `KzgCommitment` introduced in this PR) instead of using the common `Bytes48` for the public interface to the bindings. 
This would give us higher type safety where we don't accidentally replace a proof with a commitment when passing values to the public functions. 
@asn-d6  I remember a conversation sometime back about using `Bytes48` to have it consistent with the other bindings. Not sure if that is still relevant somehow.

pawanjay176 (commented):
I really like this PR. I was unsuccessful in my first attempt to auto generate the bindings with bindgen and consequently didn't explore this further. I think @divagant-martian has addressed most of the issues I faced originally.

re splitting the changes to introduce `KzgCommitment` and `KzgProof` types, I'm happy to help with making another PR for that :)

bindings/rust/Cargo.toml
divagant-martian: but this should be release soon

divagant-martian:
Thanks @pawanjay176 . I guess I didn't directly respond to @asn-d6 's question about whether this is necessary or can be addressed in another PR. This is necessary. Since the `KZG*` types are type aliases, a good part of the functions we implement over them are duplicated. To make that work within this PR without introducing the new types, we would need to remove functions like `to_bytes`, `from_hex` etc to prevent the duplicates. We would then re-add the functions in a new PR. I find the change of introducing the types here less disruptive since they align with the state of the current bindings

bindings/rust/src/bindings/generated.rs
asn-d6: Just merged #244. Would be great if you could re-run bindgen for this PR please.

bindings/rust/src/bindings/mod.rs
asn-d6: Sounds good! Will review tomorrow more deeply.

bindings/rust/build.rs
```diff
@@ -72,20 +72,98 @@ fn main() {
...
+         */
+        // Inject the constant as C code so that the C compiler can use it.
+        // -D is not supported by bindgen https://github.com/rust-lang/rust-bindgen/issues/2394
+        .header_contents(
+            "consts",
+            &format!("#define FIELD_ELEMENTS_PER_BLOB {field_elements_per_blob}"),
+        )
+        .header(header_file)
+        .clang_args([format!("-I{inc_dir}")])
+        // Since this is not part of the header file, needs to be allowed explicitely.
```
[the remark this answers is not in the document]
divagant-martian: This was addressed in https://github.com/ethereum/c-kzg-4844/pull/243/commits/a66fef3136b258d5538a3e131b05e724a1df84f2

divagant-martian:
> Are there better names that can be given to those two types? So that the difference is more clear and does not hide into the capitalization?

We can so something very explicitly like adding them as `PublicApiKzgCommitment` and re-export them with the nice `KzgCommitment` to make development easier. It's to note tho, that if this crate is published, an user that goes to the documentation of `KzgCommitment`, they will land in a page that has the explicit (but ugly) `PublicApiKzgCommitment` name.

Another option is to internally alias `KZGCommitment` to `BindingKzgCommitment` and use this name instead of the upper case , easy to mistake one here in the crate. This sounds like a better solution to me.

> If no better names exist, can we please thoroughly document these types and their difference?

I can try to improve docs on this whatever way we go. Not sure how but probably something can be done

What do you guys think is best? What's your preference

divagant-martian:
I just updated this, blocklisting the type aliases and replacing them with custom implementations, which makes the bindings in this PR even closer to the current ones, and removed the existence of two similarly names types

bindings/rust/src/bindings/mod.rs
asn-d6: `Bytes48` signifies an untrusted bytes object that might happen to be a valid KZG proof. Whereas `KZGProof` is a valid KZG proof returned by the KZG library (or used internally in the KZG library).

The distinction between the two types is important to preserve.

I understand that `Bytes48` is not strongly typed and maybe you would have liked to be `Bytes48Proof` and `Bytes48Commitment`.  If this would be useful to lighthouse let's discuss it. But preferably in another PR (and preferably soon since we are getting ready to freeze for the audit).

asn-d6:
> I just updated this, blocklisting the type aliases and replacing them with custom implementations, which makes the bindings in this PR even closer to the current ones, and removed the existence of two similarly names types.
> 
> Let me know if this is something that works better

Honestly this is great. IMO this new approach is much more tidy. I guess the only downside is that we are being a bit naughty doing `repr(C)` stuff outside of the FFI boundary.

I'm gonna go ahead and ACK this. Let's merge this early next week unless there any new concerns get generated.

Thanks a lot @divagant-martian for being meticulous here :)

asn-d6 (approved):

jtraglia (approved):
This looks great! I don't have any critiques. LGTM :+1:

PR #247, opened 2023-03-24
[Java binding] Avoid unnecessary array copying in tests

StefanBratanov:
A bit of a nit really...

Basically `toArray()` copies the whole underlying array to a new one. `toArrayUnsafe()` just uses the underlying array. Given we don't modify the inputs in the binding, It is ok to use `toArrayUnsafe()` and is a test code as well. Only did it for blobs since they could be heavy in relation to other inputs.


jtraglia (approved):
I support fixing nits. Also removed that unused function, I see. LGTM 👍

Issue #246, opened 2023-03-24
File descriptor leak in the nodejs bindings


asn-d6:
Nice find. Maybe we suffer from a similar file descriptor leak in the nodejs bindings: https://github.com/ethereum/c-kzg-4844/blob/main/bindings/node.js/src/kzg.cxx#L142

Should we fix it as part of this PR or open a new one?

_Originally posted by @asn-d6 in https://github.com/ethereum/c-kzg-4844/issues/241#issuecomment-1482342386_

asn-d6:
Handled  in #248.

PR #248, opened 2023-03-25
close file descriptor

matthewkeil:
This PR closes the leaked file descriptor in node.js bindings.

Relates to #246

### DIFF
--- a/bindings/node.js/src/kzg.cxx
+++ b/bindings/node.js/src/kzg.cxx
@@ -149,7 +149,10 @@ Napi::Value LoadTrustedSetup(const Napi::CallbackInfo& info) {
       return env.Undefined();
   }
   data->is_setup = true;
-  fclose(file_handle);
+  int close_err = fclose(file_handle);
+  if (close_err != 0) {
+      Napi::Error::New(env, "Error closing trusted setup file: " + file_path).ThrowAsJavaScriptException();
+  }
   return env.Undefined();
 }
 
--- a/bindings/node.js/src/kzg.cxx
+++ b/bindings/node.js/src/kzg.cxx
@@ -150,6 +150,7 @@ Napi::Value LoadTrustedSetup(const Napi::CallbackInfo& info) {
   }
   int close_err = fclose(file_handle);
   if (close_err != 0) {
+      free_trusted_setup(&(data->settings)); 
       Napi::Error::New(env, "Error closing trusted setup file: " + file_path).ThrowAsJavaScriptException();
       return env.Undefined();
   }
--- a/bindings/node.js/src/kzg.cxx
+++ b/bindings/node.js/src/kzg.cxx
@@ -1,9 +1,6 @@
-#include <inttypes.h>
 #include <stdio.h>
 #include <iostream>
 #include <sstream>  // std::ostringstream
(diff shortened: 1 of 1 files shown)


bindings/node.js/src/kzg.cxx
```diff
@@ -149,6 +149,10 @@ Napi::Value LoadTrustedSetup(const Napi::CallbackInfo& info) {
       return env.Undefined();
   }
   data->is_setup = true;
+  int close_err = fclose(file_handle);
+  if (close_err != 0) {
+      Napi::Error::New(env, "Error closing trusted setup file: " + file_path).ThrowAsJavaScriptException();
+  }
```
asn-d6, line 155:
```suggestion
  if (close_err != 0) {
      Napi::Error::New(env, "Error closing trusted setup file: " + file_path).ThrowAsJavaScriptException();
      return env.Undefined();
  }
```

Let's complete this error check correctly with a return value so that future changes to this function are not prone to bugs.

matthewkeil: Good idea.  I also moved the `is_setup` to just before the final return so it will "not be setup" when it returns early

matthewkeil:
If `fclose` fails the setup is `free`'d where the error is thrown not in the cleanup function for clarity

asn-d6:
@jtraglia i will wait for you to merge this one

jtraglia:
One problem, the file descriptor is still leaked if the call to `load_trusted_setup_file` fails.

jtraglia:
I believe that would work, but I would do it a little differently. I think this is simpler:

```cpp
Napi::Value LoadTrustedSetup(const Napi::CallbackInfo& info) {
  Napi::Env env = info.Env();

  // Check if the trusted setup is already loaded
  KzgAddonData *data = env.GetInstanceData<KzgAddonData>();
  if (data->is_setup) {
    Napi::Error::New(env, "Error trusted setup is already loaded").ThrowAsJavaScriptException();
    return env.Undefined();
  }

  // Open the trusted setup file
  std::string file_path = info[0].As<Napi::String>().Utf8Value();
  FILE *file_handle = fopen(file_path.c_str(), "r");
  if (file_handle == nullptr) {
    Napi::Error::New(env, "Error opening trusted setup file: " + file_path).ThrowAsJavaScriptException();
    return env.Undefined();
  }

  // Load the trusted setup from that file
  C_KZG_RET ret = load_trusted_setup_file(&(data->settings), file_handle);

  // Close the trusted setup file
  if (fclose(file_handle) != 0) {
    if (ret == C_KZG_OK) {
      free_trusted_setup(&(data->settings));
    }
    Napi::Error::New(env, "Error closing trusted setup file").ThrowAsJavaScriptException();
    return env.Undefined();
  }

  // Check that it was successful
  if (ret != C_KZG_OK) {
    Napi::Error::New(env, "Error loading trusted setup file: " + file_path).ThrowAsJavaScriptException();
    return env.Undefined();
  }

  data->is_setup = true;
  return env.Undefined();
}
```

matthewkeil:
> I believe that would work, but I would do it a little differently. I think this is simpler:

You are a gentleman and a scholar sir!

jtraglia:
LGTM 👍

Issue #226, opened 2023-03-17
Test hard-wired to 4096, no check for FIELD_ELEMENTS_PER_BLOB not a power of 2.


GottfriedHerold:
The current tests run against the (mainnet) trusted setup, which has a value of FIELD_ELEMENTS_PER_BLOB of 4096 and a correspondingly sized trusted setup.

However, the library actually takes FIELD_ELEMENTS_PER_BLOB as an external input (defined in the Makefile as a compile-flag). This suggest that the library should also work for a different value of FIELD_ELEMENTS_PER_BLOB. In particular, the way at least parts of the code are written suggest that the library should also work for FIELD_ELEMENTS_PER_BLOB *not* a power of 2.

With the test setup as is, it is not possible to check that (if it is even true in the first place).

A possible fix would be for the tests to create their own mock (un-)trusted setup of the correct size. (Such a mock trusted setup would basically choose a *known* tau rather than powers-of-tau ceremony-- note that this would not be secure in production).

asn-d6:
Hmmm. I don't think that working with arbitrary values of `FIELD_ELEMENTS_PER_BLOB` was ever a design goal of c-kzg. Given the amount of FFTs and barycentric formulas happening (which depend on a multiplicative subgroup) I'm not sure if this would work with a non-power-of-two.

If anything, I would check at compile time that `FIELD_ELEMENTS_PER_BLOB` is a power of two below 2^31 or something.

asn-d6:
Hmm, not sure what should be done about this ticket.

I don't really have a strong opinion of how this ticket should be handled, apart from ideally that it shouldn't change the core library code. @GottfriedHerold could you take it from here? Thanks!

PR #249, opened 2023-03-27
Only support FIELD_ELEMENTS_PER_BLOB values that are a power of 2.

GottfriedHerold:
We do not want to support or maintain this case, so let us be explicit about it.

This should solve #226


ppopth:
LGTM

ppopth (approved):

PR #252, opened 2023-03-28
Replace CKZGRet with error in go bindings

jtraglia:
Fixing another nit. I think using `error` would be more Go-like than `CKZGRet.C_KZG_*`.

I will update the example in another PR. A little complicated to do in this one.

### DIFF
--- a/bindings/go/main.go
+++ b/bindings/go/main.go
@@ -8,6 +8,7 @@ import "C"
 
 import (
 	"errors"
+	"fmt"
 	"unsafe"
 
 	// So its functions are available during compilation.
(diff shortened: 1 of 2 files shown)


jtraglia:
I will update the example in another PR. A little complicated to do in this one.

asn-d6:
I don't know enough golang to know if this is more idiomatic, however I was kinda used to the old approach, and I feel like the new approach is slightly more complicated. I have no strong opinion here.

@GottfriedHerold any thoughts from the golang side?

GottfriedHerold:
From my side: Using an error as done by this PR is much more idiomatic from the Golang side, so I'm in favor of this change.
Regarding complexity, this is always a very subjective thing. My gut feeling is that Go using a C-style interface adds a lot of conceptual complexity to the (Go) library user, because it's very unusual and a well-written Go program would have to do essentially what this PR does on the client side. It's better to do this at the library side.

Two nits: Firstly, I would prefer a function makeErrorFromErrorCode (or whatever name) and do makeErrorFromErrorCode(ret) rather than errorMap[ret]

roughly:
func makeErrorFromErrorCode(errorCode C.C_KZG_RET) error{
  err, ok := errorMap[errorCode]
  if !ok{
     panic("Unexpected error code")
  }
  return err
}

The reason is that errorMap[ret] will return nil (i.e. no error) if ret is not in the domain of the map (i.e. if it encounters an unknown error code). This is not robust and would lead to a silent bug if we ever add another error code and don't update the Go bindings. makeErrorFromErrorCode(ret) should panic if the error code is not known. This makes it much easier to see that the function "does the correct thing".

Secondly, I would prefer renaming ret -> errorCode (ret suggests that this is returned as-is), which is more precise.

jtraglia:
Good ideas, will update the PR soon.

jtraglia:
Hey @GottfriedHerold how does it look now?

jtraglia:
> Secondly, I would prefer renaming ret -> errorCode (ret suggests that this is returned as-is), which is more precise.

To address this, I would prefer to keep it as `ret` which matches the other bindings.

PR #253, opened 2023-03-28
Ignore generated snapshots in rust bindings

jtraglia:
After building/testing the rust bindings on my system, there was a new untracked file in `git status`.

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)
        snapshots/bindings_aarch64-apple-darwin_4096.rs
```

I'm pretty sure we should be ignoring these generated files.

### DIFF
--- a/bindings/rust/.gitignore
+++ b/bindings/rust/.gitignore
@@ -1,2 +1,3 @@
 src/bindings/generated.rs
+snapshots/
 target/


jtraglia:
I will update the example in a following PR. A little complicated to do in this one.

jtraglia:
Or should I add this snapshot file to the repo? That seems unnecessary IMO but maybe that was the intention.

divagant-martian:
I kinda wanted us to add them to now what they look like per platform, since I don't know if they will have differences. If they differ from the linux ones maybe we should add them for the record. Also to identify differences if with time we upgrade bindgen. What are your thoughts on this?

jtraglia:
Hmm I expect these to all be the exact same for all of the platform combinations; I don't see anything that should change. And just to confirm, the macOS/ARM64 & Linux/x86_64 bindings have zero differences.

divagant-martian:
I'm working on the windows ones and they have one minimal difference (please don't check this yet). I'm ok with ignoring them as long as we update them if we identify relevant changes

divagant-martian (approved):

jtraglia:
Interesting 🤔 let's see what @asn-d6 thinks.

asn-d6:
No strong opinion. I defer to @divagant-martian and/or @pawanjay176 .

jtraglia:
Closing. I think we will want to either add snapshots for everything or ignore all snapshots. Something to discuss later.

PR #254, opened 2023-03-28
Check the return value of fclose in C# bindings

jtraglia:
Check the return value of this operation. Clean up that function a bit.

### DIFF
--- a/bindings/csharp/ckzg.c
+++ b/bindings/csharp/ckzg.c
@@ -4,19 +4,39 @@
 #include "c_kzg_4844.h"
 #include "ckzg.h"
 
-KZGSettings* load_trusted_setup_wrap(const char* file) {
-  KZGSettings* out = malloc(sizeof(KZGSettings));
+KZGSettings* load_trusted_setup_wrap(const char *file) {
+  C_KZG_RET ret;
+  KZGSettings *s = NULL;
+  FILE *f = NULL;
 
-  if (out == NULL) return NULL;
+  /* Allocate memory for the trusted setup */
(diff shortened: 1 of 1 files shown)


asn-d6:
So after reviewing the various fclose PRs, I think the complexity that gets added to the code is not worth it. I just spent about 20 mins reviewing the C# PR to make sure we are not double freeing stuff.

My new suggestion is to close all these PRs and ignore the retval of fclose.

My argument for this is that if load_trusted_setup_file() succeeds, then we don't care about what happens to fclose. We are opening the file just for reading so it's not like the failed `fclose()`can result in the file ending in an inconsistent state. And since the read (i.e. loading the trusted setup) has succeeded, I think there is no better action to take than ignore the retval.

tl;dr I think my suggestion to check the retval of fclose was wrong, and we can abort the plan now, so that we don't add needless complexity to all the bindings.

(Whether we should revert the nodejs changes, I have no strong opinion about)

ppopth:
I think this is fine. I would say that not checking fclose is done successfully is probably not much better than not doing fclose at all, but both are fine.

The only effect of not doing fclose (and failing to do fclose) is that the file descriptor will remain open. If the clients (or anyone who uses c-kzg) call `load_trusted_setup` many times, they will probably run out of the file descriptors.

As long as the clients call `load_trusted_setup` only once, this is not a problem. The only concern is that someone (other than Ethereum clients) who uses c-kzg may be surprised that they run out of the file descriptors after calling `load_trusted_setup` many times which is what they don't expect.

ps. I put the note just so that we can keep the effect in mind. I'm okay with the status quo.

PR #255, opened 2023-03-28
Check the return value of fclose in java bindings

jtraglia:
Like the other bindings, this will check the return value of `fclose`.

Also, cleaned up the `reset_trusted_setup` function a little. We don't need to do null checks anymore.


asn-d6:
Closing after discussion with Justin. Please see https://github.com/ethereum/c-kzg-4844/pull/254#issuecomment-1488049094 for more context. Please reopen if needed.

PR #259, opened 2023-03-29
Windows bindings for rust

divagant-martian:
Adds windows bindings for rust


divagant-martian:
ok some info here for whoever is interested. I know exactly how the windows bindings are failing. The error is pretty simple, `cc` is giving the wrong flag to output the object file to the c compiler. After this issue is solved this PR should be easy to complete. 

Hang in there for more to come

jtraglia:
You're making good progress, I see that you've gotten the full windows experience 😂

Regarding the most recent issue:

> D:\a\c-kzg-4844\c-kzg-4844\blst\bindings\blst_aux.h(53): error C2016: C requires that a struct or union have at least one member

For Node.js we decided to patch this file during the build process:

https://github.com/ethereum/c-kzg-4844/blob/34f4fa1609ac53620f8dd9889b34d7f63c9ed694/bindings/node.js/Makefile#L37-L38

There may be an alternative solution, but this may help. That command should work on all the platforms.

divagant-martian:
Indeed such a nightmare. So I go this "working" (as in, compiling and linking) correctly using clang, but tests fail with an error that looks like it might be related to something you got. Can you check this run please? 

https://github.com/ethereum/c-kzg-4844/actions/runs/4580768026/jobs/8089770603

jtraglia:
Yeah, that's a stack overflow. It's probably because the stack on Windows (1MB) is significantly smaller than the stack on Linux/macOS (8MB). Stack overflows can be pretty difficult to track down & it's normally done with a debugger, but good luck with that without a Windows system.

My advice:

* First, I see that you've tested increasing the stack size; I would test that again (via `-Wl,-stack_size,<bytes>` like you were doing before), but to an even larger value, like 20MB instead of 8MB. Just to see if the error goes away.
* Next, I'd take a look at the Rust tests to see if there's an array of Blobs/Polynomials being allocated on the stack. These types are really large (128KB each) and could cause probably if you have 4 of them on the stack at a given time. I think this C test is the reason I needed to raise the stack size in the C code (I've fixed this in yet-to-be-merged PR and the tests pass on Windows with the original stack size now):
https://github.com/ethereum/c-kzg-4844/blob/34f4fa1609ac53620f8dd9889b34d7f63c9ed694/src/test_c_kzg_4844.c#L1460-L1466
* Lastly, if you still can't figure it out, I'd start by disabling all of the tests (comment them out) and re-enable them one-by-one (or binary-search style) to help determine which particular test is causing the stack overflow.

jtraglia:
Oh another thing, you test raising the stack size in Rust. Add something like this to `Cargo.toml`:
```toml
linker = "rust-lld"
rustflags = ["-C", "link-arg=-Wl,--stack,8388608"]
```

divagant-martian:
so I've tracked it down to the unsafe parts of handling the file in windows before hitting the ffi. It would be ideal if the c code received a char pointer instead of a file to avoid this kind of code

jtraglia:
Ah nice find! 🥇  I wasn't expecting the problem to be with loading the trusted setup.

> It would be ideal if the c code received a char pointer instead of a file to avoid this kind of code.

I'm not sure why we do it like this. Maybe something we can look into changing.

.github/workflows/rust-tests.yml
jtraglia, line 1: For this file, instead of `windows-latest-tests` and `windows-2019-tests`, could you add a matrix like this?

https://github.com/ethereum/c-kzg-4844/blob/34f4fa1609ac53620f8dd9889b34d7f63c9ed694/.github/workflows/c-tests.yml#L11-L18

This is how the other workflow files do it.

bindings/rust/build.rs
```diff
@@ -30,38 +21,85 @@ fn main() {
...
+    /*
+     * BLST env
+     */
+
+    let target_arch = env::var("CARGO_CFG_TARGET_ARCH").unwrap();
+
+    if target_arch.eq("x86_64") || target_arch.eq("aarch64") {
+    } else {
+        cc.define("__BLST_NO_ASM__", None);
     }
```
jtraglia, line 39: This is a neat section. I think you should add a `force-arm-feature-crypto` (maybe with a better name) for ARM systems. Some ARM systems (such as new the M1/M2 Macs) have specialized crypto instructions and blst supports that. For reference, [this is where this is checked for in blst](https://github.com/supranational/blst/blob/a7fd1f584d26b0ae6cdc427976ea1d8980f7e15d/src/sha256.h#L14-L15).

bindings/rust/build.rs
```diff
@@ -30,38 +21,85 @@ fn main() {
...
-        ))
+    // Obtain the header files exposed by blst-bindings' crate.
+    let blst_headers_dir =
+        std::env::var_os("DEP_BLST_BINDINGS").expect("BLST exposes header files for bindings");
+    /*
+     * Hack ahead
+     */
+    const AWK_COMMAND: &str =
+        "{gsub(/typedef struct \\{\\} blst_uniq;/, \"typedef struct { int dummy; } blst_uniq;\")}1";
+    let awk_succeeds = Command::new("awk")
```
jtraglia, line 75: We should look into ways of removing this hack. On Windows, if you compile with GCC, I don't think this hack will be necessary. I believe this is only a problem with MSVC. But if we want to support both MSVC & GCC, then yeah I think this is necessary.

bindings/rust/build.rs
```diff
@@ -30,38 +21,85 @@ fn main() {
...
+        panic!("awk command failed")
+    }
+
+    /*
+     * End of hack
+     */
+
+    #[cfg(windows)]
+    cc.flag("-D_CRT_SECURE_NO_WARNINGS");
+    // .flag("-Wl,-z,-stack_size,8388608");
```
jtraglia, line 94: Remove commented out code.

bindings/rust/build.rs
```diff
@@ -182,6 +211,10 @@ fn make_bindings<P>(
     bindings
         .write_to_file(bindings_out_path)
         .expect("Failed to write bindings");
+    // let generated = bindings.to_string();
+    // for line in generated.lines() {
+    // println!("cargo:warning=Generated\t{line}");
+    // }
```
jtraglia, line 217: Remove commented out code.

bindings/rust/src/bindings/mod.rs
```diff
@@ -102,12 +100,38 @@ impl KZGSettings {
...
+            let file_path =
+                CString::new(file_path.as_os_str().to_str().unwrap().as_bytes()).unwrap();
+
+            let fp = unsafe { libc::fopen(file_path.as_ptr(), mode.as_ptr()) };
+            if fp.is_null() {
+                let e = std::io::Error::last_os_error();
+                println!("Failed to get file_descriptor {e}");
+                return Err(Error::CError(C_KZG_RET::C_KZG_BADARGS));
+            }
+            fp
```
jtraglia, line 131: I think this chunk is the exact same as unix version. Is it possible to de-dup a bit here?

bindings/rust/build.rs
```diff
@@ -30,38 +21,85 @@ fn main() {
...
+        "{gsub(/typedef struct \\{\\} blst_uniq;/, \"typedef struct { int dummy; } blst_uniq;\")}1";
+    let awk_succeeds = Command::new("awk")
+        .current_dir(blst_headers_dir.clone())
+        .arg("-i")
+        .arg("inplace")
+        .arg(AWK_COMMAND)
+        .arg("blst_aux.h")
         .status()
-        .unwrap();
+        .expect("ask is installed")
```
jtraglia, line 82: Minor typo here:
```suggestion
        .expect("awk is installed")
```

jtraglia (commented):
Looking great! Not sure if this was ready for review, but here's an initial one.

bindings/rust/build.rs
divagant-martian: This was actually not necessary, and doesn't work with cl

bindings/rust/build.rs
divagant-martian: since blst is compiled by the library, the safer option is using the same compiler (cl) instead of gcc

bindings/rust/build.rs
divagant-martian: I actually removed all of this until it's clear how c-kzg maps to blst feature flags. This whole code is executed by the blst build script and is then unnecessary

divagant-martian:
@jtraglia please take into account this is still wip, give it till monday

bindings/rust/build.rs
divagant-martian: Do you think we could submit this change to the blst crate itself instead?

jtraglia: I believe that would be the ideal solution. Let's see if we can make that happen.

jtraglia: Made a PR here. Not sure if they'll like it or how long it could take.

* https://github.com/supranational/blst/pull/157

jtraglia: I suggest we proceed without that blst PR. If it get's merged in the future, we'll remove whatever hack we do here.

divagant-martian:
@jtraglia this is ready for review. I will probably need to get back to this to remove the blst dependency to avoid double linking but this gives us a pretty good baseline

jtraglia (approved):
Nice. This turned out to be pretty clean. LGTM 👍

PR #260, opened 2023-03-29
Remove tsc build step and export raw .js file and .d.ts from node bindings

matthewkeil:
The motivation for this PR is removing all of the extra stack frames for function calls.  This refactor will also result in better readability of the code and a simpler build process.

The original .ts export file had a significant amount of business logic and thus was better to build with the typescript compiler `tsc`.  The output of the build process is the files, `.js` (actual file used during runtime) and `.d.ts` (TS definitions for compiler/intellisense), that get consumed by end users.

After moving the logic to `c`, most of the functions look like this
```ts
export function blobToKzgCommitment(blob: Blob): KZGCommitment {
  return kzg.blobToKzgCommitment(blob);
}
```

## Inclusions
- This PR removes all the proxied function calls and exports the native functions directly.
- It also removes the`tsc` build step and uses a hand build `.js` and `.d.ts` files


bindings/node.js/lib/kzg.js
```diff
@@ -0,0 +1,51 @@
...
+    "\n" +
+    trustedSetup.setup_G2.map((p) => p.substring(2)).join("\n");
+  const outputPath = filePath.replace(".json", ".txt");
+  fs.writeFileSync(outputPath, setupText);
+  return outputPath;
+}
+
+const originalLoadTrustedSetup = bindings.loadTrustedSetup;
+// docstring in ./kzg.d.ts with exported definition
+bindings.loadTrustedSetup = function loadTrustedSetup(filePath) {
```
matthewkeil, line 36: note that adding the function name instead of an anonymous function will preserve the name in error stack traces

bindings/node.js/test/kzg.test.ts
```diff
@@ -1,6 +1,6 @@
 import { randomBytes } from "crypto";
 import { readFileSync } from "fs";
-import { join, resolve } from "path";
```
matthewkeil, line 3: This was a leftover unused import that I removed

bindings/node.js/lib/kzg.js
```diff
@@ -0,0 +1,51 @@
...
+    trustedSetup.setup_G1.map((p) => p.substring(2)).join("\n") +
+    "\n" +
+    trustedSetup.setup_G2.map((p) => p.substring(2)).join("\n");
+  const outputPath = filePath.replace(".json", ".txt");
+  fs.writeFileSync(outputPath, setupText);
+  return outputPath;
+}
+
+const originalLoadTrustedSetup = bindings.loadTrustedSetup;
+// docstring in ./kzg.d.ts with exported definition
```
matthewkeil, line 35: Is this a sufficient docstring?

matthewkeil:
@dapplion and @g11tech this is a primarily JS/TS PR.  Would you mind taking a peek to double check my work?

jtraglia:
LGTM 👍 excellent work, I like this a lot.

dapplion (approved):
Sounds good to me given how simple the JS glue layer now is

PR #261, opened 2023-03-29
Simplify fclose() logic in nodejs

asn-d6:
This PR removes the error handling of fclose() since it was complicating the function without adding much.

See https://github.com/ethereum/c-kzg-4844/pull/254#issuecomment-1488049094 for more context.


jtraglia (approved):
Looks great, thanks @asn-d6 👍

PR #262, opened 2023-03-29
Add bundle to publish command

matthewkeil:
@g11tech published the new bindings and noticed a bug running `publish`.  The command work when all run but `publish` fails when run independently.

## Inclusions
- Add bundle to publish command dependencies


jtraglia (approved):
LGTM 👍 thanks!

PR #263, opened 2023-03-29
Update C tests to support minimal config

jtraglia:
Wanted to be able to run all of these & have all of the tests pass:

* `make`
* `FIELD_ELEMENTS_PER_BLOB=4096 make`
* `FIELD_ELEMENTS_PER_BLOB=4 make`

There were just a few tests that needed to be updated.

Additionally, I went on a bit of a journey trying to figure out why it would stack overflow on Windows in the minimal test. Unfortunately, I don't have a good answer. But I did the following things:

* Use `c_kzg_malloc` for allocating 16 blobs, instead of stack.
  * I think this is the reason it had an issue on mainnet before.
  * 16 * 128KB = 4MB, which is larger than the default Windows stack size (1MB).
  * Then I removed the manually stack size increase for the Windows build.
* Minimal would crash but mainnet wouldn't.
  * This doesn't make much sense to me.
* It didn't crash with debug prints, which leads me to believe it's an optimization thing.
* Build & test in different steps.
  * This allows me to download the test binary & inspect it with Ghidra.
  * In one step, the upload step won't run.

All in all, it seems to work now, but I will investigate more if it segfaults again.


src/test_c_kzg_4844.c
```diff
@@ -12,6 +12,20 @@
...
+
+#if FIELD_ELEMENTS_PER_BLOB == 4096
+#define MAINNET
+#define TRUSTED_SETUP_FILE "trusted_setup.txt"
+#define MAX_WIDTH 32
+#elif FIELD_ELEMENTS_PER_BLOB == 4
+#define MINIMAL
+#define TRUSTED_SETUP_FILE "trusted_setup_4.txt"
+#define MAX_WIDTH 4
+#endif
```
asn-d6, line 27: Maybe we need a fallback else that crashes here, if it's neither 4096 or 4?

asn-d6 (commented):
Small nit. Other than that, looks great!

asn-d6 (approved):

asn-d6:
Hmm, seems like the C tests are failing for some Makefile reasons.

jtraglia:
> Hmm, seems like the C tests are failing for some Makefile reasons.

Ah Windows strikes again. No worries, I'll fix it.

ppopth:
> Lowered `n_samples` in `test_verify_kzg_proof_batch__succeeds_round_trip`.

I have two issues with this one
1. Why call it `n_samples`? It seems to me that it's not the number of samples, but it's the number of blobs. This confused me in the beginning.
2. Why set it to 3? `MAX_BLOBS_PER_BLOCK` in Deneb is 4. Will it overflow the stack in Windows if we do verify_blob_kzg_proof_batch with the number of blobs of 4?

src/test_c_kzg_4844.c
```diff
@@ -1459,7 +1485,7 @@ static void test_compute_and_verify_blob_kzg_proof__fails_invalid_blob(void) {
 
 static void test_verify_kzg_proof_batch__succeeds_round_trip(void) {
     C_KZG_RET ret;
-    const int n_samples = 16;
+    const int n_samples = 3;
     Bytes48 proofs[n_samples];
```
ppopth, line 1489: I think this is worth allocating the memory in the heap to avoid stack overflow. I think the number of blobs of 3 is too low.

With 4844, the number of blobs will be 4 per block (MAX_BLOBS_PER_BLOCK).
With Danksharding, the number of samples can be as high as 70 per block.

ppopth:
Other than my comments above, I think it's good to go.

PR #264, opened 2023-03-29
Add more information to the readme

jtraglia:
The README feels empty and I would like to add more information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,8 +58,8 @@ cd src && make
 ### Testing
 
 All the bindings are tested against the [KZG reference tests defined in the
-consensus-spec-tests](KZG spec
-tests)[https://github.com/ethereum/consensus-spec-tests/tree/master/tests/general/deneb/kzg).
+consensus-spec-tests]
+(https://github.com/ethereum/consensus-spec-tests/tree/master/tests/general/deneb/kzg).
 Additionally, a suite of custom unit tests for the C functions is located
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,43 +29,56 @@ There are functions for KZG operations:
...
+### Why C?
 
-```
-cd src && make
-```
+The primary reason for choosing C is because
+[blst](https://github.com/supranational/blst), the BLS12-381 signature library
+we wanted to use, is mostly written in C. We were concerned that using blst with
+another language, like Rust, and then building bindings on top of it would
+introduce too much overhead.  Furthermore, the C toolchain is ubiquitous, and it
```
asn-d6, line 82: Given that the FFI overhead from rust is not really a concern anymore, perhaps you can remove the overhead sentence. Another valid positive of doing it in C is that it doesn't have the heavy Rust runtime (smaller memory/binary footprint).

asn-d6:
This all looks great! Not merging yet since this is a draft.

README.md
jtraglia: Will update this paragraph to reflect this 👍

jtraglia:
@asn-d6 Alright, I think I'm happy with the current state of this. Ready for a final review.

Also, happy to add other remarks if you can think of any.

ppopth:
LGTM!

PR #266, opened 2023-03-30
Clean-up node error messages

matthewkeil:
@jtraglia noticed that some of the older error messages are lower-case.  He requested to standardize them.  He also requested the `C_KZG_RET` be added to error messages where applicable

### DIFF
--- a/bindings/node.js/src/kzg.cxx
+++ b/bindings/node.js/src/kzg.cxx
@@ -7,8 +7,8 @@
 #include "blst.h"
 
 /**
- * Convert C_KZG_RET to a string representation for error messages 
-*/
+ * Convert C_KZG_RET to a string representation for error messages.
+ */
 std::string from_c_kzg_ret(C_KZG_RET ret) {
     switch (ret) {
       case C_KZG_RET::C_KZG_OK:
--- a/bindings/node.js/src/kzg.cxx
+++ b/bindings/node.js/src/kzg.cxx
@@ -7,6 +7,22 @@
 #include "blst.h"
 
 /**
+ * Convert C_KZG_RET to a string representation for error messages 
+*/
+std::string from_c_kzg_ret(C_KZG_RET ret) {
+    switch (ret) {
+      case C_KZG_RET::C_KZG_OK:
+        return "C_KZG_OK";
+      case C_KZG_RET::C_KZG_BADARGS:
+        return "C_KZG_BADARGS";
+      case C_KZG_RET::C_KZG_ERROR:
+        return "C_KZG_ERROR";
+      case C_KZG_RET::C_KZG_MALLOC:
+        return "C_KZG_MALLOC";
+    }
+}
+
+/**
  * Structure containing information needed for the lifetime of the bindings
  * instance. It is not safe to use global static data with worker instances.
  * Native node addons are loaded as a dll's once no matter how many node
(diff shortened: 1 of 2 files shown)


bindings/node.js/src/kzg.cxx
```diff
@@ -6,6 +6,22 @@
...
+std::string from_c_kzg_ret(C_KZG_RET ret) {
+    switch (ret) {
+      case C_KZG_RET::C_KZG_OK:
+        return "C_KZG_OK";
+      case C_KZG_RET::C_KZG_BADARGS:
+        return "C_KZG_BADARGS";
+      case C_KZG_RET::C_KZG_ERROR:
+        return "C_KZG_ERROR";
+      case C_KZG_RET::C_KZG_MALLOC:
+        return "C_KZG_MALLOC";
```
matthewkeil, line 21: There is a "not all control paths return a value" lint error for windows.  All options are covered but there is no "default".  Should I return one of the values by default to remove the lint error?

jtraglia: Sure, a default case is fine.

matthewkeil: Which would you like returned for default, `C_KZG_ERROR`?

bindings/node.js/src/kzg.cxx
```diff
@@ -152,7 +168,9 @@ Napi::Value LoadTrustedSetup(const Napi::CallbackInfo& info) {
 
   // Check that loading the trusted setup was successful
   if (ret != C_KZG_OK) {
-    Napi::Error::New(env, "Error loading trusted setup file: " + file_path).ThrowAsJavaScriptException();
+    std::ostringstream msg;
+    msg << "[C_KZG_RET::" << from_c_kzg_ret(ret) << "] Error loading trusted setup file: " << file_path;
```
jtraglia, line 172: In reference to my other comment, this one is a little more complicated.

And now I see why you did it the way you did.

Maybe keep it simple and just print the return value, like this:
```suggestion
    msg << "Error loading trusted setup file: " << from_c_kzg_ret(ret);
```

Because above this, it will print the file path if it doesn't exist:

https://github.com/ethereum/c-kzg-4844/blob/e83ec0374ce95003603f3e552f15d10f1707408f/bindings/node.js/src/kzg.cxx#L142-L146

jtraglia (commented):
Looks good, just a couple of nits.

bindings/node.js/src/kzg.cxx
```diff
@@ -376,7 +399,9 @@ Napi::Value VerifyBlobKzgProof(const Napi::CallbackInfo& info) {
     kzg_settings);
 
   if (ret != C_KZG_OK) {
-    Napi::TypeError::New(env, "Error in verifyBlobKzgProof").ThrowAsJavaScriptException();
+    std::ostringstream msg;
+    msg << "[C_KZG_RET::" << from_c_kzg_ret(ret) << "] Error in verifyBlobKzgProof";
```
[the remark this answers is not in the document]
matthewkeil: Updated.

bindings/node.js/src/kzg.cxx
jtraglia: Hmm good question. Is it possible to throw an exception?

If not, I think I'd return something like "UNKNOWN (27)" where "27" is integer return value.

jtraglia: Ah that's right. Sounds good.

jtraglia (approved):
Nice, thanks for making those changes. LGTM 👍

Will merge after the CI tests pass (without that warning).

PR #267, opened 2023-03-30
Normalize native return for JS errors

matthewkeil:
Normalize the return value when a JS error is raised.  As a note this was a final clean-up thing.  I was speaking with @jtraglia yesterday and he mentioned a couple updates that were added to #266 .  During the discussion I brought this PR up and suggested I would put it up.  This is not something Ii feel strongly about, just think it will make the code a bit cleaner.  I will be happy to update/merge or close as requested by the team.

If an exception is pending the value returned from the native function will never manifest in the JS context.  It is just for readability and standardization.  The actual values do not get used.

`undefined` is the uninitialized/default value in JS so I used that for all cases


jtraglia (approved):
Nice, LGTM 👍 

> If an exception is pending the value returned from the native function will never manifest in the JS context. It is just for readability and standardization. The actual values do not get used.

I'm glad we have some clarity on this. Thanks again.

PR #268, opened 2023-03-30
Format node bindings with clang-format

matthewkeil:
No code changes were made to `kzg.cxx` only format was run

## Inclusions
- copy `.clang-format` to node bindings so IDE format command works
- run format on `kzg.cxx`
- add clang-format to `Makefile`

### DIFF
--- a/bindings/node.js/src/kzg.cxx
+++ b/bindings/node.js/src/kzg.cxx
@@ -1,10 +1,10 @@
-#include <stdio.h>
-#include <iostream>
-#include <sstream>  // std::ostringstream
-#include <string_view>
-#include <napi.h>
-#include "c_kzg_4844.h"
 #include "blst.h"
+#include "c_kzg_4844.h"
+#include <iostream>
+#include <napi.h>
+#include <sstream> // std::ostringstream
+#include <stdio.h>
+#include <string_view>
 
 /**
  * Structure containing information needed for the lifetime of the bindings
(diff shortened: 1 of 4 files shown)


matthewkeil:
@jtraglia I added the lint command to the `Makefile` but `clang-format` is not on the workflow VM.  What do you think is the best course of action for that?

bindings/node.js/src/kzg.cxx
```diff
@@ -376,173 +388,201 @@ Napi::Value VerifyKzgProof(const Napi::CallbackInfo& info) {
...
  *
  * @remark blobs[0] relates to commitmentBytes[0] and proofBytes[0]
  *
  * @param[in] {Blob}    blobs - An array of serialized blobs to verify
- * @param[in] {Bytes48} commitmentBytes - An array of serialized commitments to verify
- * @param[in] {Bytes48} proofBytes - An array of serialized KZG proofs for verification
+ * @param[in] {Bytes48} commitmentBytes - An array of serialized commitments to
+ *                                        verify
+ * @param[in] {Bytes48} proofBytes - An array of serialized KZG proofs for
+ *                                   verification
```
matthewkeil, line 435: Not a fan of how these turned out.  They don't stack but I tried a couple options and this looks the least bad...  Open to suggestions @jtraglia

jtraglia: Yeah, I don't think it looks great either but I also think it's the best option.

jtraglia (approved):
LGTM 👍 If you guys want to change the formatting config later, that's fine too. I don't really care what the style is as long as everything is consistent.

PR #270, opened 2023-03-30
Run C tests on all platforms

jtraglia:
This PR will build/test the C library on all platforms (Linux, macOS, and Windows).

Also, did general cleanup of the Makefile:

* Define new `PLATFORM` variable which is assigned at the top.
* Define new `LIBS` variable with has all of the libraries to build with.
* Replace all of configuration options with setting in the rules.
* Fix some indentation issues.
* Organize everything better.

### Some notes

I used gcc rather than clang for building on Windows because clang isn't part of the mingw32 toolchain. Using clang led to an unresolved symbol error related to `___chkstk_ms`. This happened because blst was built with gcc, not clang. Although we could build blst with `build.bat` instead of `build.sh`, it was more difficult and caused the Java bindings build to fail on Windows as they also used gcc.

When building with gcc on Windows, the system expected `blst.lib`, but the actual file was `libblst.a`. To fix this, I specified the `../lib/libblst.a` file instead of `-lblst`.

Eventually, I encountered a "Stack Overflow" exception with the error code `-1073741571` (or `0xC00000FD`). This was likely due to the smaller stack size on Windows compared to Linux/macOS, especially when working with large structures like `Blob` and `Polynomials`. As a solution, I increased the stack size to 8MB for Windows. Further investigation into the exact cause of the exception could be done in the future.

### DIFF
--- a/.github/workflows/c-tests.yml
+++ b/.github/workflows/c-tests.yml
@@ -34,15 +34,18 @@ jobs:
         working-directory: src
         run: make sanitize
       - name: Clang Static Analyzer
+        if: matrix.os != 'windows-latest'
         working-directory: src
         run: make analyze
       - name: Install LLVM
-        if: matrix.os != 'macos-latest'
+        if: matrix.os == 'ubuntu-latest'
         uses: egor-tensin/setup-clang@v1
       - name: Generate coverage report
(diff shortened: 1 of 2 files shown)


asn-d6:
Looks great!

PR #271, opened 2023-03-30
Fix some nodejs nits

jtraglia:
* Rename "test" to "tests" to match other workflows.
* Remove punctuation from comment to match others.
* Hide comment from showing up in the output.


matthewkeil (approved):

jtraglia:
Thanks 🙂

PR #273, opened 2023-03-31
Update Go bindings example

jtraglia:
We recently updated the interface a little. This updates the example to reflect that.


asn-d6:
LGTM!

PR #274, opened 2023-03-31
delegate blst build to crate in preparation for windows bindings

divagant-martian:
Moves the responsibility of building blst from this crate using make, to the rust crate for blst. The blst crate knows the best way to build blst according to features, platform-dependent flags etc. This is necessary in preparation for the windows bindings.

Also updates use of paths everywhere to use os-specific paths instead of linux paths


jtraglia (approved):
LGTM 👍

PR #275, opened 2023-04-01
Add some comments & fix some nits

jtraglia:
* Add comments where I felt they were missing.
* Change `log2_pow2` to use simpler Implementation.
  * This is only used in the trusted setup anyway.
* Move a `bytes_to_kzg_proof` call up, to fail faster.
* Change `(uint64_t)1` to `1ULL` to be more concise.
* Move some trusted setup functions down to its section.
* Group trusted setup free funcs together.
* Update error message for `FIELD_ELEMENTS_PER_BLOB` in header.
* Replace `//` comments with `/* */` comments.
* Remove some excess parentheses.


ppopth:
Other than my comments above, I think it's good to go.

ppopth (approved):

PR #277, opened 2023-04-03
For load_trusted_setup_file, take string file path

### LINKED ISSUE #276: Consider using a char buffer instead of FILE
Using a char buffer would be IMO a more compatible version of the current api. 

Chars have a standard representation across platforms, while FILEs are a linux api that windows supports _to some degree_. The windows version of it are handles (HFILE). I see other languages are lenient on dealing with files/os paths but in general these don't exactly map since windows paths are encoded using wide chars. Under the current api we need to deal with these nuances at the ffi boundary. Using a char buffer instead we would delegate opening the file to the language implementation, and simply pass the contents to the c code

jtraglia:
Instead of taking a `FILE *`, take a `const char *` which is easier for the bindings.

In this comment, I said we should look into changing this behavior:

* https://github.com/ethereum/c-kzg-4844/pull/259#issuecomment-1493192541

Thanks for pointing out how weird this was @divagant-martian.

Fixes #276.

### DIFF
--- a/bindings/csharp/ckzg.c
+++ b/bindings/csharp/ckzg.c
@@ -9,13 +9,8 @@ KZGSettings* load_trusted_setup_wrap(const char* file) {
 
   if (out == NULL) return NULL;
 
-  FILE* f = fopen(file, "r");
+  if (load_trusted_setup_file(out, file) != C_KZG_OK) { free(out); return NULL; }
 
-  if (f == NULL) { free(out); return NULL; }
-
-  if (load_trusted_setup_file(out, f) != C_KZG_OK) { free(out); fclose(f); return NULL; }
-
-  fclose(f);
   return out;
 }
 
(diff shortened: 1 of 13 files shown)


jtraglia:
Also, the `generated.rs` file is out of date now, but I'd have to manually modify it to fix it.

I don't have an x86 system with Linux.

ppopth:
One issue that I can see from this approach is that the bindings cannot know if there is an error from opening the file anymore.

You can see that the error message in node.js, java, and go bindings just disappeared.

ppopth (approved):

jtraglia:
Yeah, that's a good point. One solution would be more fine-grain `C_KZG_RET` values. For example, instead of returning `C_KZG_BADARGS`, we could return `C_KZG_ERROR_NO_SUCH_FILE` or something.

divagant-martian:
This achieves not dealing with files across ffi boundary which on itself is an improvement. For these and other reasons I suggested we simply deal with the file in the bindings and pass the contents instead in #276

jtraglia:
Closing. See discussion in #276.

PR #279, opened 2023-04-05
Remove rust snapshots

jtraglia:
An alternative to #253. Instead of adding all of the snapshot files, I propose we get rid of snapshots entirely.

If the interface changes in the future, developers will be unable to update these snapshots unless they build the bindings on the same target (architecture/operating system/preset). For example, someone without a Mac would be unable to update the snapshot files for Mac. Also, there would be a lot of snapshot files if we save all of them.

Is there a good reason (besides to reference) to have these?

@divagant-martian & @pawanjay176 what do you think?


divagant-martian:
yeah I'm ok with this

divagant-martian (approved):
lgtm

PR #280, opened 2023-04-06
Update the nodejs exported  Buffer types to more universal Uint8Array

g11tech:
Replace the `Buffer` types with a more universal `Uint8Array` in the kzg interface

Even though the current implementation only supports nodejs as of now (and using the `Buffer` objects which are just inherited from `Uint8Array`), the libraries which import this interface (think lodestar/ethereumjs) start breaking in the browser.

Even if c-kzg may not aspire to be browser compatible, the interface can be standard and libraries can be dynamically loaded/replaced by the underlying application.

### DIFF
--- a/bindings/node.js/test/kzg.test.ts
+++ b/bindings/node.js/test/kzg.test.ts
@@ -64,8 +64,40 @@ const proofBadLength = randomBytes(BYTES_PER_PROOF - 1);
 const fieldElementValidLength = randomBytes(BYTES_PER_FIELD_ELEMENT);
 const fieldElementBadLength = randomBytes(BYTES_PER_FIELD_ELEMENT - 1);
 
-function bytesFromHex(hexString: string): Buffer {
-  return Buffer.from(hexString.slice(2), "hex");
+function bytesEqual(a: Uint8Array | Buffer, b: Uint8Array | Buffer): boolean {
+  if (a.length !== b.length) {
+    return false;
+  }
+  for (let i = 0; i < a.length; i++) {
+    if (a[i] !== b[i]) return false;
+  }
+  return true;
+}
+
+function padToEven(a: string): string {
+  return a.length % 2 ? `0${a}` : a;
+}
+
+function bytesFromHex(hex: string): Uint8Array {
+  if (typeof hex !== "string") {
+    throw new Error(`hex argument type ${typeof hex} must be of type string`);
+  }
+
+  if (hex.startsWith("0x")) {
+    hex = hex.slice(2);
+  }
+
+  if (hex.length % 2 !== 0) {
+    hex = padToEven(hex);
+  }
+
+  const byteLen = hex.length / 2;
+  const bytes = new Uint8Array(byteLen);
+  for (let i = 0; i < byteLen; i++) {
+    const byte = parseInt(hex.slice(i * 2, (i + 1) * 2), 16);
+    bytes[i] = byte;
+  }
+  return bytes;
 }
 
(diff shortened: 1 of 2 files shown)


bindings/node.js/test/kzg.test.ts
```diff
@@ -64,8 +64,40 @@ const proofBadLength = randomBytes(BYTES_PER_PROOF - 1);
...
+    hex = padToEven(hex);
+  }
+
+  const byteLen = hex.length / 2;
+  const bytes = new Uint8Array(byteLen);
+  for (let i = 0; i < byteLen; i++) {
+    const byte = parseInt(hex.slice(i * 2, (i + 1) * 2), 16);
+    bytes[i] = byte;
+  }
+  return bytes;
```
jtraglia, line 100: Alternatively, I'd suggest doing this which I believe is a lot simpler:

```js
function bytesFromHex(hexString: string): Uint8Array {
  const buffer = Buffer.from(hexString.slice(2), "hex");
  return new Uint8Array(buffer.buffer, buffer.byteOffset, buffer.byteLength);
}
```

jtraglia (changes requested):
Good idea, I agree that these should be `Uint8Array` types. Previously, we made these `Buffer` types because comparing against a buffer wasn't working. Just a couple of nits.

bindings/node.js/test/kzg.test.ts
```diff
@@ -116,7 +148,7 @@ describe("C-KZG", () => {
         }
 
         expect(test.output).not.toBeNull();
-        expect(proof).toEqual(test.output.map((hex) => bytesFromHex(hex)));
+        expect(proof.reduce((acc, pitem, pindex) => acc && bytesEqual(pitem, bytesFromHex(test.output[pindex])), true));
```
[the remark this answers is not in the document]
g11tech: ProofResult (`proof`) seems to be an array of Uint8.../Buffer (so correspondingly test.output.map) so this is just running a map reduce to compare if each array items are equal

bindings/node.js/test/kzg.test.ts
g11tech: I initially thought of this but decided against this in case the c-kzg becomes browser compatible, since this is a test file i don't really have strong opinion so can update it accordingly

matthewkeil: can also 

```suggestion
function bytesFromHex(hexString: string): Uint8Array {
  return Uint8Array.from(Buffer.from(hexString.slice(2), "hex"))
}
```

bindings/node.js/test/kzg.test.ts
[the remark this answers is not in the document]
matthewkeil: The type was changed to a tuple

```ts
export type ProofResult = [KZGProof, Bytes32];
```

bindings/node.js/test/kzg.test.ts
jtraglia: > decided against this in case the c-kzg becomes browser compatible

I'm not that experienced with javascript, why would this make a difference?

jtraglia: @matthewkeil ah yes, that is even cleaner 👍

g11tech: Because buffer is not available in browser and if you would run these tests in karma (to always maintain browser compatibility)
but that can be done when c-kzg  gets at it i guess.

jtraglia (approved):
LGTM 👍

PR #282, opened 2023-04-06
[Go] Clean things up a bit

jtraglia:
This PR will:

* Change package name from `cgokzg4844` to `ckzg4844`.
  * Maybe it'll be easier to find on pkg.go.dev now. 
* Change README header to "go bindings".
* Make `GetRand*` testing functions private.
* Check for "0x" in `UnmarshalText` before stripping.
  * This will allow `0xabcd` and `abcd` now.
* Move `UnmarshalText` functions to `main.go`.
  * Cannot add methods to non-local types.
  * I think these are good to define publicly.
* Remove private functions (`sha256`).
  * Unnecessary, really. Only used for benchmark.
* Update the example's `go.mod` file.
  * And the example output in the readme, just to be extra clear.

### DIFF
--- a/bindings/go/README.md
+++ b/bindings/go/README.md
@@ -21,8 +21,8 @@ For reference, see the `example` module in this directory. You can test it out w
 
 ```
 user@system ~/c-kzg-4844/bindings/go/example $ go run .
-go: downloading github.com/ethereum/c-kzg-4844 v0.0.0-20230321204456-577d146c0a5a
-go: downloading github.com/supranational/blst v0.3.11-0.20230124161941-ca03e11a3ff2
+go: downloading github.com/ethereum/c-kzg-4844 v0.0.0-20230407130613-fd0a51aa35bc
+go: downloading github.com/supranational/blst v0.3.11-0.20230406105308-e9dfc5ee724b
 88f1aea383b825371cb98acfbae6c81cce601a2e3129461c3c2b816409af8f3e5080db165fd327db687b3ed632153a62
 ```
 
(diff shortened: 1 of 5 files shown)


ppopth:
Is it canonical to have the tests in a separate package? I think this is a common Go problem, so Go community probably has a consensus on it.

I'm curious if we can find any Go project that does this. Currently I can find none.

jtraglia:
Good question. I don't believe there is consensus. It's a different strategy, see this link:
* https://stackoverflow.com/a/31443271

As for projects that do this, here are some instances I could find. It's way less common than I thought though, as it was pretty difficult to find these. I think I'm going to revert this change to keep things simple. It didn't accomplish my original goal anyway, to minimize dependencies.

* [go/src/hash/marshal_test.go](https://github.com/golang/go/blob/949fdd9f0dc53c32dc6a5a46ad7a85032df28055/src/hash/marshal_test.go#L9)
* [go/src/math/all_test.go](https://github.com/golang/go/blob/949fdd9f0dc53c32dc6a5a46ad7a85032df28055/src/math/all_test.go#L5)
* [go/src/strconv/atob_test.go](https://github.com/golang/go/blob/949fdd9f0dc53c32dc6a5a46ad7a85032df28055/src/strconv/atob_test.go#L5)
* [prometheus/storage/fanout_test.go](https://github.com/prometheus/prometheus/blob/8dba9163f1e923ec213f0f4d5c185d9648e387f0/storage/fanout_test.go)

jtraglia:
@ppopth Sorry if this is confusing, but I decided to repurpose this PR. Please re-review.

ppopth:
LGTM!

ppopth (approved):

PR #283, opened 2023-04-06
[Go] Allow minimal preset

jtraglia:
A small thing I've been meaning to do for a while. Only define `FIELD_ELEMENTS_PER_BLOB` if it isn't already defined. This allows users (developers) to specify their own, such as the minimal value: 4.


ppopth:
LGTM!

ppopth (approved):

PR #285, opened 2023-04-07
[Java] Change sprintf to snprintf

jtraglia:
This call could potentially cause a buffer overflow. All of the instances of it are safe, but I'd feel a lot better if this used the safe version (`snprintf`) instead, just in case.

### DIFF
--- a/bindings/java/c_kzg_4844_jni.c
+++ b/bindings/java/c_kzg_4844_jni.c
@@ -38,7 +38,7 @@ void throw_c_kzg_exception(JNIEnv *env, C_KZG_RET error_code, const char *messag
 void throw_invalid_size_exception(JNIEnv *env, const char *prefix, size_t size, size_t expected_size)
 {
   char message[100];
-  sprintf(message, "%s Expected %zu bytes but got %zu.", prefix, expected_size, size);
+  snprintf(message, sizeof(message), "%s Expected %zu bytes but got %zu.", prefix, expected_size, size);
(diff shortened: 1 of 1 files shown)


kevaundray (approved):

bindings/java/c_kzg_4844_jni.c
```diff
@@ -38,7 +38,7 @@ void throw_c_kzg_exception(JNIEnv *env, C_KZG_RET error_code, const char *messag
 void throw_invalid_size_exception(JNIEnv *env, const char *prefix, size_t size, size_t expected_size)
 {
   char message[100];
```
kevaundray, line 40: For reference: why is this 100?

jtraglia: I believe it's an arbitrary length that's enough for all of the current messages.

cc @StefanBratanov

kevaundray: Sounds good -- If messages grow beyond 100 characters, it will truncate so I think its fine regardless.

jtraglia: Agreed. Thanks for the review 🥇 will merge.

StefanBratanov: Yeah pretty much what Justin said, wasn't sure the best way to handle this

jtraglia: It's all good, that's how I would do it too. We can always raise that value later if we need to.

PR #287, opened 2023-04-10
[Rust] Remove blst to avoid double link issues, expose blst features

divagant-martian:
The big chunks of code are taken from blst without major modifications

### DIFF
--- a/bindings/rust/Cargo.toml
+++ b/bindings/rust/Cargo.toml
@@ -7,15 +7,30 @@ links = "ckzg"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
+
+# BLST Compilation:
+# By default, compile with ADX extension if the host supports it. Binary can be
+# executed on systems similar to the host.
 default = ["mainnet-spec"]
 mainnet-spec = []
 minimal-spec = []
+# BLST Compilation:
+# Compile in portable mode, without ISA extensions.
+# Binary can be executed on all systems.
+portable = []
+# BLST Compilation:
+# Enable ADX even if the host CPU doesn't support it.
+# Binary can be executed on Broadwell+ and Ryzen+ systems.
+force-adx = []
+# BLST Compilation:
+# Suppress multi-threading.
+# Engaged on wasm32 target architecture automatically.
+no-threads = []
 
 [dependencies]
 hex = "0.4.2"
 libc = "0.2"
 serde = { version = "1.0", features = ["derive"] }
-blst = { path = "../../blst/bindings/rust" }
 
 [dev-dependencies]
 criterion = "0.4"
@@ -26,6 +41,9 @@ serde_yaml = "0.9.17"
 [build-dependencies]
 bindgen = { git = "https://github.com/rust-lang/rust-bindgen" , rev = "0de11f0a521611ac8738b7b01d19dddaf3899e66" }
 cc = "1.0"
+
+[target.'cfg(target_env = "msvc")'.build-dependencies]
+glob = "0.3"
 
 [[bench]]
 name = "kzg_benches"
--- a/bindings/rust/build.rs
+++ b/bindings/rust/build.rs
@@ -1,8 +1,121 @@
 use std::env;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 const MAINNET_FIELD_ELEMENTS_PER_BLOB: usize = 4096;
 const MINIMAL_FIELD_ELEMENTS_PER_BLOB: usize = 4;
+
+/// Compiles blst.
+//
+// NOTE: This code is taken from https://github.com/supranational/blst `build.rs` `main`. The crate
+// is not used as a depedency to avoid double link issues on dependants.
+fn compile_blst(blst_base_dir: PathBuf) {
+    // account for cross-compilation [by examining environment variables]
+    let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap();
+    let target_arch = env::var("CARGO_CFG_TARGET_ARCH").unwrap();
+
+    if target_os.ne("none") && !env::var("BLST_TEST_NO_STD").is_ok() {
+        println!("cargo:rustc-cfg=feature=\"std\"");
+        if target_arch.eq("wasm32") {
+            println!("cargo:rustc-cfg=feature=\"no-threads\"");
+        }
+    }
+    println!("cargo:rerun-if-env-changed=BLST_TEST_NO_STD");
+
+    println!("Using blst source directory {}", blst_base_dir.display());
+
+    // Set CC environment variable to choose alternative C compiler.
+    // Optimization level depends on whether or not --release is passed
+    // or implied.
+
+    #[cfg(target_env = "msvc")]
+    if env::var("CARGO_CFG_TARGET_POINTER_WIDTH").unwrap().eq("32") && !env::var("CC").is_ok() {
+        match std::process::Command::new("clang-cl")
+            .arg("--version")
+            .output()
+        {
+            Ok(out) => {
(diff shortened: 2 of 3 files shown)


divagant-martian:
@jtraglia again no idea why I can't use github's review request

jtraglia:
Yeah idk either. It seems that I can only add a subset of people myself. Like I can't request a review from @ppopth even though he has write-access & is part of the organization. I'll make an issue with devops.

bindings/rust/build.rs
```diff
@@ -1,9 +1,122 @@
...
+        (true, true) => panic!("Cannot compile with both `portable` and `force-adx` features"),
+    }
+    if env::var("CARGO_CFG_TARGET_ENV").unwrap().eq("msvc") {
+        cc.flag("-Zl");
+    }
+    cc.flag_if_supported("-mno-avx") // avoid costly transitions
+        .flag_if_supported("-fno-builtin")
+        .flag_if_supported("-Wno-unused-function")
+        .flag_if_supported("-Wno-unused-command-line-argument");
+    if target_arch.eq("wasm32") {
```
jtraglia, line 91: Would the [default CFLAGS for blst](https://github.com/supranational/blst/blob/e9dfc5ee724b5b25d50a3b6226bee8c2c9d5e65d/build.sh#L26) work here?

```
CFLAGS=${CFLAGS:--O2 -fno-builtin -fPIC -Wall -Wextra -Werror}
```

Particularly, `-fPIC -Wall -Wextra -Werror`. Though `-fPIC` shouldn't be enabled on Windows systems.

bindings/rust/Cargo.toml
```diff
@@ -7,15 +7,30 @@ links = "ckzg"
...
+# Binary can be executed on all systems.
+portable = []
+# BLST Compilation:
+# Enable ADX even if the host CPU doesn't support it.
+# Binary can be executed on Broadwell+ and Ryzen+ systems.
+force-adx = []
+# BLST Compilation:
+# Suppress multi-threading.
+# Engaged on wasm32 target architecture automatically.
+no-threads = []
```
jtraglia, line 28: Nice. I see that, for the most part, this is [copied from blst](https://github.com/supranational/blst/blob/e9dfc5ee724b5b25d50a3b6226bee8c2c9d5e65d/bindings/rust/Cargo.toml#L26-L38).

bindings/rust/Cargo.toml
```diff
@@ -27,6 +42,9 @@ serde_yaml = "0.9.17"
 bindgen = { git = "https://github.com/rust-lang/rust-bindgen" , rev = "0de11f0a521611ac8738b7b01d19dddaf3899e66" }
 cc = "1.0"
 
+[target.'cfg(target_env = "msvc")'.build-dependencies]
+glob = "0.3"
+
```
jtraglia, line 47: Just curious, what's different with MSVC here?

bindings/rust/build.rs
```diff
@@ -1,9 +1,122 @@
...
+            }
+        }
+        (false, false) =>
+        {
+            #[cfg(target_arch = "x86_64")]
+            if target_arch.eq("x86_64") && std::is_x86_feature_detected!("adx") {
+                println!("Enabling ADX because it was detected on the host");
+                cc.define("__ADX__", None);
+            }
+        }
```
jtraglia, line 81: I think it would be nice if there was a way to force `__SHA__` and `__ARM_FEATURE_CRYPTO` too but I'm not sure it's worth the complexity. Unless you feel like it should be there too, don't worry about it. For reference, [this is where this is checked in blst](https://github.com/supranational/blst/blob/a7fd1f584d26b0ae6cdc427976ea1d8980f7e15d/src/sha256.h#L14-L15).

jtraglia (commented):
This looks really good. Just a few comments/questions before approving.

bindings/rust/Cargo.toml
divagant-martian: yeah very much

bindings/rust/build.rs
divagant-martian: most of these are enabled by `cc` by default, which is what the current version (having `blst` as a dependency in the `Cargo.toml` ) does

jtraglia: Ah, you're right, I didn't know Rust enabled those by default.

It looks like these aren't on by default though:

* [`warnings_into_errors`](https://docs.rs/cc/latest/cc/struct.Build.html#method.warnings_into_errors)
* `-fPIC`

bindings/rust/Cargo.toml
divagant-martian: it seems to be blst on windows is compiled with it's [own assembly files](https://github.com/supranational/blst/tree/master/build/win64). This is taken from [`blst rust bindings`](https://github.com/supranational/blst/blob/e9dfc5ee724b5b25d50a3b6226bee8c2c9d5e65d/bindings/rust/build.rs#L8-L26)

bindings/rust/build.rs
jtraglia: Eh disregard this. If blst isn't doing it either, it should be fine.

bindings/rust/Cargo.toml
jtraglia: Oh that makes sense. It needs glob to find the files. Thanks.

divagant-martian:
@jtraglia this code is copied from the blst `build.rs` (so what we currently use). I'm not really aware behind the decisions to use some or other compile flags on their side, tho I'm trying to dig in the blst history to understand where they come from

jtraglia:
Ah, yeah I see that now. Maybe I should bring up my concerns with `blst` instead.

I think this PR looks fine, let me know when you're ready & I'll merge it.

jtraglia (approved):

divagant-martian:
all good on my side.  I should note tho, I _think_ we could avoid having to maintain this code in sync if we weren't vendoring blst but declaring it as a depedency from crates.io and if the `blst` crate had a release. From the commit that upgraded the version to `0.3.10` there have been 65 commits to date in the `blst` main repository, which seem to include expansion of the headers. I'm not aware if using the submoduled one is something bindings *MUST* do, or simply there to facilitate this

divagant-martian:
yeah, just like we were doing before. The blst crate exposes the headers via environmental variables

bindings/rust/build.rs
```diff
@@ -20,9 +133,11 @@ fn main() {
 
     eprintln!("Using FIELD_ELEMENTS_PER_BLOB={}", field_elements_per_blob);
 
-    // Obtain the header files exposed by blst-bindings' crate.
-    let blst_headers_dir =
-        std::env::var_os("DEP_BLST_BINDINGS").expect("BLST exposes header files for bindings");
```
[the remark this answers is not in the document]
jtraglia: Yep you're right, forgot about that.

jtraglia:
Alright, sweet. Let's add this logic. I think it's a good idea.

divagant-martian:
What way are you suggesting we proceed ?

jtraglia:
Maybe I misunderstood part of what you said, but I was referring to this:

> To be clear, if the project is using blst >=0.3.11 use that instead of the submodule.

Where "project" is the project using these bindings.

To do that, somehow you'd falsely mark the submodule'd blst as `0.3.11`.

divagant-martian:
Yes, I mean a project using these bindings. For that to work, there must be a `blst` release, since it's the only way to allow users to find a compatible version. I don't think we can mark the submodule's blst as `0.3.11`and even if we did, these wouldn't be available to users given that the source is different

jtraglia:
Okay haha, yeah I misunderstood. Sorry about that. When `0.3.11` comes out, let's see if it's possible to remove this manual blst compilation and let cargo do version resolution. Going to merge this PR now.

PR #289, opened 2023-04-24
Export methods; test on multiple platforms

flcl42:
Windows build does not contain proper exports:

```batch
dumpbin /EXPORTS "ckzg.dll"
...
1    0 00001080 free_trusted_setup_wrap
2    1 00001000 load_trusted_setup_wrap
...
```

happens most likely because [wrapper header file](https://github.com/ethereum/c-kzg-4844/blob/main/src/c_kzg_4844.h#L188) cannot change methods' signatures already defined in [c_kzg_4844.h](https://github.com/ethereum/c-kzg-4844/blob/main/src/c_kzg_4844.h#L188)

# Changes
- Fix exports on Windows
- Run unit tests on different platforms
- Support mingw, powershell and cmd on Windows

# Remarks

DLLEXPORT is empty for any case but windows C# wrapper build

### DIFF
--- a/src/c_kzg_4844.h
+++ b/src/c_kzg_4844.h
@@ -32,14 +32,6 @@
 extern "C" {
 #endif
 
-/**
- * Used to modify the signatures of the interface functions by the wrappers.
- * Required to export the functions when it is compiled as a dll on Windows.
- */
-#ifndef DLLEXPORT
-#define DLLEXPORT
-#endif
-
 ///////////////////////////////////////////////////////////////////////////////
 // Macros
 ///////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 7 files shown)


src/c_kzg_4844.h
```diff
@@ -32,6 +32,10 @@
 extern "C" {
 #endif
 
+#ifndef DLLEXPORT
+#define DLLEXPORT
+#endif
```
asn-d6, line 37: Hm.. This feels slightly dirty to have on the core c_kzg_4844.h file. Let's at least document what this does.

asn-d6:
Are you aware of any other C# bindings projects that do a similar DLLEXPORT macro trick? I would be curious to see how other projects manage this.

src/c_kzg_4844.h
flcl42: Fully agree, it's ugly but I've found no other way. A comment has been added

flcl42: If I initialize it in the common file, other bindings will have more exports than probably expected by them, not sure if it's OK.

jtraglia: Ah okay. That's a good reason not to.

flcl42:
Good news @asn-d6 @jtraglia it works now via def files without macro!

jtraglia:
Excellent. That looks much better IMO 👍

jtraglia (approved):

PR #295, opened 2023-04-27
Support for Windows cross-compilation from Linux

zah:
Since `DirSep` matches the default separator of the target OS, it's not suitable for performing path operations during compile-time on the host OS.

The particular fix here solves the issue for cross-compiling a Windows build from a Linux host, but Nim seems to need a more general solution for detecting the host OS during compilation, so a host OS specific separator can be used when deriving paths from `currentSourcePath`.

### DIFF
--- a/bindings/nim/kzg_abi.nim
+++ b/bindings/nim/kzg_abi.nim
@@ -5,14 +5,14 @@
 import
   std/[strformat, strutils]
 
-from os import DirSep
+from os import DirSep, AltSep
 
 const
   # FIELD_ELEMENTS_PER_BLOB is overrideable from
   # compiler switch -d: or --define:
   FIELD_ELEMENTS_PER_BLOB* {.intdefine.} = 4096
   # kzgPath: c-kzg-4844 project path, removing 3 last elem
-  kzgPath  = currentSourcePath.rsplit(DirSep, 3)[0] & "/"
+  kzgPath  = currentSourcePath.rsplit({DirSep, AltSep}, 3)[0] & "/"
   blstPath = kzgPath & "blst/"
   srcPath  = kzgPath & "src/"
   bindingsPath = blstPath & "bindings"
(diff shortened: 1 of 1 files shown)


bindings/nim/kzg_abi.nim
```diff
@@ -5,14 +5,14 @@
...
   std/[strformat, strutils]
 
-from os import DirSep
+from os import DirSep, AltSep
 
 const
   # FIELD_ELEMENTS_PER_BLOB is overrideable from
   # compiler switch -d: or --define:
-  FIELD_ELEMENTS_PER_BLOB* {.intdefine.} = 4096
+  FIELD_ELEMENTS_PER_BLOB* {.strdefine.} = 4096
```
asn-d6, line 13: This was just changed here: https://github.com/ethereum/c-kzg-4844/pull/294

Before merging this I think the rest of the nimbus team should review this to make sure that it's final.

zah: I'm a member of the Nimbus team, but this change was accidental. Will remove it from the PR.

zah: The fix has been pushed.

bindings/nim/kzg_abi.nim
```diff
@@ -5,14 +5,14 @@
...
-from os import DirSep
+from os import DirSep, AltSep
 
 const
   # FIELD_ELEMENTS_PER_BLOB is overrideable from
   # compiler switch -d: or --define:
   FIELD_ELEMENTS_PER_BLOB* {.intdefine.} = 4096
   # kzgPath: c-kzg-4844 project path, removing 3 last elem
-  kzgPath  = currentSourcePath.rsplit(DirSep, 3)[0] & "/"
+  kzgPath  = currentSourcePath.rsplit({DirSep, AltSep}, 3)[0] & "/"
```
jtraglia, line 15: I believe the `"/"` appended at the end is platform specific, right? Should it be replaced with `DirSep`, or in the case of cross-compiling for Windows `AltSep`?

jangko: Both backward and forward slash will works on Windows, combination of both also works. The literal "/" will works for cross-compiling for Windows.
If we use `DirSep` or `AltSep` it will become a problem because Nim will translate it into host OS separator, not target  OS separator. IIRC there is a lengthy discussion and various reasons came up why adding cross compilation features to Nim compiler(especially about path handling) is difficult.

zah: No, the current solution is best because it works correctly on both Linux and Windows, both locally and when cross-compiling. We rely on the fact that Windows supports `/` as a separator.

jtraglia: Okay gotcha, thanks guys.

jtraglia (approved):
LGTM 👍

PR #296, opened 2023-05-03
Add new validate_kzg_g1 tests

jtraglia:
Adds the new deserializing G1 tests from bls12-381-tests, which these tests mirror:
* https://github.com/ethereum/bls12-381-tests/blob/6c5229a9571703ed4b47e08f8611b6a1ab389e15/main.py#L790-L818


asn-d6:
Thanks!

PR #299, opened 2023-05-09
Speed up loading trusted setup

jtraglia:
This PR speeds up loading the trusted setup. It does this by:

* Swapping out the G1 points (in monomial form) with G1 Lagrange points.
  * Both are provided in the [trusted setup in the specs](https://github.com/ethereum/consensus-specs/blob/dev/presets/mainnet/trusted_setups/testing_trusted_setups.json).
  * This will require clients update their trusted setup files.
  * Another benefit of this, we get to remove the `fft` functions & tests.
* When loading the trusted setup, replace `validate_kzg_g1` with less expensive alternative.
  * Now it just does `blst_p1_uncompress` and `blst_p1_from_affine`. 
  * The `blst_p1_in_g1` (subgroup) check takes about 170ms on my system.

On my system, this results in 28x speed-up (2050ms -> 71ms).

Without the extra extra optimization (replacing `validate_kzg_g1`), it's an 8x speed-up (2050ms -> 250ms).


src/c_kzg_4844.c
```diff
@@ -1773,15 +1701,17 @@ C_KZG_RET load_trusted_setup(
     if (ret != C_KZG_OK) goto out_error;
     ret = new_g2_array(&out->g2_values, n2);
     if (ret != C_KZG_OK) goto out_error;
-    ret = new_g1_array(&g1_projective, n1);
-    if (ret != C_KZG_OK) goto out_error;
 
     /* Convert all g1 bytes to g1 points */
     for (i = 0; i < n1; i++) {
```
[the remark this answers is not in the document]
asn-d6: neat!

src/c_kzg_4844.c
```diff
@@ -1743,9 +1671,9 @@ static C_KZG_RET new_fft_settings(FFTSettings *fs, unsigned int max_scale) {
  * @remark Free after use with free_trusted_setup().
  *
  * @param[out] out      Pointer to the stored trusted setup data
- * @param[in]  g1_bytes Array of G1 points
+ * @param[in]  g1_bytes Array of G1 points in Lagrange form
  * @param[in]  n1       Number of `g1` points in g1_bytes
- * @param[in]  g2_bytes Array of G2 points
+ * @param[in]  g2_bytes Array of G2 points in monomial form
```
ppopth, line 1676: Is it really called monomial form? I don't have a strong name for this, but I have some options.

1. coefficient form?
2. normal form?

I think I would go for "normal form". The monomial one is quite confusing to me.

jtraglia: That's what @kevaundray refers to them in go-kzg. For example: [this search](https://github.com/search?q=repo%3Acrate-crypto%2Fgo-kzg-4844%20monomial&type=code).

I'm not sure if it's the most accurate name. @asn-d6 @dankrad thoughts?

asn-d6: IMO either monomial form or coefficient form is fine in this context.

jtraglia:
Nice, thanks! Will review your PR & pull it into this branch soon.

src/c_kzg_4844.c
```diff
@@ -1755,37 +1641,46 @@ C_KZG_RET load_trusted_setup(
...
 
-    out->fs = NULL;
     out->g1_values = NULL;
     out->g2_values = NULL;
 
     /* Sanity check in case this is called directly */
     CHECK(n1 == TRUSTED_SETUP_NUM_G1_POINTS);
     CHECK(n2 == TRUSTED_SETUP_NUM_G2_POINTS);
 
+    /* It's the smallest power of 2 >= n1 */
```
asn-d6, line 1653: Why did we change the position of this inside the func?

src/c_kzg_4844.c
```diff
@@ -1755,37 +1641,46 @@ C_KZG_RET load_trusted_setup(
...
+    unsigned int max_scale = 0;
+    while ((1ULL << max_scale) < n1)
+        max_scale++;
+
+    out->max_width = 1ULL << max_scale;
```
ppopth, line 1658: I probably should assign to `out->max_width` at the end of the function instead to avoid mutating the output when the function fails?

asn-d6: No strong opinion. As long as we document it, I think it's fine to mutate even when the function fails.

src/c_kzg_4844.c
```diff
@@ -1658,7 +1658,8 @@ C_KZG_RET load_trusted_setup(
     out->max_width = 1ULL << max_scale;
 
     /* Allocate all of our arrays */
-    ret = new_fr_array(&out->roots_of_unity, out->max_width);
+    /* We need max_width+1 when expanding */
+    ret = new_fr_array(&out->roots_of_unity, out->max_width + 1);
```
ppopth, line 1660: Yeah, I forgot that this is a consequence of https://github.com/jtraglia/c-kzg-4844/pull/2/commits/66b358c10801bf3e04517fb4033c2b33e43d495d

jtraglia: No worries. I'm actually going to revert back to the old method, using an intermediate array for the expanded roots. I think that would be the better solution long term.

src/c_kzg_4844.c
```diff
@@ -1796,29 +1705,18 @@ C_KZG_RET load_trusted_setup(
...
-    ret = fft_g1(out->g1_values, g1_projective, true, n1, out->fs);
+    ret = new_roots_of_unity(out->roots_of_unity, max_scale);
     if (ret != C_KZG_OK) goto out_error;
     ret = bit_reversal_permutation(out->g1_values, sizeof(g1_t), n1);
     if (ret != C_KZG_OK) goto out_error;
 
     goto out_success;
 
 out_error:
-    c_kzg_free(out->fs);
```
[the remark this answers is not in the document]
jtraglia: Fixed.

src/c_kzg_4844.c
```diff
@@ -1796,29 +1705,18 @@ C_KZG_RET load_trusted_setup(
...
     c_kzg_free(out->g1_values);
     c_kzg_free(out->g2_values);
```
asn-d6, line 1719:
```suggestion
    c_kzg_free(out->g1_values);
    c_kzg_free(out->g2_values);
    c_kzg_free(out->roots_of_unity);
```

jtraglia: Fixed 👍 good spot.

src/c_kzg_4844.c
```diff
@@ -34,6 +34,9 @@
 #define CHECK(cond) \
     if (!(cond)) return C_KZG_BADARGS
 
+/** Returns number of elements in a statically defined array. */
+#define NUM_ELEMENTS(a) sizeof(a) / sizeof(a[0])
```
asn-d6, line 38:
```suggestion
#define NUM_ELEMENTS(a) (sizeof(a) / sizeof(a[0]))
```

src/c_kzg_4844.c
jtraglia: Moved `free_kzg_settings` up & called that function instead.

src/c_kzg_4844.c
```diff
@@ -1796,29 +1723,17 @@ C_KZG_RET load_trusted_setup(
...
     ret = bit_reversal_permutation(out->g1_values, sizeof(g1_t), n1);
     if (ret != C_KZG_OK) goto out_error;
 
     goto out_success;
 
 out_error:
-    c_kzg_free(out->fs);
-    c_kzg_free(out->g1_values);
-    c_kzg_free(out->g2_values);
+    free_trusted_setup(out);
```
asn-d6, line 1735: This is slightly confusing because usually `free_*()` helpers also free the input pointer but the public func `free_trusted_setup()` doesn't. This is the correct behavior because this function should not free `out`, but it might confuse the reader.

No strong opinion whether explicit `c_kzg_free()` are better, but thought I should mention it.

src/c_kzg_4844.c
jtraglia: This will be set to zero if it fails now. Do you still want to document this?

src/c_kzg_4844.c
```diff
@@ -1661,79 +1589,37 @@ static C_KZG_RET expand_root_of_unity(
...
- * settings, and is a power of two by construction. The same settings may be
- * used to calculated FFTs of smaller power sizes.
- *
- * @remark As with all functions prefixed `new_`, this allocates memory that
- * needs to be reclaimed by calling the corresponding `free_` function. In
- * this case, free_fft_settings().
- *
- * @remark These settings may be used for FFTs on both field elements and G1
- * group elements.
+ * Initialise the roots of unity.
```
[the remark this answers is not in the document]
jtraglia: I've organized it so that it will only modify roots_of_unity_out if it's going to succeed.

src/c_kzg_4844.c
jtraglia: Hmm good point. Worthy of a comment?

asn-d6: Yes, a comment could do it.

src/c_kzg_4844.c
asn-d6: I think it's OK either way.

src/c_kzg_4844.c
jtraglia: Made a comment. Feel free to recommend something else.

asn-d6: Looks good. Thanks.

src/c_kzg_4844.c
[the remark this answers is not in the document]
jtraglia: Reverted that change. Will add documentation.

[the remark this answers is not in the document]
jtraglia: Added a simple remark.

bindings/node.js/lib/kzg.js
```diff
@@ -23,7 +23,7 @@ function transformTrustedSetupJson(filePath) {
   const setupText =
     bindings.FIELD_ELEMENTS_PER_BLOB +
     "\n65\n" +
-    trustedSetup.setup_G1.map((p) => p.substring(2)).join("\n") +
+    trustedSetup.setup_G1_lagrange.map((p) => p.substring(2)).join("\n") +
```
asn-d6, line 26: Kind of surprised this one line change is sufficient for the nodejs bindings. What happens to `setup_G1` now?

src/c_kzg_4844.c
```diff
@@ -1796,29 +1725,22 @@ C_KZG_RET load_trusted_setup(
...
-    while ((1ULL << max_scale) < n1)
-        max_scale++;
-
     /* Initialize the KZGSettings struct */
-    ret = c_kzg_malloc((void **)&out->fs, sizeof(FFTSettings));
-    if (ret != C_KZG_OK) goto out_error;
-    ret = new_fft_settings(out->fs, max_scale);
-    if (ret != C_KZG_OK) goto out_error;
-    ret = fft_g1(out->g1_values, g1_projective, true, n1, out->fs);
+    ret = new_roots_of_unity(out->roots_of_unity, max_scale);
```
asn-d6, line 1729: IMO we should rename `new_roots_of_unity()` to `get_roots_of_unity()` or `compute_roots_of_unity()`. That's because `new_*()` functions are usually constructor functions and this one doesn't really construct something new.

src/c_kzg_4844.c
```diff
@@ -1796,29 +1725,22 @@ C_KZG_RET load_trusted_setup(
         blst_p2_from_affine(&out->g2_values[i], &g2_affine);
     }
 
-    /* It's the smallest power of 2 >= n1 */
-    unsigned int max_scale = 0;
-    while ((1ULL << max_scale) < n1)
-        max_scale++;
-
     /* Initialize the KZGSettings struct */
```
asn-d6, line 1728: This comment now feels out of place:

```suggestion
    /* Compute roots of unity and permute the G1 trusted setup */
```

bindings/node.js/lib/kzg.js
jtraglia: The nodejs bindings convert the trusted setup from the specs to the c-kzg format. The `setup_G1` field will be unused now, like `setup_G1_lagrange` was and `roots_of_unity` are. I don't think we should modify that file.

src/c_kzg_4844.c
jtraglia: Good idea. Went with the `compute` one.

src/c_kzg_4844.c
jtraglia: Agreed. Fixed.

asn-d6:
LGTM!

src/c_kzg_4844.h
```diff
@@ -130,42 +130,18 @@ typedef enum {
...
-     * `reverse_roots_of_unity`. The redundant element is just there to simplify
-     * index calculations in some formulas.
-     */
-
-    /** Ascending powers of the root of unity, length `max_width + 1`. */
-    fr_t *expanded_roots_of_unity;
-    /** Descending powers of the root of unity, length `max_width + 1`. */
-    fr_t *reverse_roots_of_unity;
     /** Powers of the root of unity in bit-reversal permutation order, length
      * `max_width`. */
```
GottfriedHerold, line 139:
```suggestion
    /** Powers of the primitive root of unity determined by `SCALE2_ROOT_OF_UNITY` 
     *   in bit-reversal permutation order, length `max_width`. */
```
Keep the reference to SCALE2_ROOT_OF_UNITY, as everything is relative to that.

src/c_kzg_4844.c
```diff
@@ -1755,37 +1672,49 @@ C_KZG_RET load_trusted_setup(
...
+    out->max_width = 0;
+    out->roots_of_unity = NULL;
     out->g1_values = NULL;
     out->g2_values = NULL;
 
     /* Sanity check in case this is called directly */
     CHECK(n1 == TRUSTED_SETUP_NUM_G1_POINTS);
     CHECK(n2 == TRUSTED_SETUP_NUM_G2_POINTS);
 
+    /* It's the smallest power of 2 >= n1 */
```
GottfriedHerold, line 1686:
```suggestion
    /* 1<<max_scale is the smallest power of 2 >= n1 */
```
The previous wording sounded like max_scale (rather than 1<<max_scale) was a power of 2.

asn-d6:
Merging this. Thanks everyone!

zilm13:
I don't understand this change from the crypto side, but is there a way to check that we have new `trusted_setup` and throw if an old one provided? When using an old one I still produce some commitments and proofs, but they are not correct and fails on verification. So it would be great to throw if trusted_setup doesn't follow new expectations. Is it possible?

PR #301, opened 2023-05-09
Use g2 points from spec in minimal preset

jtraglia:
The g2 points in the minimal preset (`trusted_setup_4.txt`) had different points than the spec for `g2_points[4:]`. This was fine, because only the first 4 get used, but I think it's something we should fix regardless.

See minimal [testing_trusted_setups.json](https://github.com/ethereum/consensus-specs/blob/dev/presets/minimal/trusted_setups/testing_trusted_setups.json) in the specs.


asn-d6:
LGTM!

PR #302, opened 2023-05-11
Make C & C# tests use portable blst

jtraglia:
There are frequent test failures for these two languages. I think using the portable version of blst will fix this.


jtraglia:
There's a complicated mess between ckzg & blst CFLAGS. Not something I want to mess with now.

PR #305, opened 2023-05-18
Make BLS field elements big-endian

jtraglia:
This PR changes the endianness from little-endian to big-endian, as defined here:

* https://github.com/ethereum/consensus-specs/pull/3354


bindings/csharp/Ckzg.Bindings/Ckzg.cs
```diff
@@ -102,6 +102,7 @@ public static unsafe void BlobToKzgCommitment(Span<byte> commitment, ReadOnlySpa
         ThrowOnUninitializedTrustedSetup(ckzgSetup);
         ThrowOnInvalidLength(proof, nameof(proof), BytesPerProof);
         ThrowOnInvalidLength(blob, nameof(blob), BytesPerBlob);
+        ThrowOnInvalidLength(commitment, nameof(commitment), BytesPerCommitment);
```
jtraglia, line 105: Worth pointing out, this check was missing from the C# bindings.

Looks like the new reference tests caught it.

bindings/go/main_test.go
```diff
@@ -36,7 +36,7 @@ func getRandFieldElement(seed int64) Bytes32 {
 	// This leaves the last byte in fieldElementBytes as
 	// zero, which guarantees it's a canonical field element.
 	var fieldElementBytes Bytes32
-	copy(fieldElementBytes[:], bytes)
+	copy(fieldElementBytes[1:], bytes)
```
asn-d6, line 39: Hmm, what is this change again?

jtraglia: So a little above this we generate 31 random bytes. Previously, this would copy 31 bytes to the 32-byte field element and leave the last byte as zero... so we know it's valid. Now, since it's big-endian, the first byte should be zero. This changes the starting position of the copy so the first byte remains zero. If you want, I can add a comment or do it differently. For example, generate/copy 32 bytes and manually set the first one to zero.

bindings/go/main_test.go
```diff
@@ -36,7 +36,7 @@ func getRandFieldElement(seed int64) Bytes32 {
 	// This leaves the last byte in fieldElementBytes as
```
asn-d6, line 36:
```suggestion
	// This leaves the first byte in fieldElementBytes as
```

bindings/go/main_test.go
asn-d6: Ah that makes sense. I guess then we need https://github.com/ethereum/c-kzg-4844/pull/305/files#r1198905335 ?

jtraglia: Ah yes, thank you. Fixed.

asn-d6 (approved):
Looks great!

jtraglia:
This is in the spec now. Going to merge.

PR #306, opened 2023-05-25
[Go] Make UnmarshalText funcs safer

jtraglia:
Before this PR, these functions would panic if given a string fewer than 2 bytes. This PR addresses that issue & makes it a little easier to read. Also, fix some indentation nits for consistency.


asn-d6:
LGTM! Thanks!

PR #308, opened 2023-05-26
Check arg count in nodejs bindings

jtraglia:
This PRs adds an argument count check to each of the public functions. Previously, if someone passed in an invalid number of argument it would not handle it gracefully.

### DIFF
--- a/bindings/node.js/src/kzg.cxx
+++ b/bindings/node.js/src/kzg.cxx
@@ -164,6 +164,12 @@ inline Bytes48 *get_bytes48(
 Napi::Value LoadTrustedSetup(const Napi::CallbackInfo &info) {
     Napi::Env env = info.Env();
 
+    if (info.Length() != 1) {
+        Napi::Error::New(env, "Invalid number of arguments")
+            .ThrowAsJavaScriptException();
+        return env.Undefined();
+    }
+
     // Check if the trusted setup is already loaded
     KzgAddonData *data = env.GetInstanceData<KzgAddonData>();
     if (data->is_setup) {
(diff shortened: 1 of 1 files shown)


matthewkeil (commented):
We actually had argument length checks way back and removed them. It's a built in feature of node to auto validate too many args (more than the expected are just ignored and have no effect). It also auto protects against too few arguments and returns `Undefined` in the case an expected arg is not passed. That is checked for in all cases. As far as the args in the expected positions are fully validated for both types and values. There are also unit tests to validate everything. There should not be any need for length tests.

https://github.com/ethereum/c-kzg-4844/blob/5019e3a08d36ade6878292c6d241937d65c52200/bindings/node.js/test/kzg.test.ts#L240

jtraglia:
Yeah I think you're right. If I make a test for this, it will catch it.

image

matthewkeil:
Here is the line that catches too few args

https://github.com/nodejs/node-addon-api/blob/358b2d3b4f4cf0b3bdfc9897be888494ee36ae1d/napi-inl.h#L3798

matthewkeil:
> Yeah I think you're right. If I make a test for this, it will catch it.
> 
> image

As a note, those are TS args which happen before code execution.  So it will throw from TS but there are also runtime checks in the event the library is running as JS where there is no transpilation.

Should be covered for all cases and 100% safe (type safety and runtime safety)

jtraglia:
Sure, let's add some unit tests just to alleviate any concerns.

asn-d6:
So I guess there is another PR pending about the unit tests?

PR #309, opened 2023-05-31
Sanity check that trusted setup was loaded in Lagrange form

asn-d6:
We received reports from client devs that they loaded old trusted setup files (with G1 in monomial form) and KZG did not work but without any errors from c-kzg. This PR adds a basic sanity check that the trusted setup was loaded with G1 in lagrange form.


jtraglia (approved):
Nice, thanks for this 🥇 LGTM!

PR #313, opened 2023-06-06
Add macros to prefix public methods with arbitrary strings

asn-d6:
This is a feature request from https://github.com/ethereum/c-kzg-4844/issues/312 .

This is still a PoC and more love is needed if we decide we want this.

PR #317, opened 2023-06-09
Use LIB_PREFIX in Rust bindings

michaelsproul:
Extension of #313 to the Rust bindings.


jtraglia (approved):
So I don't love the new macros, but it's probably the best option we have. This PR is a good tradeoff of usability & minor changes to the codebase post-audit. I made a commit that fixes clang-format's complaints and alters the comments a little to match the style of others. The CI check that is failing is a known problem which we can ignore for now. LGTM 👍

michaelsproul:
Still WIP on the Lighthouse integration unfortunately. We've hit a few linking issues that look like they might require further changes

michaelsproul:
Ok, we've got it working reliably in Lighthouse now. I've pushed two commits here that seemed to help:

1. 5f3a7c1f35a5c4982bc87df75bb510620cec5241 moves the bindgen-generated Rust code to the Cargo build directory, rather than the source directory. This is the [standard practice](https://rust-lang.github.io/rust-bindgen/tutorial-3.html) for bindgen and helps prevent strange caching behaviour when the `c-kzg` crate is compiled with different mainnet/minimal specs.
2. 6449e2cccc13b9bf2358322f3a5539fbc6613d6d removes all unprefixed symbols from the generated object files. We were getting some link errors about symbols like `blob_verify_kzg_proof` being duplicated. These errors seemed to only occur occasionally, and only on Linux, and could actually have been a manifestation of (1). If you prefer not to make this change I can try reverting it and testing with (1) alone.

divagant-martian:
I added the generated bindings to the source to be able to see them. So I don't think anyone else has an attachment to that setup. I'm fine with moving them to the build dir

jtraglia:
Hey @michaelsproul I made a PR to your branch (https://github.com/michaelsproul/c-kzg-4844/pull/1) with some changes I'd like to make. I think this will simplify things a bit, but there were a lot of changes so I didn't feel comfortable committing them directly. Thoughts?

jtraglia:
Going to merge. I believe it's ready 👍

PR #322, opened 2023-06-28
Add band-aid fix for Windows benchmark issue

jtraglia:
While we wait for a solution, I propose this band-aid fix which should mask the issue. @pawanjay176 mentioned that the benchmarks didn't crash with opt-level set to zero, so that's what this PR does.

### DIFF
--- a/bindings/rust/Cargo.toml
+++ b/bindings/rust/Cargo.toml
@@ -48,3 +48,8 @@ glob = "0.3"
 [[bench]]
 name = "kzg_benches"
 harness = false
+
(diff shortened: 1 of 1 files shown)


divagant-martian (approved):
This is a good enough tradeoff

pawanjay176 (approved):
I'm fine with this too. 👍 
I'm curious if the status access violation happens only for the benchmarks. I haven't been able to reproduce the issue with normal release tests or running binaries with same 
parameters that crashes on benches.

Keen to try running lighthouse on windows in the next devnet and see if it crashes there.

PR #323, opened 2023-07-03
Use boxed blobs in Rust bindings

jtraglia:
Right now, this is a proof of concept with code that might not be production quality.


jtraglia:
I want to take a slightly different approach. Closing.

Issue #318, opened 2023-06-09
Broken CI in rust-windows because of rust-1.7.0


asn-d6:
See https://github.com/ethereum/c-kzg-4844/pull/311#issuecomment-1582439003

jtraglia:
Temporarily fixed/masked the problem with this PR:

* https://github.com/ethereum/c-kzg-4844/pull/322

When there's a new release of rust/criterion that fixes the problem, please revert that PR.

PR #325, opened 2023-07-07
Use boxes & refs in Rust bindings

jtraglia:
This PR fixes the issue with Windows & should be overall more efficient.

* When dealing with individual blobs, box them (store them on the heap).
  * This fixes the segmentation fault crash on Windows on 4844-devnet-7.
  * My theory is that these (relatively large) blobs were overflowing the stack.
* For all of the KZG interface functions, take references.
  * I feel like this should have done this a while ago.
  * It should be more efficient as it saves us from cloning.

Note, this does **not** fix the benchmark issue on Windows:

* #318

### DIFF
--- a/bindings/rust/Cargo.toml
+++ b/bindings/rust/Cargo.toml
@@ -50,3 +50,8 @@ glob = "0.3"
 [[bench]]
 name = "kzg_benches"
 harness = false
+
+# The benchmarks crash on Windows with Rust 1.70. This is a band-aid fix for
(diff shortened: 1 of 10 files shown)


jtraglia:
Actually, dang. Hold off on reviewing this. This PR still has the Windows segfault issue on 4844-devnet-7. I'll need to look into why the "boxed-blobs" branch works and this one doesn't.

jtraglia:
Still unable to get this working. Will re-open if something changes.

PR #329, opened 2023-08-08
Adding constant for bytes per g1/g2 points

StefanBratanov:
Just a small nit for the java binding. We use those constants in Teku, so it makes sense to have them in the lib.


asn-d6 (approved):
LGTM! Thanks!

PR #332, opened 2023-08-09
Test Go bindings on all platforms

jtraglia:
* Test the Go bindings on Linux, macOS, and Windows.
* Use the latest stable version of Go instead of a static version.


asn-d6 (approved):
Thanks!

PR #335, opened 2023-08-11
Force use of portable blst in Nim bindings

jtraglia:
* Ignore `build` directory that `nimble test` creates.
* If not using an external blst, pass flag to compiler to build portable version of blst.


asn-d6 (approved):
Thanks!

PR #337, opened 2023-08-11
Force use of portable blst in C

jtraglia:
* Add a new `BLST_BUILDSCRIPT_FLAGS` variable which defines blst's portable macro.
* Pass that new variable to blst's build script.


asn-d6 (approved):
LGTM!

PR #338, opened 2023-08-11
Force use of portable blst in Rust bindings

jtraglia:
* Remove features for `portable` and `force-adx` which were used by blst.
  * Leave the `no-threads` feature since that's still useful.
* Remove the big `match` block which checks the `portable` and `force-adx` features.
* Add non-configurable define for `__BLST_PORTABLE__` when building blst.

### DIFF
--- a/bindings/rust/Cargo.toml
+++ b/bindings/rust/Cargo.toml
@@ -11,20 +11,10 @@ license = "Apache-2.0"
 
 [features]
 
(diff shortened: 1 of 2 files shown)


jtraglia:
I'd also appreciate a review from @pawanjay176, but I can't request it for some reason.

asn-d6 (approved):
LGTM!

PR #340, opened 2023-08-13
Use blst's generator functions

jtraglia:
Minor complexity improvement. Rather than redefining the G1/G2 generators, use blst's `blst_p{1,2}_generator` functions. These return pointers which will most likely get plugged in during comptime, so there's no performance penalty.


asn-d6 (approved):
LGTM!

PR #341, opened 2023-08-15
Remove go bindings example

jtraglia:
I would like to remove this example project. It's annoying to up-keep. Right now it's still using little-endian fields.


asn-d6 (approved):
LGTM! Thanks!

PR #342, opened 2023-08-15
feat: add serde impls to Blob and Bytes48

Rjected:
This adds serde `Serialize` and `Deserialize` implementations that work with a `0x` prefix.


asn-d6:
Hey. Can you tell us why you need the serde impls and the from_hex() is not sufficient for your needs? Is it just for testing purposes?

Rjected:
> Hey. Can you tell us why you need the serde impls and the from_hex() is not sufficient for your needs? Is it just for testing purposes?

I'd like to use `Blob` and `Bytes48` in other types in reth and not have to write a manual `Serialize` or `Deserialize` impl for these higher level types, for example using this branch lets us `derive` here:
https://github.com/paradigmxyz/reth/blob/bfa130d1875f1e1510e181e09fced33578d70563/crates/primitives/src/transaction/eip4844.rs#L547-L557

We provide these traits on almost all basic types so it's possible for API consumers to use the serde ecosystem in their own applications. With the serde impl, it would also become possible to use these types directly for RPC serialization instead of requiring conversion to another type that uses `Bytes`:
https://github.com/paradigmxyz/reth/blob/bfa130d1875f1e1510e181e09fced33578d70563/crates/rpc/rpc-types/src/eth/engine/payload.rs#L215-L221

jtraglia (commented):
Makes sense. For completeness, could we add serde impls for `Bytes32` too?

Rjected:
> Makes sense. For completeness, could we add serde impls for Bytes32 too?

just added!

jtraglia (approved):
Great. Thanks for adding that. LGTM 👍

asn-d6 (approved):

asn-d6:
Thanks!

PR #351, opened 2023-08-23
Use blst crate for Rust blst dep

michaelsproul:
We were getting nasty linker errors in Lighthouse due to different versions of the `blst` library being linked by:

- `crypto/bls`: our BLS wrapper that depends on the `blst` crate from `crates.io`.
- `c_kzg/c_kzg_min`: mainnet/minimal variants of c-kzg which depend on their own `libblst.a` which is built at compile-time.

As well as causing link errors when the `blst` versions get out of sync, the current setup also wastes compilation time by compiling `blst` 3 times.

I think the changes in this PR improve the situation by using a single copy of `blst` throughout. Due to the [`links` attribute](https://doc.rust-lang.org/cargo/reference/build-scripts.html#the-links-manifest-key) in `blst`'s `Cargo.toml`, Cargo will now error if we try to compile with multiple copies of the `blst` crate. We can also make semver-compatible updates to `blst` downstream without having to bump the `blst` version in `c-kzg`, so long as the headers are compatible with the ones vendored in `c-kzg`.

### DIFF
--- a/bindings/rust/Cargo.lock
+++ b/bindings/rust/Cargo.lock
@@ -63,6 +63,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "630be753d4e58660abd17930c71b647fe46c27ea6b63cc59e1e3851406972e42"
 
 [[package]]
+name = "blst"
+version = "0.3.11"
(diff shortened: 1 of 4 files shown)


jtraglia (approved):
Nice. LGTM 👍

PR #352, opened 2023-08-24
Runtime field count

jtraglia:
With this PR, one build of c-kzg-4844 will work with both mainnet and minimal trusted setups. If merged, I will update the bindings in separate PRs. Until then, the CI tests for the bindings are expected to fail.
 
* Remove the `FIELD_ELEMENTS_PER_BLOB` macro (which was either 4096 or 4).
  * This is now accessible via `kzg_settings->field_elements_per_blob` after loading the trusted setup.
* Allocate blobs and polynomials on-demand on the heap.
  * We cannot allocate them on the stack anymore, since the count isn't a constant.
  * This will also reduce stack usage, which should prevent stack overflows.
* Remove type definitions for `Blob` and `Polynomial`.
  * It will now be up to the bindings to define the blob type.
  * I don't want to do `typedef uint8_t Blob` because that could be confusing.
* Remove the `LIB_PREFIX` hack for supporting both builds.
  * This is no longer necessary. 

For the tests:

* Use new `MAINNET` and `MINIMAL` macros to determine which to test.
* Replace stack allocated blobs/polynomials with heap allocated ones.

### DIFF
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -48,15 +48,6 @@
     } while (0)
 
 ///////////////////////////////////////////////////////////////////////////////
-// Types
-///////////////////////////////////////////////////////////////////////////////
-
-/** Internal representation of a polynomial. */
-typedef struct {
-    fr_t evals[FIELD_ELEMENTS_PER_BLOB];
-} Polynomial;
-
-///////////////////////////////////////////////////////////////////////////////
 // Constants
 ///////////////////////////////////////////////////////////////////////////////
 
@@ -600,11 +591,11 @@ static C_KZG_RET bytes_to_kzg_proof(g1_t *out, const Bytes48 *b) {
  * @param[out] p    The output polynomial (array of field elements)
  * @param[in]  blob The blob (an array of bytes)
  */
-static C_KZG_RET blob_to_polynomial(Polynomial *poly, const uint8_t *blob) {
+static C_KZG_RET blob_to_polynomial(fr_t *poly, const uint8_t *blob) {
     C_KZG_RET ret;
     for (size_t i = 0; i < FIELD_ELEMENTS_PER_BLOB; i++) {
         ret = bytes_to_bls_field(
-            &poly->evals[i], (Bytes32 *)&blob[i * BYTES_PER_FIELD_ELEMENT]
+            &poly[i], (Bytes32 *)&blob[i * BYTES_PER_FIELD_ELEMENT]
         );
         if (ret != C_KZG_OK) return ret;
     }
@@ -787,7 +778,7 @@ static void compute_powers(fr_t *out, const fr_t *x, uint64_t n) {
  * @param[in]  s    The trusted setup
  */
 static C_KZG_RET evaluate_polynomial_in_evaluation_form(
-    fr_t *out, const Polynomial *poly, const fr_t *x, const KZGSettings *s
+    fr_t *out, const fr_t *poly, const fr_t *x, const KZGSettings *s
 ) {
     C_KZG_RET ret;
     fr_t tmp;
@@ -809,7 +800,7 @@ static C_KZG_RET evaluate_polynomial_in_evaluation_form(
          * would divide by zero otherwise.
          */
         if (fr_equal(x, &roots_of_unity[i])) {
-            *out = poly->evals[i];
+            *out = poly[i];
             ret = C_KZG_OK;
             goto out;
         }
@@ -822,7 +813,7 @@ static C_KZG_RET evaluate_polynomial_in_evaluation_form(
     *out = FR_ZERO;
     for (i = 0; i < FIELD_ELEMENTS_PER_BLOB; i++) {
         blst_fr_mul(&tmp, &inverses[i], &roots_of_unity[i]);
-        blst_fr_mul(&tmp, &tmp, &poly->evals[i]);
+        blst_fr_mul(&tmp, &tmp, &poly[i]);
         blst_fr_add(out, out, &tmp);
     }
     fr_from_uint64(&tmp, FIELD_ELEMENTS_PER_BLOB);
@@ -849,10 +840,10 @@ out:
  * @param[in]  s    The trusted setup
  */
 static C_KZG_RET poly_to_kzg_commitment(
-    g1_t *out, const Polynomial *poly, const KZGSettings *s
+    g1_t *out, const fr_t *poly, const KZGSettings *s
 ) {
     return g1_lincomb_fast(
-        out, s->g1_values, (const fr_t *)(&poly->evals), FIELD_ELEMENTS_PER_BLOB
+        out, s->g1_values, poly, FIELD_ELEMENTS_PER_BLOB
     );
 }
 
@@ -867,15 +858,20 @@ C_KZG_RET BLOB_TO_KZG_COMMITMENT(
     KZGCommitment *out, const uint8_t *blob, const KZGSettings *s
 ) {
     C_KZG_RET ret;
-    Polynomial poly;
+    fr_t *poly = NULL;
     g1_t commitment;
 
-    ret = blob_to_polynomial(&poly, blob);
-    if (ret != C_KZG_OK) return ret;
-    ret = poly_to_kzg_commitment(&commitment, &poly, s);
-    if (ret != C_KZG_OK) return ret;
+    ret = new_fr_array(&poly, FIELD_ELEMENTS_PER_BLOB);
+    if (ret != C_KZG_OK) goto out;
+    ret = blob_to_polynomial(poly, blob);
+    if (ret != C_KZG_OK) goto out;
+    ret = poly_to_kzg_commitment(&commitment, poly, s);
+    if (ret != C_KZG_OK) goto out;
     bytes_from_g1(out, &commitment);
-    return C_KZG_OK;
+
+out:
+    c_kzg_free(poly);
+    return ret;
 }
 
 /* Forward function declaration */
@@ -972,7 +968,7 @@ static C_KZG_RET verify_kzg_proof_impl(
 static C_KZG_RET compute_kzg_proof_impl(
     KZGProof *proof_out,
     fr_t *y_out,
-    const Polynomial *poly,
+    const fr_t *poly,
     const fr_t *z,
     const KZGSettings *s
 );
@@ -995,18 +991,21 @@ C_KZG_RET COMPUTE_KZG_PROOF(
     const KZGSettings *s
 ) {
     C_KZG_RET ret;
-    Polynomial poly;
+    fr_t *poly = NULL;
     fr_t frz, fry;
 
-    ret = blob_to_polynomial(&poly, blob);
+    ret = new_fr_array(&poly, FIELD_ELEMENTS_PER_BLOB);
+    if (ret != C_KZG_OK) goto out;
+    ret = blob_to_polynomial(poly, blob);
     if (ret != C_KZG_OK) goto out;
     ret = bytes_to_bls_field(&frz, z_bytes);
     if (ret != C_KZG_OK) goto out;
-    ret = compute_kzg_proof_impl(proof_out, &fry, &poly, &frz, s);
+    ret = compute_kzg_proof_impl(proof_out, &fry, poly, &frz, s);
     if (ret != C_KZG_OK) goto out;
     bytes_from_bls_field(y_out, &fry);
 
 out:
+    c_kzg_free(poly);
     return ret;
 }
 
@@ -1024,19 +1023,19 @@ out:
 static C_KZG_RET compute_kzg_proof_impl(
     KZGProof *proof_out,
     fr_t *y_out,
-    const Polynomial *poly,
+    const fr_t *poly,
     const fr_t *z,
     const KZGSettings *s
 ) {
     C_KZG_RET ret;
     fr_t *inverses_in = NULL;
     fr_t *inverses = NULL;
+    fr_t *q_poly = NULL;
 
     ret = evaluate_polynomial_in_evaluation_form(y_out, poly, z, s);
     if (ret != C_KZG_OK) goto out;
 
     fr_t tmp;
-    Polynomial q;
     const fr_t *roots_of_unity = s->roots_of_unity;
     uint64_t i;
     /* m != 0 indicates that the evaluation point z equals root_of_unity[m-1] */
@@ -1045,6 +1044,8 @@ static C_KZG_RET compute_kzg_proof_impl(
     ret = new_fr_array(&inverses_in, FIELD_ELEMENTS_PER_BLOB);
     if (ret != C_KZG_OK) goto out;
     ret = new_fr_array(&inverses, FIELD_ELEMENTS_PER_BLOB);
+    if (ret != C_KZG_OK) goto out;
+    ret = new_fr_array(&q_poly, FIELD_ELEMENTS_PER_BLOB);
     if (ret != C_KZG_OK) goto out;
 
     for (i = 0; i < FIELD_ELEMENTS_PER_BLOB; i++) {
@@ -1055,7 +1056,7 @@ static C_KZG_RET compute_kzg_proof_impl(
             continue;
         }
         // (p_i - y) / (ω_i - z)
-        blst_fr_sub(&q.evals[i], &poly->evals[i], y_out);
+        blst_fr_sub(&q_poly[i], &poly[i], y_out);
         blst_fr_sub(&inverses_in[i], &roots_of_unity[i], z);
     }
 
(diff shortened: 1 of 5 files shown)


ppopth:
I'm not quite convinced that the `LIB_PREFIX` thing is a hack. I heard that the problem is that lighthouse has to use two dependent crates for this and it has to add another slash for the second crate, as follows.
```rust
c-kzg = { git = "https://github.com/ethereum/c-kzg-4844", rev = "fa3c62989527073fdce8b2138bb27a52bb2407c5" , features = ["mainnet-spec"]}
c_kzg_min = { package = "c-kzg", git = "https://github.com/ethereum//c-kzg-4844", rev = "fa3c62989527073fdce8b2138bb27a52bb2407c5", features = ["minimal-spec"], optional = true }
````

I think, this one is unnecessary because you can just use one crate for both mainnet and minimal, instead of using two. You can probably put the mainnet in one module and the minimal in another module.

Can you give a reason why you think we should eliminate the `LIB_PREFIX` thing?

For the stack overflow problem in Rust, I think we don't have to remove the Blob definition. It's better to keep it as is. If some bindings have the problem with it, they can just ignore it and redefine that type.

For example, the Rust binding can redefine the Blob type as some kind of `Vec<u8>` and, when it wants to use the c-kzg functions, it should be able to transmute the pointer of the new Blob type to the pointer of the original Blob type.

Please correct me if I'm wrong.

**Comment by jtraglia** [comment]
> Can you give a reason why you think we should eliminate the `LIB_PREFIX` thing?

Simplicity.

* This change makes [kzg code in lighthouse](https://github.com/pawanjay176/lighthouse/commit/a2c780a2231d395c821fe7c88bc988bb4d806b46) much simpler.
* It would allow us to publish a crate which works for both presets.
* It will reduce the size of distributions, as only one library per platform is required, instead of two.

> For the stack overflow problem in Rust, I think we don't have to remove the Blob definition. It's better to keep it as is. If some bindings have the problem with it, they can just ignore it and redefine that type.

While this PR will result in less stack usage, the stack overflow problem with the rust bindings has been fixed:

* #354

Regarding the blob definition, I really wish we could keep this type but I don't think there's a strong reason it needs to exist. For example, nowhere in the C library (excluding the C tests) do we instantiate a blob. Only pointers are used.

We could create a blob type like: `typedef uint8_t Blob;` But I dislike this because it could be misused. For example, if someone were to instantiate a blob via `Blob blob;` it would be the incorrect size (just one byte). I'm definitely willing to compromise here though.

> For example, the Rust binding can redefine the Blob type as some kind of `Vec<u8>` and, when it wants to use the c-kzg functions, it should be able to transmute the pointer of the new Blob type to the pointer of the original Blob type.

Sure, I suppose that's possible but it sounds a bit more complicated.

File: src/c_kzg_4844.c [hunk 1]
```diff
@@ -53,8 +53,9 @@
 
 /** Internal representation of a polynomial. */
 typedef struct {
-    fr_t evals[FIELD_ELEMENTS_PER_BLOB];
-} Polynomial;
+    fr_t *evals;
+    uint64_t length;
```
mratsim, line 57: `size_t`?

Polynomial is internal so we don't need to adhere to a fixed size.

mratsim (commented):
Sounds good to me.

In terms of API I would rename KZGSettings as a context using `ctx` for it and make it the first parameter as it's pretty standard in cryptographic libraries but that's another PR or RFC.

src/c_kzg_4844.c
jtraglia: I also prefer `size_t`. Fixed.

jtraglia:
> In terms of API I would rename KZGSettings as a context using `ctx` for it and make it the first parameter as it's pretty standard in cryptographic libraries but that's another PR or RFC.

I've been wanting to rename `KZGSettings` to `Context` for a while. Good idea making it the first parameter. If the team is okay with it, I'll make another PR specifically for this later.

src/c_kzg_4844.h
```diff
@@ -159,6 +93,10 @@ typedef enum {
  * Stores the setup and parameters needed for computing KZG proofs.
  */
 typedef struct {
+    /** Number of field elements in a blob. */
+    uint64_t field_elements_per_blob;
+    /** Number of bytes in a blob. */
```
pawanjay176, line 98: Does it make sense to have an additional field for `bytes_per_blob`? we can just do `field_elements_per_blob * BYTES_PER_FIELD_ELEMENT` wherever required right? Or define a function that does that

jtraglia: I'm open to removing it. I thought it was a nice convenience field; it's used several times. I don't think a function would be appropriate.

nisdas:
This would be very useful for us in testing, a lot of our end to end tests rely on running minimal configs, so the ability to toggle the blob sizes would be very helpful for us and make maintaining the test infrastructure for blobs much easier.

jtraglia:
Changing the base branch to a new `runtime-fields` branch. I would like to:

* Merge this PR into that branch.
* Make additional PRs to that branch with binding updates.
* Finally, merge `runtime-fields` into `main`.

This allows us to see the "final product" without breaking `main` while the bindings are being updated.

src/c_kzg_4844.c
```diff
@@ -229,6 +224,31 @@ static C_KZG_RET new_fr_array(fr_t **x, size_t n) {
...
 
+/**
+ * Allocate memory for a polynomial.
+ *
+ * @remark Free the space later using free_poly().
+ *
+ * @param[out] out    Pointer to the poly_t struct
+ * @param[in]  length The length of the polynomial
+ */
+static C_KZG_RET new_poly(poly_t *out, uint64_t length) {
```
asn-d6, line 235: I know that I was pressuring you to make a new type called `poly_t` instead of having it raw. But now that I see how it looks like, I think your initial intuition was correct.

My suggestion would be to switch to `fr_t *polynomial` and `uint8_t *blob` as it was in ef592fcd5de64c3623cac7bef3cd7e7ebaf8c2ba.

One thing that would be nice is to always be consistent about the naming of the above. So for example not `fr_t *p`, but instead always `fr_t *polynomial`. We have less type-safety this way, so let's use the variable names to our advantage.

asn-d6 (changes requested):
This is a meta-comment but since this PR will get reviewed by a bunch of people, I think it would be a good idea to make it more pleasant to review.

In particular, ef592fcd5de64c3623cac7bef3cd7e7ebaf8c2ba is the core of the feature but it actually does three things at once:
- refactors FIELD_ELEMENTS_PER_BLOB
- refactors `Polynomial`
- refactors `Blob`

It would be great if the branch could be rebased such that these three operations are performed in separate commits. IMO it would make this PR easier to review.

src/c_kzg_4844.h
asn-d6: IMO it's OK to keep `bytes_per_blob` there. We currently have a `BYTES_PER_BLOB` macro, so it keeps the same logic as we have currently.

Not super strong opinion here. We could also remove it, if people insist on it.

jtraglia:
The rebase (splitting changes into smaller commits) is still a work in progress. Will finish tomorrow morning.

asn-d6:
Done an initial review and this looks really good! Thanks!

I pushed three commits. @jtraglia wanna review them and then squash them and rebase the PR so that it's clean again?
(You will have to squash `eed8f3506` and `05c8d7406` into `bdc1f0ef5`)

It would be great if @GottfriedHerold and @ppopth can also give it a review as well.

jtraglia:
Your changes look great, thanks for that 🥇

asn-d6:
> Your changes look great, thanks for that 🥇

Rebased and force-pushed as per our TG conversation.

I also pushed the pre-rebased branch at https://github.com/asn-d6/c-kzg-4844/tree/pr352_pre_rebase (the one with the three extra commits) in case you want to diff it with the force-pushed code to make sure I didn't screw up the rebase.

ppopth:
This may be a dump question, but why don't we put the size of polynomials and blobs into a field of a struct?
Something like this
```c
typedef struct {
    fr_t *evals;
    uint64_t size;
} Polynomial;
```
and
```c
typedef struct {
    uint8_t *bytes;
    uint64_t size;
} Blob;
```

I understand that all the polynomials and the blobs have the same size, but, by doing this, all the polynomials and the blobs will be self-contained and don't depend on external objects like KZGSettings.

Previously we didn't keep the size and we used `FIELD_ELEMENTS_PER_BLOB` instead which quite makes sense, because `FIELD_ELEMENTS_PER_BLOB` is just a global constant, not a variable. Now they depend on a variable not a constant, so I think it makes sense to rethink this.

src/c_kzg_4844.c
```diff
@@ -1367,9 +1397,9 @@ static C_KZG_RET verify_kzg_proof_batch(
  * @param[in]  n                 The number of blobs/commitments/proofs
  * @param[in]  s                 The trusted setup
  */
-C_KZG_RET VERIFY_BLOB_KZG_PROOF_BATCH(
+C_KZG_RET verify_blob_kzg_proof_batch(
     bool *ok,
-    const Blob *blobs,
+    const uint8_t *blobs,
```
ppopth, line 1402: Why not `**blobs`? I think `**blobs` implies an array of blobs.

src/c_kzg_4844.c
ppopth: > I know that I was pressuring you to make a new type called `poly_t` instead of having it raw. But now that I see how it looks like, I think your initial intuition was correct.

Did you do `typedef fr_t *poly_t`?

I can't see the commit with `poly_t` anymore.

ppopth: For documentation purpose, the commit is https://github.com/ethereum/c-kzg-4844/commit/ff405cafe49a9bd5e108bf1a82ebe9c802753084

src/c_kzg_4844.c
jtraglia: That would be an array of pointers. We expect a flattened, contiguous array of the blobs' bytes.

jtraglia:
> This may be a dumb question, but why don't we put the size of polynomials and blobs into a field of a struct?

So we did try this with polynomials (see this [commit](https://github.com/ethereum/c-kzg-4844/commit/ff405cafe49a9bd5e108bf1a82ebe9c802753084)), but decided against it for a few reasons. I prefer the way it is now, but I'm open to using `poly_t` and `blob_t` types. Not a strong opinion. I think this could be done in a separate PR.

* Didn't feel that it really simplified that much.
* Required new initialization/free functions which didn't quite match others in usage.
  * For example, in `c_kzg_free` we just pass the value, but `free_poly` takes a pointer.
* Realized that it's easier to make a mistake with the structure.
  * It must be initialized with `{NULL, 0}` which is fine, but a little ugly.
  * If you forget to do this, you could free an invalid address (value from the stack).

asn-d6:
> > This may be a dumb question, but why don't we put the size of polynomials and blobs into a field of a struct?
> 
> So we did try this with polynomials (see this [commit](https://github.com/ethereum/c-kzg-4844/commit/ff405cafe49a9bd5e108bf1a82ebe9c802753084)), but decided against it for a few reasons. I prefer the way it is now, but I'm open to using `poly_t` and `blob_t` types. Not a strong opinion. I think this could be done in a separate PR.
> 
> * Didn't feel that it really simplified that much.
> * Required new initialization/free functions which didn't quite match others in usage.
>   
>   * For example, in `c_kzg_free` we just pass the value, but `free_poly` takes a pointer.
> * Realized that it's easier to make a mistake with the structure.
>   
>   * It must be initialized with `{NULL, 0}` which is fine, but a little ugly.
>   * If you forget to do this, you could free an invalid address (value from the stack).

Yep agreed. I was part of this entire refactoring-then-rerefactoring operation, and I agree that the status quo (without `poly_t`) seems simpler to read.

src/c_kzg_4844.c
GottfriedHerold: I would prefer to explicitly clarify that in the doc. Currently, it says that "all input arrays contain n elements", which does not match the "flattened array of bytes" view.

jtraglia: Hmm okay. Thoughts @asn-d6? I'd like to point out that `commitments_bytes` and `proofs_bytes` behave the same way. We're passing all of the bytes, not pointers to each commitment/proof. And under the hood, the blobs argument is passed to this function the same as it was before.

jtraglia: I'm not opposed to changing this btw! It would remove the need to flatten the blobs in the bindings. This would be faster and less complicated. But it would require more changes to the bindings.

src/c_kzg_4844.c
```diff
@@ -1766,39 +1799,76 @@ C_KZG_RET LOAD_TRUSTED_SETUP(
...
     /* Read the number of g1 points */
-    num_matches = fscanf(in, "%" SCNu64, &i);
-    CHECK(num_matches == 1);
-    CHECK(i == TRUSTED_SETUP_NUM_G1_POINTS);
+    uint64_t num_g1_points;
+    num_matches = fscanf(in, "%" SCNu64, &num_g1_points);
+    if (num_matches != 1) {
+        ret = C_KZG_BADARGS;
+        goto out;
+    }
```
GottfriedHerold, line 1815: The code currently checks that n1 matches either the mainnet or minimal setting.
While that's an OK thing to do, the code actually should work in a more general setting and the current behaviour is surprising; my expectation from reading the functions' doc would be that it worked with any value of n1, n2.

As it stands, the fact that it (currently?) only works for n1, n2 matching these settings needs to be added to the documentation.

src/c_kzg_4844.c
```diff
@@ -1766,39 +1799,76 @@ C_KZG_RET LOAD_TRUSTED_SETUP(
  * @param[out] out Pointer to the loaded trusted setup data
  * @param[in]  in  File handle for input
  */
```
GottfriedHerold, line 1801: I would suggest adding a comment with regards to seeking operations.
Notably: Either clarify that the function expects that the file's read pointer is at the start  or actually perform fseek. 
Similar, the post-state with regards to position in the filed stream should be documented.

(Notably, calling this function twice on the same openend file will fail, which might be unexpected)

src/c_kzg_4844.c
jtraglia: Good point. Would you prefer we update it to be more generic or add a note to the doc?

src/c_kzg_4844.c
jtraglia: Another good point. Alternatively, we could fseek to the beginning of the file before & after reading it.

GottfriedHerold (approved):
Generally, the changes look good to me.

src/c_kzg_4844.c
GottfriedHerold: I would prefer adding a note.

src/c_kzg_4844.c
GottfriedHerold: I would prefer to keep the logic as-is, just document it. The reason being that some FILE objects do not support arbitrary fseeks and this adds more potential failure cases. The caller is better equipped to deal with such cases.

jtraglia:
> Generally, the changes look good to me.

Thanks @GottfriedHerold. How do these ([2328ac5](https://github.com/ethereum/c-kzg-4844/pull/352/commits/2328ac56fa16437fd96948ec36e290440db523b9)) changes look? I also moved the `n1` and `n2` checks from `load_trusted_setup_file` to `load_trusted_setup` so they would both behave the same.

src/c_kzg_4844.c
asn-d6: Agreed that this should be a doc fix, instead of a "make it more generic" fix.

src/c_kzg_4844.c
asn-d6: I think this is a good point raised by @ppopth and @GottfriedHerold . IMO we should clarify the documentation of the function and not change its internals.

jtraglia: Okay cool. Fixed in [e7d980e](https://github.com/ethereum/c-kzg-4844/pull/352/commits/e7d980e9d4bd15595706d6e1cc45445bbb7f6f23). Just a remark, but I think that's enough.

asn-d6 (approved):
LGTM! Thanks!

ppopth:
> * This change makes [kzg code in lighthouse](https://github.com/pawanjay176/lighthouse/commit/a2c780a2231d395c821fe7c88bc988bb4d806b46) much simpler.
> 
> * It would allow us to publish a crate which works for both presets.

I looked at https://github.com/pawanjay176/lighthouse/commit/a2c780a2231d395c821fe7c88bc988bb4d806b46 already. It seems to me that the commit makes things simpler because it changes from
```rust
pub struct Kzg<P: KzgPreset> {
    trusted_setup: P::KzgSettings,
}
```
to
```rust
pub struct Kzg {
    trusted_setup: KzgSettings,
}
```
I think the latter can be done without this PR. As I mentioned in https://github.com/ethereum/c-kzg-4844/pull/352#issuecomment-1713376826, both mainnet and minimal presets can be combined into one crate which makes the latter possible.

> I think, this one is unnecessary because you can just use one crate for both mainnet and minimal, instead of using two. You can probably put the mainnet in one module and the minimal in another module.

The last one

> * It will reduce the size of distributions, as only one library per platform is required, instead of two.

I think this is not an issue. The idea of duplicating codes is very common. We do it all the time when we use generics (like in Rust and C++).

PR #353, opened 2023-08-31
Return uint_8 instead of bool from all ckzg functions

pawanjay176:
I think we might have finally found the root cause of all our windows rust issues.
Thanks to @michaelsproul for helping find this 🙏 

blst.h has this redefinition of bool 
https://github.com/supranational/blst/blob/78fee18b25e16975e27b2d0314f6a323a23e6e83/bindings/blst.h#L27-L31

Because we import `blst.h` In `c-kzg-4844.h`, `bool` in the c source code is redefined as an `int` on windows. So in the c interface for all `verify_*` functions, `bool *ok` is in fact `int *ok`.

When generating the rust bindings, bindgen made the stupid mistake of assuming that a bool in C would be a bool in rust. So on calling the underlying C function, rust allocates 1 byte for the bool type where C required 4. This is probably what was causing the undefined behaviour all this time.

Making this PR just to run CI on the issue. There's probably a better way of fixing this in C.

Running this with lighthouse on devnet 8. No crashes so far 🎉

### DIFF
--- a/bindings/rust/Cargo.toml
+++ b/bindings/rust/Cargo.toml
@@ -43,8 +43,3 @@ glob = "0.3"
 [[bench]]
 name = "kzg_benches"
 harness = false
-
-# The benchmarks crash on Windows with Rust 1.70. This is a band-aid fix for
-# that. Refer to #318 for more details. This should be removed if fixed.
-[profile.bench]
-opt-level = 0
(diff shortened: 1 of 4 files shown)


jtraglia:
This is no longer needed. The root issue was identified. Fixed here:

* https://github.com/ethereum/c-kzg-4844/pull/354

PR #354, opened 2023-08-31
Set /std:c11 ccflag for Windows in Rust bindings

### LINKED ISSUE #318: Broken CI in rust-windows because of rust-1.7.0

### LINKED ISSUE #331: Stack overflow in Rust benchmarks on Windows
https://github.com/ethereum/c-kzg-4844/actions/runs/5813951682/job/15762618420?pr=330

It's a bit jumbled, but the benchmarks on Windows are overflowing the stack somehow:
```
Benchmarking verify_blob_kzg_proof
Benchmarking verify_blob_kzg_proof: Warming up for 3.0000 s
Benchmarking verify_blob_kzg_proof: Collecting 100 samples in estimated 5.0302 s (1300 iterations)
Benchmarking verify_blob_kzg_proof: Analyzing
Benchmarking verify_blob_kzg_proof_batch/1
Benchmarking verify_blob_kzg_proof_batch/1: Warming up for 3.0000 s
verify_blob_kzg_proof   time:   [3.8203 ms 3.8244 ms 3.8285 ms]
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild


thread 'main' has overflowed its stack
error: bench failed, to rerun pass `--bench kzg_benches`

Caused by:
  process didn't exit successfully: `D:\a\c-kzg-4844\c-kzg-4844\bindings\rust\target\release\deps\kzg_benches-887efed028bd7e9a.exe --bench` (exit code: 0xc00000fd, STATUS_STACK_OVERFLOW)
```

jtraglia:
This PR fixes our Windows issues. This is an alternative to #353.

See #353 for details about the problem. To be brief, if `__STDC_VERSION__` isn't defined as c99 or greater, blst will typedef a bool as an int (this check is [here](https://github.com/supranational/blst/blob/78fee18b25e16975e27b2d0314f6a323a23e6e83/bindings/blst.h#L28-L31) btw). There is a bug in bindgen associated with this. MSVC doesn't define `__STDC_VERSION__` by default, but we can use the `/std` flag to set it to `c11`, which is the earliest version it supports.

Fixes #318.
Fixes #331.

Big thanks to @pawanjay176 & @michaelsproul for identifying this.

### DIFF
--- a/bindings/rust/Cargo.toml
+++ b/bindings/rust/Cargo.toml
@@ -43,8 +43,3 @@ glob = "0.3"
 [[bench]]
 name = "kzg_benches"
 harness = false
-
-# The benchmarks crash on Windows with Rust 1.70. This is a band-aid fix for
-# that. Refer to #318 for more details. This should be removed if fixed.
-[profile.bench]
-opt-level = 0
--- a/bindings/rust/build.rs
+++ b/bindings/rust/build.rs
@@ -30,7 +30,15 @@ fn main() {
     let mut cc = cc::Build::new();
 
     #[cfg(windows)]
-    cc.flag("-D_CRT_SECURE_NO_WARNINGS");
+    {
+        cc.flag("-D_CRT_SECURE_NO_WARNINGS");
+
+        // In blst, if __STDC_VERSION__ isn't defined as c99 or greater, it will typedef a bool to
+        // an int. There is a bug in bindgen associated with this. It assumes that a bool in C is
+        // the same size as a bool in Rust. This is the root cause of the issues on Windows. If/when
+        // this is fixed in bindgen, it should be safe to remove this compiler flag.
+        cc.flag("/std:c11");
+    }
 
     cc.include(blst_headers_dir.clone());
     cc.warnings(false);


asn-d6:
Thanks!

kaliubuntu0206:
@jtraglia @asn-d6 I think this is not a correct fix and causing issues when the toolchain is set to GNU

https://stackoverflow.com/questions/58226545/how-to-switch-between-rust-toolchains

GCC compiler is complaining that `/std:c11` is a wrong flag, so I think the flag should be only added when the toolchain is MSVC

https://rust-lang.github.io/rustup/installation/windows.html

kaliubuntu0206:
Tried to fix the issue by switching the toolchain to `x86_64-pc-windows-gnu` which seems to be a correct one and the one supported by jemalloc https://github.com/gnzlbg/jemallocator#platform-support

1. c-kzg bindings from EF complained about the wrong flag specified https://github.com/ethereum/c-kzg-4844/pull/354#issuecomment-1856355981

2. jemallocator bindings also complained that `sh` is missing. Since windows doesn't use shell script I think it should be fixed from the bindings as well.

jtraglia:
Hey @kaliubuntu0206 👋 I'm okay with only setting it for MSVC, sounds like a mistake on my part. Would you like to make a PR for this? Also worth noting, blst appears to have fixed this problem (https://github.com/supranational/blst/commit/9c87d4a09d6648e933c818118a4418349804ce7f). I don't think we should update blst right now though, since we would prefer to use a stable release.

PR #356, opened 2023-09-01
Change the rust interface to take references

pawanjay176:
All kzg functions in the library now are passed by reference instead of by value


jtraglia (approved):
This looks great. I've been wanting to do this for a while. Thanks!

PR #358, opened 2023-09-11
feat(rust): add KzgSettings::clone

mattsse:
Adds Clone impl for KzgSettings

since c_kzg_4844.c does alloc+free, Clone must alloc accordingly, free occurs on drop.

This is only possible because the number of G1 and G2 are constant and enforced by KzgSettings API.

### DIFF
--- a/bindings/rust/src/bindings/mod.rs
+++ b/bindings/rust/src/bindings/mod.rs
@@ -241,6 +241,42 @@ impl Drop for KZGSettings {
     }
 }
 
+impl Clone for KZGSettings {
+    fn clone(&self) -> Self {
+        use alloc::alloc::{alloc, Layout};
+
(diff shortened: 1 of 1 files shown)


mattsse:
> Would this feature be possible without unsafe? And if yes, what would be the drawbacks of that approach?

it is not, because copying/dereferencing a pointer always requires unsafe

> Also, what do you currently do in your codebase, now that Clone is not available?

Currently, we're required to use `Arc<KzgSettings>`, because we need it in a few places.
though even with a clone impl, I'd continue using an Arc where possible, because Clone requires a few allocations.

but having the option to clone it would make it easier to use under some circumstances where an API does not support `Arc<KzgSettings>`

mattsse:
I thought about this a bit, and this feature is actually not sound, but KzgSettings can still be created with arbitrary length via unsafe external calls, so there's no real guarantee the num of points are constant

PR #359, opened 2023-09-12
Add script for converting trusted setups

jtraglia:
As discussed on our call, this script will convert the trusted setup from the spec to the format that c-kzg uses.


asn-d6:
I would wait until the official trusted setup comes out before we merge this? To make sure that it corresponds to the final version of the json format.

jtraglia:
That works. Right now it conforms to the current format. It will need some minor changes for the new format.

jtraglia:
Depends on #377 to get the new trusted setup.

asn-d6 (approved):
Looks great!

PR #360, opened 2023-09-14
chore: make num g2 points pub

mattsse:
the constant is not generated as constants in the bindings, pub is useful to have access to


rakita (approved):
Small ergonomic change

asn-d6 (approved):

PR #361, opened 2023-09-15
Added AsRef<[u8]> method for Blob

ethDreamer:
This is needed so that we can always store blobs in `c_kzg::Blob` form which allows us to eliminate a lot of copying of the blob bytes between formats.


jtraglia (approved):
LGTM 👍

PR #362, opened 2023-09-15
Update go bindings to use runtime field count

jtraglia:
This PR updates the Go bindings to work with runtime field counts.

### DIFF
--- a/bindings/go/main.go
+++ b/bindings/go/main.go
@@ -2,9 +2,6 @@ package ckzg4844
 
 // #cgo CFLAGS: -I${SRCDIR}/../../src
 // #cgo CFLAGS: -I${SRCDIR}/blst_headers
-// #ifndef FIELD_ELEMENTS_PER_BLOB
-// #define FIELD_ELEMENTS_PER_BLOB 4096
-// #endif
 // #include "c_kzg_4844.c"
 import "C"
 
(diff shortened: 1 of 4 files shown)


bindings/go/main.go
```diff
@@ -20,19 +17,17 @@ import (
...
-	FieldElementsPerBlob = C.FIELD_ELEMENTS_PER_BLOB
 )
 
 type (
 	Bytes32       [32]byte
 	Bytes48       [48]byte
 	KZGCommitment Bytes48
 	KZGProof      Bytes48
-	Blob          [BytesPerBlob]byte
+	Blob          []byte
```
asn-d6, line 30: Is this type useful now that the length is unspecified? Or should we ditch the `Blob` type and have it be an array of bytes in all the functions like we did it in C? This way the `len(blob)` pattern everywhere is a bit more clear.

jtraglia: Hmm. I would prefer to keep this type for readability. There's not a strong reason to keep it though.

asn-d6: Sounds good. I guess it fits more how all the other bindings are structured as well.

bindings/go/main.go
```diff
@@ -109,13 +107,31 @@ func (b *Blob) UnmarshalText(input []byte) error {
 	if err != nil {
 		return err
 	}
-	if len(bytes) != len(b) {
+	if len(bytes) != GetBytesPerBlob() {
 		return ErrBadArgs
 	}
-	copy(b[:], bytes)
+	*b = bytes
```
asn-d6, line 113: The copy pattern does not work anymore? Would `copy(*b, bytes)` work? Is it the same as what you did?

asn-d6: Here is how a similar function in geth is handled: https://github.com/ethereum/go-ethereum/blob/master/common/hexutil/json.go#L56

Perhaps we can replicate the geth logic and trust in their authority?

jtraglia: That's pretty similar to how we do it now, but less straight-forward IMO. They use `hex.Decode` instead of `hex.DecodeString` and they have a `checkText` function. I would prefer to keep what we have.

asn-d6: Thanks. Current approach looks fine!

asn-d6 (approved):

PR #364, opened 2023-09-15
Update java bindings to use runtime field count

jtraglia:
This PR updates the Java bindings to work with runtime field counts.

### DIFF
--- a/.github/workflows/java-tests.yml
+++ b/.github/workflows/java-tests.yml
@@ -5,8 +5,6 @@ on:
     branches:
       - main
   pull_request:
-    branches:
-      - main
 
 jobs:
   tests:
--- a/bindings/java/Makefile
+++ b/bindings/java/Makefile
@@ -42,15 +42,7 @@ else
   GRADLE_COMMAND=./gradlew
 endif
 
-PRESET ?= mainnet
-
-ifeq ($(PRESET),minimal)
-  FIELD_ELEMENTS_PER_BLOB ?= 4
-else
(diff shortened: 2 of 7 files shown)


jtraglia:
@StefanBratanov do we need to expose a `getBytesPerBlob()` method?

StefanBratanov:
> @StefanBratanov do we need to expose a `getBytesPerBlob()` method?

We already do right in CKZG4844JNI. Or you mean have it as a native method?

StefanBratanov:
Can change actually the javadoc of `getFieldElementsPerBlob()` as well since it is not correct.

jtraglia:
> > @StefanBratanov do we need to expose a `getBytesPerBlob()` method?
> 
> We already do it in CKZG4844JNI right? Or you mean have it as a native method?

Ah that's right. I forgot it was [in CKZG4844JNI](https://github.com/ethereum/c-kzg-4844/blob/fbef59a3f9e8fa998bdb5069d212daf83d586aa5/bindings/java/src/main/java/ethereum/ckzg4844/CKZG4844JNI.java#L83-L90). It doesn't need to be a native method.

StefanBratanov:
> > Can change actually the javadoc of `getFieldElementsPerBlob()` as well since it is not correct.
> 
> Fixed in [072cb97](https://github.com/ethereum/c-kzg-4844/pull/364/commits/072cb9707962efe1484cadf834c4d3312bc79db9). Happy to make changes if it doesn't read well.

Looks good. Thanks. Happy to merge.

asn-d6 (approved):
LGTM! Thanks!

PR #365, opened 2023-09-15
Update python bindings to use runtime field count

jtraglia:
This PR updates the Python bindings to work with runtime field counts.


asn-d6 (approved):
LGTM!

PR #366, opened 2023-09-15
Update nodejs bindings to use runtime field count

jtraglia:
This PR updates the Node.js bindings to work with runtime field counts.

Note: the JSON trusted setup was outdated. It contained 4096 G2 points. Pulled in the latest version.


asn-d6:
@matthewkeil would be great if you could take a look here. Thanks!

bindings/node.js/lib/kzg.js
```diff
@@ -43,6 +45,7 @@ bindings.loadTrustedSetup = function loadTrustedSetup(filePath) {
   if (path.parse(filePath).ext === ".json") {
     filePath = transformTrustedSetupJson(filePath);
   }
+  console.log(filePath);
   originalLoadTrustedSetup(filePath);
```
asn-d6, line 49:
```suggestion
  originalLoadTrustedSetup(filePath);
```

bindings/node.js/src/kzg.cxx
```diff
@@ -490,11 +513,15 @@ Napi::Value VerifyBlobKzgProofBatch(const Napi::CallbackInfo &info) {
...
         // after each iteration since data is being memcpy
         Napi::HandleScope scope{env};
-        Blob *blob = get_blob(env, blobs_param[index]);
+        uint8_t *blob = get_blob(env, kzg_settings, blobs_param[index]);
         if (blob == nullptr) {
             goto out;
         }
-        memcpy(&blobs[index], blob, BYTES_PER_BLOB);
+        memcpy(
+            &blobs[index * kzg_settings->bytes_per_blob],
```
asn-d6, line 521: Hmm, I see we don't define a new custom `Blob` type as we do in the other bindings and here the memcpy index logic has to change. Is this intentional?

bindings/node.js/src/kzg.cxx
```diff
@@ -161,6 +162,28 @@ inline Bytes48 *get_bytes48(
...
+}
+
+Napi::Value GetBytesPerBlob(const Napi::CallbackInfo &info) {
+    Napi::Env env = info.Env();
+    KZGSettings *kzg_settings = get_kzg_settings(env, info);
+    if (kzg_settings == nullptr) {
+        return env.Null();
+    }
+    return Napi::Number::New(
+        env, static_cast<double>(kzg_settings->bytes_per_blob)
```
matthewkeil, line 183: I didn't notice where `bytes_per_blob` and `field_elements_per_blob` is getting added to the `KZGSettings`.  Are you sure this doesn't throw/segfault?

bindings/node.js/lib/kzg.d.ts
```diff
@@ -6,7 +6,6 @@ export type Bytes32 = Uint8Array; // 32 bytes
 export type Bytes48 = Uint8Array; // 48 bytes
 export type KZGProof = Uint8Array; // 48 bytes
 export type KZGCommitment = Uint8Array; // 48 bytes
-export type Blob = Uint8Array; // 4096 * 32 bytes
```
matthewkeil, line 9: Is the idea that `blob` params everywhere will be untrusted length so reverting to the base `Uint8Array` for param type?

bindings/node.js/lib/kzg.d.ts
```diff
@@ -88,28 +88,28 @@ export function verifyKzgProof(commitment: Bytes48, zBytes: Bytes32, yBytes: Byt
...
  *
  * @return {boolean} - true/false depending on proof validity
  *
  * @throws {TypeError} - For invalid arguments or failure of the native library
  */
-export function verifyBlobKzgProof(blob: Blob, commitment: Bytes48, proof: Bytes48): boolean;
+export function verifyBlobKzgProof(blob: Uint8Array, commitment: Bytes48, proof: Bytes48): boolean;
 
 /**
- * Given an array of blobs and their proofs, verify that they corresponds to their
```
matthewkeil, line 102: Good catch on this!!! Your eye for detail is amazing sir Traglia!!

bindings/node.js/lib/kzg.js
jtraglia: Oops. Thanks 🙏

bindings/node.js/lib/kzg.d.ts
jtraglia: Hmm thanks for pointing this out. I think I want to keep this type now.

bindings/node.js/src/kzg.cxx
jtraglia: To my knowledge, like in C, it's not very practical in C++ to have a custom `Blob` type. And yes, the index logic needs to change since it's changed from an array of blobs (known size) to an array of bytes. The two should be equivalent. Given some bytes, to get to the start of the blob N, you must jump `N * bytes_per_blob` into the array.

bindings/node.js/lib/kzg.js
```diff
@@ -43,6 +45,7 @@ bindings.loadTrustedSetup = function loadTrustedSetup(filePath) {
   if (path.parse(filePath).ext === ".json") {
     filePath = transformTrustedSetupJson(filePath);
   }
+  console.log(filePath);
```
matthewkeil: This will show in the client during startup.  It may be better to denote what it represents if you actually want this logged so it's clear what it is in the system logs.
```suggestion
  console.log(`c-kzg trusted setup filepath: ${filePath}`);
```

jtraglia: Yeah, removed that log. I was debugging a problem when loading the trusted setup.

bindings/node.js/lib/kzg.js
```diff
@@ -21,8 +21,10 @@ const bindings = require("bindings")("kzg");
 function transformTrustedSetupJson(filePath) {
   const trustedSetup = JSON.parse(fs.readFileSync(filePath, "utf8"));
   const setupText =
-    bindings.FIELD_ELEMENTS_PER_BLOB +
-    "\n65\n" +
+    trustedSetup.setup_G1_lagrange.length.toString() +
```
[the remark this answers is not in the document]
jtraglia: Fixed 👍

bindings/node.js/lib/kzg.d.ts
jtraglia: Yeah, added this back. This is now the same as the other bindings.

bindings/node.js/src/kzg.cxx
jtraglia: So I added back the `Blob` type in `lib/kzg.d.ts`. This is akin to how other bindings do it.

bindings/node.js/lib/kzg.d.ts
```diff
@@ -15,19 +15,33 @@ export interface TrustedSetupJson {
...
 export const BYTES_PER_FIELD_ELEMENT: number;
 export const BYTES_PER_PROOF: number;
-export const FIELD_ELEMENTS_PER_BLOB: number;
+
+/**
+ * Get the number of field elements in a blob.
+ *
+ * @return {number} - The number of fields
+ *
+ * @throws {TypeError} - The trusted setup isn't loaded
```
matthewkeil, line 27:
```suggestion
 * @throws {Error} - The trusted setup isn't loaded
```

bindings/node.js/lib/kzg.js
matthewkeil: 🥷

bindings/node.js/lib/kzg.d.ts
```diff
@@ -15,19 +15,33 @@ export interface TrustedSetupJson {
...
+ */
+export function getFieldElementsPerBlob(): number;
+
+/**
+ * Get the number of bytes in a blob.
+ *
+ * @return {number} - The number of bytes
+ *
+ * @throws {TypeError} - The trusted setup isn't loaded
```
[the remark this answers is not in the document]
jtraglia: Oops. Thanks!

matthewkeil (commented):
You already got most of the items @jtraglia.  Just the Error vs TypeError and the question about removing the `Blob` type.  I tend to think that was a cleaner way of handling the array of bytes so its fixed length.  What do you think?  What about you @asn-d6?

bindings/node.js/src/kzg.cxx
jtraglia: The Blob struct will no longer exist. It's been removed in that PR.

matthewkeil (approved):
LGTM!! 🚀 

Would love to actually run the tests to make sure all is well once https://github.com/ethereum/c-kzg-4844/pull/352 is approved and merged.

jtraglia:
Great. Yes, we'll be sure that all of the tests pass before merging!

asn-d6:
I'm slightly confused on why the bindings here use `Blob` in the typescript code, but `uint8_t *blob` in the C++ code.
From reading the comments above I see this is because "To my knowledge, like in C, it's not very practical in C++ to have a custom Blob type.".

Is this mixed usage of `Blob` preferable?

bindings/node.js/src/kzg.cxx
```diff
@@ -142,9 +142,10 @@ inline uint8_t *get_bytes(
     }
     return array.Data();
 }
-inline Blob *get_blob(const Napi::Env &env, const Napi::Value &val) {
-    return reinterpret_cast<Blob *>(get_bytes(env, val, BYTES_PER_BLOB, "blob")
-    );
+inline uint8_t *get_blob(
+    const Napi::Env &env, const KZGSettings *s, const Napi::Value &val
```
asn-d6, line 146: This is an ultra nit but I think it's preferable to have `KZGSettings` be the last argument here so that `val` is the second argument and matches all the other `get_bytes32()` and `get_bytes48()` function signatures.

jtraglia: I'm okay with ultra nits! Thanks for pointing this out. I appreciate consistency. Fixed in [bdcc3b9](https://github.com/ethereum/c-kzg-4844/pull/366/commits/bdcc3b9851578a4c3b785906b73a14c3cc7ba8dd).

jtraglia:
> I'm slightly confused on why the bindings here use `Blob` in the typescript code, but `uint8_t *blob` in the C++ code.

For the API, I think `Blob` is a more human friendly name. Keep in mind `Blob` is a `Uint8Array` under the hood:

https://github.com/ethereum/c-kzg-4844/blob/bdcc3b9851578a4c3b785906b73a14c3cc7ba8dd/bindings/node.js/lib/kzg.d.ts#L9

> Is this mixed usage of `Blob` preferable to just using `uint8_t` everywhere?

For me, yeah. As long as it's `Blob` everywhere in the TS code & `uint8_t *` in the C++ code I'm good.

I suppose I don't know what the alternative is in the C++ code. Would you prefer we `typedef Blob uint8_t`? That would work, but I chose not to do that in the C code. We could also use `std::vector<uint8_t>` if you think that's a good idea.

asn-d6 (approved):
LGTM! Thanks all!

PR #367, opened 2023-09-20
Update C# bindings to use runtime field count

jtraglia:
This PR updates the C# bindings to work with runtime field counts. Needed to:

* Define a `KzgSettings` structure that C# can see.
  * This only needs the first two fields, as the others aren't used.
* Mark some functions/classes as `unsafe` so they can use pointers.
* Change `int` variables to `ulong` so they play well together.
  * Probably should have been this way before.
* Change `IntPtr` to `KzgSettings*` everywhere.
* Remove `FIELD_ELEMENTS_PER_BLOB` from Makefile.

### DIFF
--- a/.github/workflows/csharp-tests.yml
+++ b/.github/workflows/csharp-tests.yml
@@ -8,8 +8,7 @@ on:
     branches:
       - main
   pull_request:
-    branches:
-      - main
+
 env:
   binding_build_number_based_version: 0.3.1.${{ github.run_number }}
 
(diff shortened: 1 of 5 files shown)


asn-d6:
Hey @flcl42 ! Would be great if you could do a quick review of this. Thanks!

flcl42 (approved):

asn-d6 (approved):
Thanks! LGTM!

PR #368, opened 2023-09-21
Update Nim bindings to use runtime field count

jangko:
Why Nim CI is not running?


jtraglia:
@jangko thanks for making this PR! 🥇 

I've fixed the CI tests. They were only running if the "into" branch was `main`.

jtraglia (approved):
LGTM 👍

PR #370, opened 2023-09-24
[Java binding] Run both mainnet and minimal tests

StefanBratanov:
Before running the tests in the java binding was dependent on the `PRESET` env value. Given it is no longer used since `fieldElementsPerBlob` is runtime value can run both mainnet and minimal tests in once.

- Removed finals to make the test file consistent


jtraglia (approved):
Nice. Good idea! LGTM 👍

PR #372, opened 2023-10-03
Improve ckzg native library search path + small improvements

flcl42:
1. `arm64-apple-macos11` is more correct macos target;
1. `dylib` is better extension on Macos for "dynamically linked shared library";
1. We do not care about old arch "x86_64-v2-win" because we will have that runtime check;
1. Library search path should be based on Ckzg.Bindings.dll location, not on the current directory. So `AppContext.BaseDirectory` is used;
1. The library may be built with any SDK.

The main reason for the change is `4.` when someone who uses Ckzg.Bindings as a dependency directly runs their project from source code, the library fails to find the native dependency.

### DIFF
--- a/bindings/csharp/README.md
+++ b/bindings/csharp/README.md
@@ -5,7 +5,7 @@ This directory contains C# bindings for the C-KZG-4844 library.
 ## Prerequisites
 
 Build requires:
-- `clang` as a prefered buid tool for the native wrapper of ckzg. On Windows it's tested with clang from [Microsoft Visual Studio components](https://learn.microsoft.com/en-us/cpp/build/clang-support-msbuild?view=msvc-170);
+- `clang` as a preferred build tool for the native wrapper of ckzg. On Windows, it's tested with clang from [Microsoft Visual Studio components](https://learn.microsoft.com/en-us/cpp/build/clang-support-msbuild?view=msvc-170);
 - [.NET SDK](https://dotnet.microsoft.com/en-us/download) to build the bindings.
 
 ## Build & test
--- a/bindings/csharp/README.md
+++ b/bindings/csharp/README.md
@@ -4,8 +4,9 @@ This directory contains C# bindings for the C-KZG-4844 library.
 
 ## Prerequisites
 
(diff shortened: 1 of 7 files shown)


bindings/csharp/README.md
```diff
@@ -4,8 +4,9 @@ This directory contains C# bindings for the C-KZG-4844 library.
 
 ## Prerequisites
 
-These bindings require .NET 6.0 (not 7.0). This can be found here:
-* https://dotnet.microsoft.com/en-us/download/dotnet/6.0
+Build requires:
+- `clang` as a prefered buid tool for the native wrapper of ckzg. On Windows it's tested with clang from [Microsoft Visual Studio components](https://learn.microsoft.com/en-us/cpp/build/clang-support-msbuild?view=msvc-170);
```
jtraglia, line 8:
```suggestion
- `clang` as a preferred build tool for the native wrapper of ckzg. On Windows, it's tested with clang from [Microsoft Visual Studio components](https://learn.microsoft.com/en-us/cpp/build/clang-support-msbuild?view=msvc-170);
```

jtraglia (changes requested):
Looks good, just one suggestion. Also, worth mentioning that, as of yesterday, there are arm64 builders for macOS available on GitHub workflows. Could simplify the build process a little.

* https://github.blog/changelog/2023-10-02-github-actions-apple-silicon-m1-macos-runners-are-now-available-in-public-beta/

flcl42:
> Looks good, just one suggestion. Also, worth mentioning that, as of yesterday, there are arm64 builders for macOS available on GitHub workflows. Could simplify the build process a little.
> 
>     * https://github.blog/changelog/2023-10-02-github-actions-apple-silicon-m1-macos-runners-are-now-available-in-public-beta/

It's available for paid plans only looks like.

jtraglia:
> It's available for paid plans only looks like.

Ah sad. I didn't catch that detail in the announcement.

jtraglia (approved):
LGTM. Thanks!

PR #373, opened 2023-10-03
Update fuzzers to work with variable sized blobs

jtraglia:
The `Blob` and `BYTES_PER_BLOB` no longer exist. To fix this, I converted the offset & input_size macros to variables.


asn-d6 (approved):
LGTM!

PR #374, opened 2023-10-03
Update comment to say 7 is the primitive root

jtraglia:
Super minor nit, this comment implied that we use 5 as the primitive root but it's actually 7.

### DIFF
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -110,7 +110,7 @@ static const g1_t G1_IDENTITY = {
  *
  * In the formula above, the restriction can be slightly relaxed to `r` being a non-square.
  * This is easy to check: We just require that r^((q-1)/2) == -1. Instead of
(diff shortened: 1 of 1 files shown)


src/c_kzg_4844.c
```diff
@@ -110,7 +110,7 @@ static const g1_t G1_IDENTITY = {
  *
  * In the formula above, the restriction can be slightly relaxed to `r` being a non-square.
  * This is easy to check: We just require that r^((q-1)/2) == -1. Instead of
- * 5, we could use 7, 10, 13, 14, 15, 20... to create the 2^i'th roots of unity below.
+ * 7, we could use 5, 10, 13, 14, 15, 20... to create the 2^i'th roots of unity below.
```
kevaundray, line 113: I think we want to remove 5, since 5 is not a primitive root of unity in the field

asn-d6:
Thanks!

PR #376, opened 2023-10-09
Rust runtime bindings take 2

pawanjay176:
This is a version of what Sean described in https://github.com/ethereum/c-kzg-4844/pull/363#issuecomment-1749509149

The changes are the following:
1. Make all kzg interface functions methods on the `KzgSettings` struct
2. In the public interface, the `Blob` argument is just a byte slice
3. Make `Blob` a private type which contains a byte slice. The `Blob` constructor does the length check and returns an error if it doesn't equal `self.bytes_per_blob()`
4. Have internal functions which accept a length validated`Blob` call the underlying FFI functions.

Some advantages to this approach:
1. No generics
2. `Blob` being just a byte slice in the public interface means that users of the library can store the Blobs in whatever container they want and don't have to do conversions before passing it to the library
3. All kzg functions being methods on `KzgSettings` is advantageous because we do the blob length validations w.r.t a given trusted setup
4. A `Blob` passed to the internal functions would be valid w.r.t the passed `KzgSettings` so there's less scope of shooting ourselves in the foot for any future changes.

Curious to see what you guys think @Rjected @asn-d6 @jtraglia 
cc @realbigsean

### DIFF
--- a/.github/workflows/rust-tests.yml
+++ b/.github/workflows/rust-tests.yml
@@ -4,8 +4,6 @@ on:
     branches:
       - main
   pull_request:
-    branches:
-      - main
 
 jobs:
   feature-checks:
--- a/bindings/rust/Cargo.lock
+++ b/bindings/rust/Cargo.lock
@@ -77,11 +77,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a3e2c3daef883ecc1b5d58c15adae93470a91d425f3532ba1695849656af3fc1"
 
 [[package]]
+name = "bytes"
+version = "1.5.0"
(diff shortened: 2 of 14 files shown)


jtraglia (approved):
Nice. Thanks guys. I think this looks great.

ralexstokes (approved):
haven't reviewed in depth but _strongly_ support any change away from using a feature flag to gate blob length

the current setup means you have to also gate all of your consumer code behind a feature flag as well which is not a very nice demand to make

michaelsproul (approved):
Looks good to me too based on a quick skim

asn-d6 (approved):
LGTM! Thanks for the updates!

PR #378, opened 2023-10-20
C# Ckzg.Bindings to become 0.4.0

flcl42:


jtraglia (approved):
LGTM. But is there a better long-term solution for this?

flcl42:
> LGTM. But is there a better long-term solution for this?

Yes it's possible to improve. I'd try to do the following:
- use the latest tag for x.y.z in x.y.z.t;
- use a counter variable that resets on new tag for t in x.y.z.t;
- maybe publish on demand only.

PR #380, opened 2023-11-11
feat: add missing new functions for byte types

mattsse:
currently there's no infallible way to create Bytes types

This adds `new(array) -> Self` functions

### DIFF
--- a/bindings/rust/src/bindings/mod.rs
+++ b/bindings/rust/src/bindings/mod.rs
@@ -266,7 +266,7 @@ impl Blob {
 }
 
 impl Bytes32 {
-    /// Creates a new blob from a byte array.
+    /// Creates a new instance from a byte array.
     pub const fn new(bytes: [u8; 32]) -> Self {
         Self { bytes }
     }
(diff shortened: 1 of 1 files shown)


jtraglia:
Hey @mattsse. There are functions for this here (might need to scroll btw):

https://github.com/ethereum/c-kzg-4844/blob/624aa60d01da524827123506975431aa5454c80d/bindings/rust/src/bindings/mod.rs#L487-L503

bindings/rust/src/bindings/mod.rs
```diff
@@ -261,6 +266,11 @@ impl Blob {
 }
 
 impl Bytes32 {
+    /// Creates a new blob from a byte array.
```
jtraglia, line 269: Nit: this comment shouldn't refer to "blob" type.

bindings/rust/src/bindings/mod.rs
```diff
@@ -280,6 +290,11 @@ impl Bytes32 {
 }
 
 impl Bytes48 {
+    /// Creates a new blob from a byte array.
```
jtraglia, line 293: Same thing here.

jtraglia (changes requested):
> hmm, I think having a const fn new would still be useful if you want to define constants

Alright yeah, I'm okay with this if it makes things easier for you guys. Just two minor nits.

It's pretty simple, but I'd appreciate another review by @pawanjay176 or @asn-d6 before merging.

jtraglia (approved):
LGTM 👍 On second thought, this is so minor I feel comfortable merging it now. Thanks!

PR #381, opened 2023-11-11
ci: ensure setuptools are installed

mattsse:
distutils is deprecated with python 3.12 and no longer included.
should fix Nodejs CI

ref https://github.com/ethereum/c-kzg-4844/pull/380


jtraglia (approved):
Oh interesting. What a weird dependency. Thanks for figuring this out. Thanks!

PR #386, opened 2024-01-23
feat: add pub methods for cell and export das items

jacobkaufmann:
add `pub` methods to `Cell` to retrieve inner data and KZG proof and to construct `Cell`.

export the following DAS items:

* `CELLS_PER_BLOB`
* `DATA_POINTS_PER_BLOB`
* `FIELD_ELEMENTS_PER_CELL`
* `Cell`

modify construction of boxed `Cell` array to not rely on `nightly` channel.

### DIFF
--- a/bindings/rust/rust-toolchain
+++ b/bindings/rust/rust-toolchain
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "nightly"
+channel = "stable"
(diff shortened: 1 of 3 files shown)


jacobkaufmann:
the proposed changes are necessary to implement DAS in e.g. `lighthouse`.

I'm also interested in a subsequent change to remove the inner `KZGProof` from the definition of `Cell`, but that would be more invasive.

jtraglia (approved):
Looks great. Thanks!

PR #388, opened 2024-01-31
fix: correct offset for cell 2D array construction

jacobkaufmann:
correct index offset in construction of `Cell` and add round-trip test for `Cell`.


jtraglia (approved):
Oops, good catch. Thanks!

PR #389, opened 2024-02-02
feat: enable blst portable by default

Rjected:
This is so we can enable the underlying blst `portable` feature in reth for creating portable release binaries

### DIFF
--- a/bindings/rust/Cargo.toml
+++ b/bindings/rust/Cargo.toml
@@ -8,7 +8,7 @@ links = "ckzg"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
-default = ["std"]
+default = ["std", "blst/portable"]
 std = ["hex/std", "libc/std", "serde?/std"]
 serde = ["dep:serde"]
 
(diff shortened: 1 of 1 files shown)


jtraglia:
Hey @Rjected thanks for the PR. This might have actually been an oversight on our part. The other bindings are using the portable version of blst by default, with the exception of Go because we can't control it from here. Semi-recently, the authors of blst added "cpu runtime detection" to the portable version of blst; so if the system is using a new cpu it will use optimized instructions and if it's an older cpu it will use the portable implementation. This is a long way of saying we should be using portable by default.

cc @asn-d6 @pawanjay176

jtraglia:
> Does this mean we should set `default = ["portable"]` or `default = ["blst/portable"]`?

Yes, and I slightly favor the second one but it's not a strong opinion.

Rjected:
> > Does this mean we should set `default = ["portable"]` or `default = ["blst/portable"]`?
> 
> Yes, and I slightly favor the second one but it's not a strong opinion.

makes sense, I just changed it to the second one

jtraglia (approved):
Looks great, thanks again! I will wait for another maintainer to approve before merging.

PR #390, opened 2024-02-03
feat: add more targets to ci

Rjected:
Adds `aarch64` targets to the rust CI alongside `x86_86`, including for macOS, using the new M1 runner. Also fixes an issue where `rust-cache` would not find `bindings/rust/Cargo.toml` due to the `workspace` field being unset.

### DIFF
--- a/.github/workflows/rust-tests.yml
+++ b/.github/workflows/rust-tests.yml
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - main
+      - dan/multi-arch-ci
   pull_request:
     branches:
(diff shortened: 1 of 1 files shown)


Rjected:
note: might remove the linux arch workflow as the benchmarks are ridiculously slow for some reason

jtraglia:
Looks good, but wow yeah let's remove the linux aarch64 workflow. That's ridiculously slow. The first run was 39 minutes. Re-ran it again and it was 41 minutes. Not sure why it's so slow.

PR #391, opened 2024-02-06
chore: make standalone portable feature, enabled by default

Rjected:
This way crates can do:
```toml
c-kzg = { version = "0.4.x", default-features = false }

[features]
default = ["portable"]
std = ["c-kzg/std"] # common pattern w/ std
portable = ["c-kzg/portable"]
```
instead of
```toml
c-kzg = { version = "0.4.1", default-features = false }
blst = { version = "0.3.11", default-features = false, optional = true }

[features]
default = ["portable"]
std = ["c-kzg/std"] # common pattern w/ std
portable = ["dep:blst", "blst?/portable"]
```


jtraglia (approved):
LGTM 👍

PR #394, opened 2024-02-21
V2

holiman:
Second attempt, making it a bit more go-idiomatic with a proper v2 module. 

First commit is, I think, pristine untouched v1, and changes come in commit two (and after)

### DIFF
--- a/bindings/go/v2/README.md
+++ b/bindings/go/v2/README.md
@@ -1,0 +1,37 @@
+# Go bindings
+
+This package implements Go bindings (using [Cgo](https://go.dev/blog/cgo)) for the
+exported functions in [C-KZG-4844](https://github.com/ethereum/c-kzg-4844).
+
+## Installation
+
+```
+go get github.com/ethereum/c-kzg-4844
+```
+
+## Go version
+
+This package requires `1.19rc1` or later. Version `1.19beta1` and before will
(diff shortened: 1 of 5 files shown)


holiman:
Hm, this is a bit cumbersome, ideally we'd just copy go.mod to the `v2` directory, however, the go.mod is sitting several directories "up". But I don't want to copy all non-go-relevant stuff in to v2, just the bare minimum. Not sure what's the Right Way to do it.. 

cc @karalabe ?

holiman:
I would also consider changing the return values?  Maybe something like this, and avoid the unconditional map-lookups on return: 

```golang
func ComputeBlobKZGProof(blob *Blob, commitmentBytes *Bytes48) (*KZGProof, error) {
	if !loaded {
		panic("trusted setup isn't loaded")
	}
	proof := KZGProof{}
	ret := C.compute_blob_kzg_proof(
		(*C.KZGProof)(unsafe.Pointer(&proof)),
		(*C.Blob)(unsafe.Pointer(blob)),
		(*C.Bytes48)(unsafe.Pointer(commitmentBytes)),
		&settings)

	if ret != C.C_KZG_OK{
		nil, makeErrorFromRet(ret)
	}
	return proof, nil
}
```

PR #395, opened 2024-02-21
Add workflow to publish python package

jtraglia:
This PR adds a new workflow which publishes the python module to PyPI.

Marking this as a draft until devops has added the API key to GitHub secrets.

@kclowes (from EF snake charmers team) requested this.

### DIFF
--- a/.github/workflows/python-release.yml
+++ b/.github/workflows/python-release.yml
@@ -1,9 +1,6 @@
 name: Python Package
 
 on:
-  push:
-    branches:
-      - main
   release:
     branches:
       - main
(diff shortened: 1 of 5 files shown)


jtraglia:
Okay, the API key should be in the repo's secrets now. Ready for review.

jtraglia:
Hmm I'm going to convert this to use `ciwheelhouse` which should simplify things a bit.

kclowes (approved):
This looks good to me @jtraglia. Thank you! I'm getting a deprecation warning from using `python3 setup.py install` [here](https://github.com/ethereum/c-kzg-4844/blob/main/bindings/python/Makefile#L9), but I did verify it's working! When I swapped that line locally with `python -m build` (the newer syntax), I was getting a `clang` error and haven't had time to debug. I can dig in more on Monday if you want, but it's just a deprecation warning so I think this is good to merge as-is. /cc @fselmo

jtraglia:
Hey @kclowes, thanks for the review! I think I'm okay with a deprecation warning.

I'll plan to merge this on Monday & make a new release to test it soon after.

PR #399, opened 2024-02-23
Define __BLST_NO_ASM__ for other ISAs

### LINKED ISSUE #398: Rust build.rs: Pass -D__BLST_NO_ASM__ on non-x86-64/non-aarch64 ISAs
The current build.rs does not pass the -D__BLST_NO_ASM__ flag when compiling for non-x86-64 / non-aarch64 ISAs.

https://github.com/supranational/blst/blob/0d46eefa45fc1e57aceb42bba0e84eab3a7a9725/build.sh#L91-L93

This is impacting zkVMs that target other ISAs like Risc-V, WASM, MIPS or plain LLVM IR.

Note: ultimately zkVMs will optimize blob proof verification using:
- https://notes.ethereum.org/@vbuterin/proto_danksharding_faq#Moderate-approach-works-with-any-ZK-SNARK
- and https://notes.ethereum.org/@dankrad/kzg_commitments_in_proofs

but this will still be helpful for testing and benchmarking.

jtraglia:
I have no way of testing this, but I believe this is what we need to do in order to support other architectures.

Does anyone have a system they could verify this works?

Might fix #398.

### DIFF
--- a/bindings/rust/build.rs
+++ b/bindings/rust/build.rs
@@ -28,6 +28,12 @@ fn main() {
         cc.flag("/std:c11");
     }
 
+    // Enable __BLST_NO_ASM__ if not x86_64 or aarch64.
+    let target_arch = env::var("CARGO_CFG_TARGET_ARCH").unwrap();
+    if !["x86_64", "aarch64"].contains(&target_arch.as_str()) {
+        cc.flag("-D__BLST_NO_ASM__");
+    }
+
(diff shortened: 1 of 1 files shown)


mratsim:
See https://github.com/ethereum/c-kzg-4844/issues/398#issuecomment-1960929643

It may be a toolchain/cross-compilation setup issue, continuing to investigate.

Issue #400, opened 2024-03-02
Python bindings' load_trusted_setup() segfaults if file does not exist


alexhooketh:
```
jordan@MacBook-Air-Alex python % python3
Python 3.11.7 (main, Dec  4 2023, 18:10:11) [Clang 15.0.0 (clang-1500.1.0.2.5)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import ckzg
>>> ckzg.load_trusted_setup("sfsfdsfdsfdsf.json") # does not exist
zsh: segmentation fault  python3
jordan@MacBook-Air-Alex python %
```

Probably related to [missing error checks on fopen](https://github.com/ethereum/c-kzg-4844/blob/main/bindings/python/ckzg.c#L21), I'm not a C dev so can't really say for sure

jtraglia:
@alexhooketh Thank you for reporting this! Definitely an oversight of ours.

PR #401, opened 2024-03-02
python: check file pointer when loading trusted setup

jtraglia:
Fixes: #400

Before:
```
>>> import ckzg
>>> ckzg.load_trusted_setup("sfsfdsfdsfdsf.json")
[1]    60039 segmentation fault  python
```

After:
```
>>> import ckzg
>>> ckzg.load_trusted_setup("sfsfdsfdsfdsf.json")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError: error reading trusted setup
```

### DIFF
--- a/bindings/python/ckzg.c
+++ b/bindings/python/ckzg.c
@@ -24,13 +24,14 @@ static PyObject* load_trusted_setup_wrap(PyObject *self, PyObject *args) {
     return PyErr_Format(PyExc_RuntimeError, "error reading trusted setup");
   }
 
-  if (load_trusted_setup_file(s, fp) != C_KZG_OK) {
+  C_KZG_RET ret = load_trusted_setup_file(s, fp);
+  fclose(fp);
+
+  if (ret != C_KZG_OK) {
     free(s);
-    fclose(fp);
     return PyErr_Format(PyExc_RuntimeError, "error loading trusted setup");
   }
 
-  fclose(fp);
   return PyCapsule_New(s, "KZGSettings", free_KZGSettings);
 }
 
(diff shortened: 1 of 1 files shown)


bindings/python/ckzg.c
```diff
@@ -10,19 +10,27 @@ static void free_KZGSettings(PyObject *c) {
...
+    return PyErr_Format(PyExc_RuntimeError, "error reading trusted setup");
+  }
+
+  if (load_trusted_setup_file(s, fp) != C_KZG_OK) {
     free(s);
+    fclose(fp);
     return PyErr_Format(PyExc_RuntimeError, "error loading trusted setup");
   }
 
+  fclose(fp);
```
asn-d6, line 33: Seems like the convention in the other bindings is to do something like:
```
	fp := C.fopen(cTrustedSetupFile, cMode)
	if fp == nil {
		panic("error reading trusted setup")
	}
	ret := C.load_trusted_setup_file(&settings, fp)
	C.fclose(fp)
	if ret == C.C_KZG_OK {
		loaded = true
	}
```

where we defer the retval check of `load_trusted_setup_file()` so that we can unconditionally `fclose()` once. Should we also adapt this code to that convention?

I think we are also missing an fclose on the error path of the Rust bindings. I can make a new issue for this or handle it here.

jtraglia: Yes, good suggestion. I will make this change. And yes to a new PR for the Rust bindings.

asn-d6: Pushed at 40f637693

asn-d6: Opened https://github.com/ethereum/c-kzg-4844/pull/402 for the Rust bindings.

PR #402, opened 2024-03-04
Rust: Always close fd on load_trusted_setup_file_inner()

asn-d6:


jtraglia (approved):
LGTM, thanks @asn-d6!

PR #403, opened 2024-03-04
Update versions in preparation of release

jtraglia:
This PR updates the versions to 0.4.3 before the new release, so it's correct.

I'm only uncertain about the nodejs version. Don't think I can go from v2 to v0. So just change the minor/patch numbers.


asn-d6 (approved):
LGTM!

PR #404, opened 2024-03-04
Upload python wheels on Linux runner

jtraglia:
So I made a mistake. `gh-action-pypi-publish` only works on Linux. So when this action ran, it failed when trying to publish the macOS packages and cancelled the other platform workflows.

Also, I didn't realize there were multiple types of "releases" like "created", "published", etc. This PR adds a line which should cause this to only run once when a release is made.

image


jtraglia:
I thought I would be able to publish via a PR. Probably good this isn't possible.

image

I will merge & get an admin to run this action manually. So we don't need to create a new release.

PR #406, opened 2024-03-04
Add more information in setuptools.setup()

jtraglia:
For some reason the Python package publishing is failing because of the long description. Hopefully this fixes it.

There are a bunch of errors like this:
```
ERROR    `long_description` has syntax errors in markup and would not be        
         rendered on PyPI.                                                      
         No content rendered from RST source.                                   
WARNING  `long_description_content_type` missing. defaulting to `text/x-rst`.  
```

### DIFF
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -1,4 +1,8 @@
 from setuptools import setup, Extension
+from pathlib import Path
+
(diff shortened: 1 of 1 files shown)


jtraglia:
Going to merge this so I can test publishing. If there's a problem we can fix it later.

PR #407, opened 2024-03-07
feat: derive hash for generated types

mattsse:
This makes it lot easier to use the type in other structs that require hash.


asn-d6:
LGTM! Thanks!

Issue #392, opened 2024-02-19
Rust bindings lack flexibility with current structure of build script


ralexstokes:
I'm using a niche devops flow with [nixos](https://nixos.org) and I've found the current structure of the Rust bindings in this repo are problematic. As far as I can tell, this has to do with the placement of the `build.rs` in a subdirectory with respect to the C code. I expect there were similar issues with other language toolchains given the placement of the Go and Nim files at the repo root.

I can point to these resources that helped me diagnose the issue:

https://github.com/ipetkov/crane/issues/473#issuecomment-1837004284
https://github.com/ipetkov/crane/pull/271#issuecomment-1500499998

And in particular (regardless of `nixos`), you'll find that trying to build the bindings locally using `cargo vendor` fails.

I haven't tried it yet, but I suspect we should just place the `Cargo.toml` and `build.rs` as the project root. If we are open to modifying the repo to get `cargo vendor` to succeed then I'm happy to make a PR to diagnose further.

There seems to be a quirk of `crates.io` that pulls the entire repo so that [something like this](https://github.com/ralexstokes/lighthouse/commit/39ce7765011723fcf1c783e0007cd30675c67a08) works for my use case, but I would prefer we pursue a successful `cargo vendor` as that should solve the problem instead of having to maintain a fork of all the consumers of this project.

jtraglia:
> If we are open to modifying the repo to get cargo vendor to succeed then I'm happy to make a PR to diagnose further.

Hey @ralexstokes! This is an interesting problem. I'm open to changing things. But I would like to hear about this from other Rust people before suggesting we do anything. @pawanjay176 @asn-d6, thoughts?

asn-d6:
Hmm, it feels dirty to dump `Cargo.toml` and `build.rs` to the project root, but it seems like we already have golang and nim-related files there. So dirtiness is already there, and hence I wouldn't oppose this change.

I wonder if it's possible to do something using `cargo` workspaces.  From reading a bit about it, even if it's possible, we would still need a `Cargo.toml` in the project root to specify the location of the rust bindings.

The extreme (not proposed) alternative is to move all bindings to their own git repo so that they can freely manage their directories.

akorchyn:
+1 to this issue. I'm using nixos for my DevOps too and it fails to work:(
This is because of strict rules for what application can access, and it fails to vendor the crate.

PR #409, opened 2024-03-12
Fixed NixOs vendoring issue

akorchyn:
Resolves: #392

You can test it by
```
cargo package
```


akorchyn:
Sorry, I missed one spot in the build.rs :)
Should be fine now, I believe.

Cargo.toml
```diff
@@ -27,17 +32,17 @@ blst = { version = "0.3.11", default-features = false }
 hex = { version = "0.4.2", default-features = false, features = ["alloc"] }
 libc = { version = "0.2", default-features = false }
 serde = { version = "1.0", optional = true, default-features = false, features = [
-    "alloc",
-    "derive",
+	"alloc",
+	"derive",
```
ralexstokes, line 36: not a big deal, but do we need to change the whitespace here?

Cargo.toml
```diff
@@ -27,17 +32,17 @@ blst = { version = "0.3.11", default-features = false }
...
 ] }
 
 [dev-dependencies]
 criterion = "0.5.1"
 glob = "0.3.1"
 rand = "0.8.5"
 serde = { version = "1.0", features = ["derive"] }
-serde_yaml = "0.9.17"
 serde_json = "1.0.105"
+serde_yaml = "0.9.17"
```
ralexstokes, line 45: same question w/ reordering the deps here

ralexstokes (commented):
nice! thanks for this :)

haven't tested out e.g. `cargo package` or `cargo vendor` yet but the diff generally looks good

Cargo.toml
jtraglia: I think this is a good change. It was alphabetically out-of-order before.

Cargo.toml
akorchyn: Restored original whitespaces.

jtraglia (approved):
LGTM 🥇thanks again @akorchyn! Just want some confirmations from the team before merging.

* @ralexstokes could you confirm that this fixes your issue?
* @Rjected & @pawanjay176 will this still work with Reth & Lighthouse?

pawanjay176 (approved):
Ran CI on lighthouse with this branch and it looks fine, changes also make sense to me

Rjected (approved):
this looks good to me, should work with revm / reth / alloy. moving cargo.toml to the top level makes sense

PR #411, opened 2024-03-21
feat: impl AsRef for Bytes32 and Bytes48

Rjected:
These impls are nice to have and trivial to impl


jtraglia (approved):
LGTM 👍 makes sense to have these, thanks!

PR #412, opened 2024-03-26
use strutils.escape to handle special characters in KZG directory paths in Nim bindings

tersec:
- https://github.com/status-im/nimbus-eth2/issues/6093

### DIFF
--- a/bindings/nim/kzg_abi.nim
+++ b/bindings/nim/kzg_abi.nim
@@ -22,8 +22,8 @@ when not defined(kzgExternalBlst):
 
 {.compile: srcPath & "c_kzg_4844.c"}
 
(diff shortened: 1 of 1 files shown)


jtraglia (approved):
Oh that's a good thing to fix. LGTM 👍

@tersec would you like us to cut a new v1.0.1 release after merging this?

tersec:
> Oh that's a good thing to fix. LGTM 👍
> 
> @tersec would you like us to cut a new v1.0.1 release after merging this?

That would be useful, yes

jtraglia:
Okay no problem. I'd like to confirm it with the team first. Should have it out by tomorrow morning.

PR #419, opened 2024-04-22
In Python setup file, change directory instead of using absolute paths

jtraglia:
This PR should fix a Windows issue associated with #415.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,35 +1,27 @@
+from os import chdir
 from pathlib import Path
 from platform import system
 from setuptools import setup, Extension
 from setuptools.command.build_ext import build_ext
 from subprocess import check_call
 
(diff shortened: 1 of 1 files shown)


jtraglia:
For some reason, this fixes an issue I'm having building/testing the python bindings on my development machine.

```
No eggs found in /var/folders/qb/vp1kjmrs03zczc9rnd4mdg9m0000gn/T/easy_install-fagxc9ra/ckzg-1.0.1/egg-dist-tmp-9ugrvgp2 (setup script problem?)
error: The 'ckzg==1.0.1' distribution was not found and is required by the application
make: *** [install] Error 1
```

With this PR, things just work. So I'm going to merge this.

PR #420, opened 2024-04-29
Fix broken nodejs/csharp CI tests

jtraglia:
Noticed some issues with macOS on these tests. This PR investigates & fixes these issues.

* For the nodejs tests, Python wasn't properly setup first.
  * This prevented `setuptools` from being installed. 
* For csharp, `macos-latest` now defaults to an arm64 runner.
  *  We were attempting to use the x86_64 library, which failed to load.


asn-d6 (approved):
LGTM! Thanks!

Babbz86 (commented):
Awesome fixing node ja chsarp tests for cl

Babbz86 (commented):
Fixing node ja using chsarp and cl 😀

PR #421, opened 2024-04-29
feat: update node bindings

matthewkeil:
Updates README.md with new DAS functions.

**note: also turned on exceptions for node.  Is a bug that needed to be fixed.  Node can potentially throw errors and if exceptions are not turned on there are additional checks during allocations and getting data from the isolate that need to happen.  Those checks are extremely verbose.  Is better to turn on exceptions**

https://github.com/nodejs/node-addon-api/blob/main/doc/error_handling.md#handling-errors-with-maybe-type-and-c-exceptions-disabled

### DIFF
--- a/bindings/node.js/README.md
+++ b/bindings/node.js/README.md
@@ -7,7 +7,7 @@ API. The core functionality was originally a stripped-down copy of
 since then. This package wraps that native `c-kzg` C code in C/C++ NAPI
 bindings for use in node.js applications.
 
-Spec: https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/polynomial-commitments.md
+Spec: <https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/polynomial-commitments.md>
 
 ## Prerequisites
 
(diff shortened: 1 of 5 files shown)


bindings/node.js/README.md
```diff
@@ -7,7 +7,7 @@ API. The core functionality was originally a stripped-down copy of
 since then. This package wraps that native `c-kzg` C code in C/C++ NAPI
 bindings for use in node.js applications.
 
-Spec: https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/polynomial-commitments.md
+Spec: <https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/polynomial-commitments.md>
```
jtraglia, line 10: This link is broken. Can we replace it with the following two links:

* https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/polynomial-commitments.md
* https://github.com/ethereum/consensus-specs/blob/dev/specs/_features/eip7594/polynomial-commitments-sampling.md

jtraglia (commented):
Looks good, thanks! Just one small thing.

bindings/node.js/README.md
matthewkeil: Updated.

matthewkeil:
@jtraglia Converted to draft until the other PR get approved/denied as I move the exception commit over and will revert that here

jtraglia (approved):

PR #422, opened 2024-04-30
feat: bundle and auto-load trusted setup

matthewkeil:
**Motivation**

Allows for easier initialization of library for Ethereum use cases.  Also bundles the trusted setup so that it does not need to be manually downloaded by consumers and stored separately.  Will simplify the `lodestar` use case.  Defaults to using the file passed by the user first.  If none is passed use the one found in the dist bundle (checks here for the prod case first).  If one is not found in the prod location fall back to the src location (for development purposes).  Still throws error if none is sourced (invalid bundle/build).

**Description of Changes**
- Bundle `trusted_setup.txt` with the c-kzg deps for node.js bindings
- Update `kzg.js` to functions to allow for passing a trusted setup or for loading the default setup from either the bundle or the repo src
- Export `getTrustedSetupFilepath` for testing purposes but do not announce via types file.
- Unit test `getTrustedSetupFilepath` to ensure proper loading.

### DIFF
--- a/bindings/node.js/README.md
+++ b/bindings/node.js/README.md
@@ -7,7 +7,7 @@ API. The core functionality was originally a stripped-down copy of
 since then. This package wraps that native `c-kzg` C code in C/C++ NAPI
 bindings for use in node.js applications.
 
-Spec: <https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/polynomial-commitments.md>
+Spec: <https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/polynomial-commitments.md>
 
 ## Prerequisites
 
@@ -64,11 +64,11 @@ const isValid = verifyBlobKzgProofBatch(blobs, commitments, proofs);
  * ".txt" extension.
  *
  * Uses user provided location first. If one is not provided then defaults to
- * the official Ethereum mainnet setup from the kzg ceremony. Should only be
- * used for cases where the Ethereum official mainnet kzg setup is acceptable.
+ * the official Ethereum mainnet setup from the KZG ceremony. Should only be
+ * used for cases where the Ethereum official mainnet KZG setup is acceptable.
  *
  * @param {string | undefined} filePath - .txt/.json file with setup configuration
- * @default - If no string is passed the default trusted setup from the Ethereum kzg ceremony is used
+ * @default - If no string is passed the default trusted setup from the Ethereum KZG ceremony is used
  *
  * @throws {TypeError} - Non-String input
  * @throws {Error} - For all other errors. See error message for more info
--- a/bindings/node.js/README.md
+++ b/bindings/node.js/README.md
@@ -7,7 +7,7 @@ API. The core functionality was originally a stripped-down copy of
 since then. This package wraps that native `c-kzg` C code in C/C++ NAPI
 bindings for use in node.js applications.
 
-Spec: https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/polynomial-commitments.md
+Spec: <https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/polynomial-commitments.md>
 
 ## Prerequisites
 
@@ -56,14 +56,22 @@ const isValid = verifyBlobKzgProofBatch(blobs, commitments, proofs);
 
 ```ts
 /**
- * Sets up the c-kzg library. Pass in a properly formatted trusted setup file
- * to configure the library.  File must be in json format, see TrustedSetupJson
- * interface for more details, or as a properly formatted utf-8 encoded file.
+ * Initialize the library with a trusted setup file.
  *
- * @remark This function must be run before any other functions in this
- *         library can be run.
+ * Can pass either a .txt or a .json file with setup configuration. Converts
+ * JSON formatted trusted setup into the native format that the base library
+ * requires. The created file will be in the same as the origin file but with a
+ * ".txt" extension.
  *
- * @param {string} filePath - The absolute path of the trusted setup
+ * Uses user provided location first. If one is not provided then defaults to
+ * the official Ethereum mainnet setup from the kzg ceremony. Should only be
+ * used for cases where the Ethereum official mainnet kzg setup is acceptable.
+ *
+ * @param {string | undefined} filePath - .txt/.json file with setup configuration
+ * @default - If no string is passed the default trusted setup from the Ethereum kzg ceremony is used
+ *
+ * @throws {TypeError} - Non-String input
+ * @throws {Error} - For all other errors. See error message for more info
  */
 loadTrustedSetup(filePath: string): void;
 ```
--- a/bindings/node.js/binding.gyp
+++ b/bindings/node.js/binding.gyp
@@ -33,7 +33,7 @@
           ],
           "msbuild_settings": {
             "ClCompile": {
-              "ExceptionHandling": "/EHsc",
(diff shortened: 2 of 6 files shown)


bindings/node.js/lib/kzg.js
```diff
@@ -6,6 +6,28 @@ const fs = require("fs");
...
+ * used for cases where the Ethereum official mainnet kzg setup is acceptable.
+ *
+ * @returns {string | undefined} - Filepath for trusted_setup.txt if found
+ */
+function getDefaultTrustedSetupFilepath() {
+  const locationsToSearch = [
+    // check the production bundle case first (this file in lib)
+    path.resolve(__dirname, "..", "deps", "c-kzg", "trusted_setup.txt"),
+    // check the development in-repo case second (this file in bindings/node.js/lib)
+    path.resolve(__dirname, "..", "..", "..", "src", "trusted_setup.txt"),
```
asn-d6, line 21: Do we need this auto-detection logic? Why can't the user of the library always specify where the trusted setup is? Just worrying that the amount of `..` here makes this brittle.

What use cases are there, where someone loads a library and they don't know where the trusted setup file is, but the library knows?

bindings/node.js/README.md
```diff
@@ -7,7 +7,7 @@ API. The core functionality was originally a stripped-down copy of
 since then. This package wraps that native `c-kzg` C code in C/C++ NAPI
 bindings for use in node.js applications.
 
-Spec: https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/polynomial-commitments.md
+Spec: <https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/polynomial-commitments.md>
```
jtraglia, line 10: Can we fix this link too?
```suggestion
Spec: <https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/polynomial-commitments.md>
```

bindings/node.js/lib/kzg.js
```diff
@@ -31,19 +53,49 @@ function transformTrustedSetupJson(filePath) {
...
+ *
+ * @param {string} filePath - User provided filePath to check for trusted setup
+ *
+ * @returns {string} - Location of a trusted setup file. Validity is checked by
+ *                     the native bindings.loadTrustedSetup
+ *
+ * @throws {TypeError} - Invalid file type
+ * @throws {Error} - Invalid location or no default trusted setup found
+ *
+ * @remarks - This function is only exported for testing purposes.  It should
```
jtraglia, line 69: Double space between sentences should be removed.
```suggestion
 * @remarks - This function is only exported for testing purposes. It should
```

bindings/node.js/lib/kzg.js
```diff
@@ -31,19 +53,49 @@ function transformTrustedSetupJson(filePath) {
...
-bindings.loadTrustedSetup = function loadTrustedSetup(filePath) {
-  if (!(filePath && typeof filePath === "string")) {
-    throw new TypeError("must initialize kzg with the filePath to a txt/json trusted setup");
-  }
-  if (!fs.existsSync(filePath)) {
-    throw new Error(`no trusted setup found: ${filePath}`);
+/**
+ * Gets location for trusted setup file. Uses user provided location first. If
+ * one is not provided then defaults to the official Ethereum mainnet setup from
+ * the kzg ceremony.
```
jtraglia, line 59: Nit: in the new code/comments, can we replace "kzg" with "KZG"? Except for "c-kzg" of course.
```suggestion
 * the KZG ceremony.
```

jtraglia (commented):
This seems reasonable to me. I like that users can still specify a trusted setup.

bindings/node.js/README.md
matthewkeil: Updated.

bindings/node.js/lib/kzg.js
matthewkeil: Updated it here and in all other locations "kzg" was in the docs and docstrings.

bindings/node.js/lib/kzg.js
matthewkeil: Updated.

jtraglia:
@matthewkeil what does "enabling exceptions" do exactly and what are the implications?

bindings/node.js/lib/kzg.js
matthewkeil: > Do we need this auto-detection logic? Why can't the user of the library always specify where the trusted setup is? Just worrying that the amount of .. here makes this brittle.

Sadly yes.  This an unfortunate pattern that shows up often in JS repos.  Generally src and artifacts are found in different places and usually prod code runs with artifacts (using transpiled JS) and unit test use src file running with pure TypeScript code and  `ts-node` under the hood somewhere.

> What use cases are there, where someone loads a library and they don't know where the trusted setup file is, but the library knows?

If a consumer/user supplies a value and it is invalid the library will throw and will not load a default config.  This way the consumer will not end up inadvertently loading a setup they were not expecting to load.  The default versions are ONLY used if no location is passed to the `loadTrustedSetup` function.

bindings/node.js/binding.gyp
```diff
@@ -28,15 +29,18 @@
           "sources": ["deps/blst/build/win64/*-x86_64.asm"],
           "defines": [
             "_CRT_SECURE_NO_WARNINGS",
+            "_HAS_EXCEPTIONS=1"
           ],
           "msbuild_settings": {
             "ClCompile": {
+              "ExceptionHandling": "Sync",
```
matthewkeil, line 36: @jtraglia @asn-d6 I think this is the correct value here but would love a double check.  There is also an option for `SyncCThrow` but the docs recommend `Sync`.

https://learn.microsoft.com/en-us/visualstudio/msbuild/cl-task?view=vs-2022
https://learn.microsoft.com/en-us/cpp/build/reference/eh-exception-handling-model?view=msvc-170

bindings/node.js/lib/kzg.js
asn-d6: > @asn-d6 sadly yes. This an unfortunate pattern that shows up often in JS repos. Generally src and artifacts are found in different places and usually prod code runs with artifacts (using transpiled JS) and unit test use src file running with pure TypeScript code and ts-node under the hood somewhere.

Hm. It's still not clear to me why we can't have prod code and unit tests always pass the right file path to the `loadTrustedSetup` function. This way, we could remove this `getDefaultTrustedSetupFilepath()` function and simplify the code.

jtraglia:
@matthewkeil thanks for the explanation. I think this is a bit over my head. Could you split the exception changes into a separate PR and ask a teammate (or just someone that would know) to review it?

matthewkeil:
@jtraglia exceptions moved to #423

jtraglia (approved):
LGTM. Thanks for making those changes!

bindings/node.js/lib/kzg.js
matthewkeil: I will see what I can do to simplify things.  This is how I've learned to do an how I've always done it.  Will have to double check that I catch all the edge cases to make sure it works everywhere.  Stay-tuned @asn-d6.  Will try to get this pushed tomorrow morning (is almost midnight here).

matthewkeil: Will do!!  Sorry @asn-d6 will get this up asap

matthewkeil: Updated in https://github.com/ethereum/c-kzg-4844/pull/424

PR #424, opened 2024-05-03
fix: simplify how default setup is found

matthewkeil:
Simplifies the logic for importing the default `trusted_setup.txt` in the node.js bindings.


asn-d6:
Thanks! Merging.

PR #427, opened 2024-05-07
fix: version and readme alpha note

matthewkeil:
Small update so the version is correct when the das branch gets merged to main.  Don't want to accidentally publish to the wrong version.  Also removed the note about alpha versions.  Wont get published until that is no longer true

### DIFF
--- a/bindings/node.js/README.md
+++ b/bindings/node.js/README.md
@@ -1,7 +1,5 @@
 # C-KZG-4844
 
-**Note: This is an alpha release with peerDas functionality.  For the current stable release use v3**
-
 This is a TypeScript library for EIP-4844 that implements the [Polynomial
 Commitments](https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/polynomial-commitments.md)
 API. The core functionality was originally a stripped-down copy of
--- a/bindings/node.js/package.json
+++ b/bindings/node.js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "c-kzg",
-  "version": "4.0.0-alpha.0",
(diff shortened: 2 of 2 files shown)


matthewkeil:
@jtraglia I saw PR #428 and the node bindings will need to be `v4.0.0` for the DAS work.  Because the alpha version got released accidentally there is no way to revert to `v3.X.X`... the `v4.0.0-alpha.0` version will always be the landing screen for `c-kzg` because its the largest according to semver.  I released a `v4.0.0-alpha.1` with the code for default trusted setup a couple of days ago to just get rid of that deprecation warning banner in case anyone sees the package website.  @g11tech and I have decided to go with `v4.0.0` as the official version with the DAS code once that gets released and is merged to main.  So the update in #428 is not going to work.  This PR should get merged into the `das` branch to update that.

jtraglia:
> go with v4.0.0 as the official version with the DAS code once that gets released and is merged to main.

Why are we doing this now? We are months away from merging this into main. What will removing "alpha" change exactly? Are you planning to make a v4.0.0 release soon? Having "alpha" in the version name is the sole reason I was semi-okay with the v4 release in the first place. The `das` branch contains unaudited code which should not be used in production. If you remove "alpha" and remove the note at the top, what's to prevent someone outside of this circle from knowing that this version should not be used yet?

> So the update in https://github.com/ethereum/c-kzg-4844/pull/428 is not going to work.

Why won't #428 work exactly? You can still make v3 releases, no? It not being at the top isn't that big of a deal.

matthewkeil:
Closing and will reopen when the `das` branch merged to `main`

PR #428, opened 2024-05-07
Update versions in preparation for release

jtraglia:
This PR updates versions in the bindings.

PS: We forgot to do this for v1.0.1, so for most bindings we jump from v1.0.0 to v1.0.2.


kevaundray (approved):

Nichebiche (commented):
How can this be easier

PR #430, opened 2024-05-26
nim-bindings: Fix GCC-14 [-Wincompatible-pointer-types] issue

jangko:


asn-d6:
Hey @jangko . Thanks for the PR!

I'm leaning towards ACKing it, but since it involves some low-level nim decisions, would it be possible to get another person of your team (i.e. another nim expert) to review your PR?

jangko:
ping @tersec @arnetheduck @cheatfate

jangko:
After discussion with other nimbus developers, we decide to take one step further by introducing breaking changes. Instead of  Nim-array to C-struct conversion using pointer casting, we choose to use the C types directly in both `kzg.nim` and `kzg_ex.nim` and avoid aliasing rules violation altogether.

Further review and approval will come after we solve issues in other wrapper libraries to make sure no more surprises from gcc-14.

bindings/nim/kzg_abi.nim
```diff
@@ -86,32 +92,32 @@ proc load_trusted_setup_file*(res: KzgSettings,
 
 proc free_trusted_setup*(s: KzgSettings) {.kzg_abi.}
 
-proc blob_to_kzg_commitment*(res: var KzgCommitment,
-                         blob: KzgBlob,
+proc blob_to_kzg_commitment*(res: ptr KzgCommitment,
```
arnetheduck, line 95: we should be able to keep the `var` here, no? given this is an "out" argument required to be non-null..

jangko: thanks, restored.

jtraglia (approved):
LGTM 👍 I don't see any obvious problems.

PR #432, opened 2024-06-03
use stand-alone `results` library

arnetheduck:
It hasn't been part of `stew` for a good while now - also fix version number in nimble file.


kzg4844.nimble
```diff
@@ -5,7 +5,7 @@ mode = ScriptMode.Verbose
 ##################################################
 
 packageName   = "kzg4844"
-version       = "0.1.0"
+version       = "1.0.1"
```
jtraglia, line 8: Oh nice catch. We're actually at v1.0.2 right now though.
```suggestion
version       = "1.0.2"
```

arnetheduck: oops, forgot to fetch ;)

jtraglia (approved):
LGTM! Thanks 😄

PR #434, opened 2024-06-05
upstream: computeCells

agnxsh:
`computeCells( )` hadn't been upstreamed earlier, this fixes that.

### DIFF
--- a/bindings/nim/kzg_ex.nim
+++ b/bindings/nim/kzg_ex.nim
@@ -187,6 +187,9 @@ template verifyBlobKzgProofBatch*(blobs: openArray[KzgBlob],
                    proofs: openArray[KzgBytes48]): untyped =
   verifyProofs(blobs, commitments, proofs)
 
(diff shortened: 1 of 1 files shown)


jtraglia:
I don't think this is necessary if it's the same name. Those are just aliases. The function is here:

https://github.com/ethereum/c-kzg-4844/blob/230e13baa7d2f1e8dcb05fcaf888bee2c6f64df5/bindings/nim/kzg_ex.nim#L146-L149

agnxsh:
> I don't think this is necessary if it's the same name. Those are just aliases. The function is here:
> 
> 
> 
> https://github.com/ethereum/c-kzg-4844/blob/230e13baa7d2f1e8dcb05fcaf888bee2c6f64df5/bindings/nim/kzg_ex.nim#L138-L140

Yes agreed, didn't quite catch that

jtraglia:
Ah okay, no worries. I'm going to close this then. Thank you!

PR #437, opened 2024-06-12
[nim] Add completeStruct pragma which workarounds Nim compiler's issue with C codegen.

cheatfate:
This helps to avoid Nim codegen issues when KZG types become parts of pure Nim objects.
PR works for both `pre-gcc-14` and `gcc-14`.


jtraglia (approved):
LGTM 👍 thanks @cheatfate!

PR #440, opened 2024-06-18
Rust: load_trusted_setup() API shouldnt care about BYTES_PER_*_POINT

asn-d6:
Rust clients had to know the `BYTES_PER_G1_POINT` and `BYTES_PER_G2_POINT` constants to be able to use `load_trusted_setup()`. Ideally this would not be the case. The clients should not be exposed to cryptographic details.

This small modification makes the function take flat byte vectors as input which is also what the c-kzg backend expects. This way Rust clients don't need to chunk bytes into group points themselves, they just supply the flat byte vector and c-kzg takes care of the rest.

### DIFF
--- a/bindings/rust/src/bindings/mod.rs
+++ b/bindings/rust/src/bindings/mod.rs
@@ -95,9 +95,9 @@ pub fn hex_to_bytes(hex_str: &str) -> Result<Vec<u8>, Error> {
 
 /// Holds the parameters of a kzg trusted setup ceremony.
 impl KZGSettings {
-    /// Initializes a trusted setup from a flat array of `FIELD_ELEMENTS_PER_BLOB` G1 points in
-    /// Lagrange form, a flat array of 65 G2 points in monomial form, and a flat array of
-    /// `FIELD_ELEMENTS_PER_BLOB` G1 points in monomial form.
+    /// Initializes a trusted setup from a flat array of `FIELD_ELEMENTS_PER_BLOB` G1 points in monomial form, a flat
+    /// array of `FIELD_ELEMENTS_PER_BLOB` G1 points in Lagrange form, and a flat array of 65 G2 points in monomial
+    /// form.
     pub fn load_trusted_setup(
         g1_monomial_bytes: &[u8],
         g1_lagrange_bytes: &[u8],
--- a/bindings/rust/src/bindings/mod.rs
+++ b/bindings/rust/src/bindings/mod.rs
@@ -104,14 +104,14 @@ impl KZGSettings {
         g2_monomial_bytes: &[u8],
         precompute: usize,
     ) -> Result<Self, Error> {
-        if g1_monomial_bytes.len() != BYTES_PER_BLOB {
+        if g1_monomial_bytes.len() != FIELD_ELEMENTS_PER_BLOB*BYTES_PER_G1_POINT {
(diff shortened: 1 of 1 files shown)


asn-d6:
Right now it's only reth who uses `load_trusted_setup()` and they haven't integrated the `das` branch yet because as an EL client they don't care about PeerDAS.

When we merge `das` into `main` we will need to communicate with them to make sure they adapt their use of `load_trusted_setup()` accordingly.

Regardless of this PR they will most likely have to make changes anyway so that they load the `g1_monomial` points from a file (which was not the case for 4844).

bindings/rust/src/bindings/mod.rs
```diff
@@ -95,23 +95,23 @@ pub fn hex_to_bytes(hex_str: &str) -> Result<Vec<u8>, Error> {
...
-        if g1_monomial_bytes.len() != FIELD_ELEMENTS_PER_BLOB {
+        if g1_monomial_bytes.len() != BYTES_PER_BLOB {
             return Err(Error::InvalidTrustedSetup(format!(
                 "Invalid number of g1 monomial points in trusted setup. Expected {} got {}",
                 FIELD_ELEMENTS_PER_BLOB,
                 g1_monomial_bytes.len()
             )));
         }
-        if g1_lagrange_bytes.len() != FIELD_ELEMENTS_PER_BLOB {
+        if g1_lagrange_bytes.len() != BYTES_PER_BLOB {
```
jtraglia, line 114: For these two checks, I think I would prefer to use `FIELD_ELEMENTS_PER_BLOB*BYTES_PER_G1_POINT` for clarity and consistency with the G2 points check.

bindings/rust/src/bindings/mod.rs
```diff
@@ -95,23 +95,23 @@ pub fn hex_to_bytes(hex_str: &str) -> Result<Vec<u8>, Error> {
 
 /// Holds the parameters of a kzg trusted setup ceremony.
 impl KZGSettings {
-    /// Initializes a trusted setup from `FIELD_ELEMENTS_PER_BLOB` g1 points in
-    /// Lagrange form, 65 g2 points in monomial form, and
-    /// `FIELD_ELEMENTS_PER_BLOB` g1 points in monomial form.
+    /// Initializes a trusted setup from a flat array of `FIELD_ELEMENTS_PER_BLOB` G1 points in
+    /// Lagrange form, a flat array of 65 G2 points in monomial form, and a flat array of
+    /// `FIELD_ELEMENTS_PER_BLOB` G1 points in monomial form.
```
jtraglia, line 100: This comment isn't exactly correct.

It says g1 Lagrange, g2 monomial, g1 monomial.
It should be g1 monomial, g1 Lagrange, then g2 monomial.

jtraglia (commented):
This very close to what I had in mind. Just some small things to fix first.

bindings/rust/src/bindings/mod.rs
asn-d6: Fixed in c4260665b

bindings/rust/src/bindings/mod.rs
asn-d6: Fixed in 75fb1ae3c

jtraglia (approved):
LGTM, thanks!

PR #441, opened 2024-06-19
DAS Branch: Improve Documentation and Variable Naming

b-wagn:
This PR  aims to improve readability and documentation of the code in the `das` branch ([this one](https://github.com/ethereum/c-kzg-4844/tree/das)).

The following has been done:
* Function `verify_cell_kzg_proof_batch`: A comment was wrong (powers of `r` start at `r^0`).
* Function `verify_cell_kzg_proof`: Comments and function documentation improved
* Function `verify_kzg_proof_multi_impl`: Comments, documentation, variable names improved + minor code reordering

### DIFF
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -3399,7 +3399,7 @@ static C_KZG_RET verify_kzg_proof_multi_impl(
     /* Compute [commitment - interpolated_poly] in G_1 */
     g1_sub(&commit_minus_interp, commitment, &interpolated_poly);
 
-    /* e(\pi, [s^n - h^n]) =?= e([p(x) - I(x)], [1]) */
+    /* Check e([p(x) - I(x)], [1]) =?= e(proof, [s^n - h^n])  */
     *out = pairings_verify(
         &commit_minus_interp, blst_p2_generator(), proof, &s_pow_minus_h_pow
     );
@@ -4012,7 +4012,7 @@ C_KZG_RET verify_cell_kzg_proof_batch(
 
     /*
      * Derive random factors for the linear combination. The exponents start
-     * with 1, for example r^1, r^2, r^3, and so on.
+     * with 0. That is, they are r^0, r^1, r^2, r^3, and so on.
      */
     ret = compute_r_powers_for_verify_cell_kzg_proof_batch(
         r_powers,
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -3398,22 +3398,22 @@ static C_KZG_RET verify_kzg_proof_multi_impl(
     if (ret != C_KZG_OK) goto out;
 
     ///////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


src/c_kzg_4844.c
```diff
@@ -3383,25 +3391,37 @@ static C_KZG_RET verify_kzg_proof_multi_impl(
...
 
-    /* Compute [x^n] in G_2 */
+    /* Commit to the interpolation polynomial, i.e., get [I(tau)] in G_1 */
+    ret = poly_to_kzg_commitment_monomial(
+        &interpolation_poly_g1, interpolation_poly, n, s
+    );
+    if (ret != C_KZG_OK) goto out;
+
+    ///////////////////////////////////////////////////////////////////////////
+    // STEP 2: Compute the vanishing polynomial V(X) (as a commitment in G_2).
```
asn-d6, line 3401: Let's use Z(x) for the vanishing since that's what the specs are using.

src/c_kzg_4844.c
```diff
@@ -3383,25 +3391,37 @@ static C_KZG_RET verify_kzg_proof_multi_impl(
...
+    //         It is zero over the coset. In our case: V(X) = X^n - h^n
+    ///////////////////////////////////////////////////////////////////////////
+
+    /* Compute [h^n] in G_2 */
     blst_fr_eucl_inverse(&h_pow, &inv_h_pow);
     g2_mul(&h_pow_g2, blst_p2_generator(), &h_pow);
 
-    /* Compute [s^n - h^n] in G_2 */
-    g2_sub(&s_pow_minus_h_pow, &s->g2_values_monomial[n], &h_pow_g2);
+    /* Compute [V(tau)] = [tau^n - h^n] in G_2 */
```
asn-d6, line 3409: Slight preference for `s` over `tau` because that's what many Ethereum-related documents have been using. It also looks a bit better in code docs because it's a single letter. No strong opinion.

src/c_kzg_4844.c
```diff
@@ -3788,15 +3810,19 @@ C_KZG_RET verify_cell_kzg_proof(
         if (ret != C_KZG_OK) goto out;
     }
 
-    /* Calculate the input value */
+    /* Calculate the value x that identifies the coset
```
asn-d6, line 3813: Seems like you are wrapping the lines a bit too early here? I don't remember if we have precise rules, but we can certainly go up to 80 columns.

asn-d6 (changes requested):
Thanks! Added some opinionated review comments. Can merge when those are resolved.

src/c_kzg_4844.c
b-wagn: I did not want to use `s` because the KZGSettings are called `s` everywhere. But I can call it `s` if you think it makes sense. What do you think?

asn-d6: Reasonable point. Let's keep it as tau. We can change it later if we want.

asn-d6 (approved):
Thanks a lot! LGTM!

PR #443, opened 2024-06-24
feat(rust): Add ethereum kzg settings

rakita:
Adds `ethereum_kzg_settings` and `ethereum_kzg_settings_arc` functions behind a feature gate.

This would allow directly using eth trusted setup and consolidating revm/alloy/reth repos. Ref https://github.com/bluealloy/revm/issues/1553


jtraglia (approved):
Seems reasonable to me! Thanks 😃

In the past, I have been against using the trusted setup from this repo outside of testing. But this is a feature that clients (Reth + Lodestar) have requested. I think it's important that we listen to the clients, so I support this.

rakita:
> Seems reasonable to me! Thanks 😃
> 
> In the past, I have been against using the trusted setup from this repo outside of testing. But this is a feature that clients (Reth + Lodestar) have requested. I think it's important that we listen to the clients, so I support this.

It is hidden behind a rust feature flag, and a nice thing about that is that it is optional for a compilation. It is very useful to have it, especially `Arc` version.

asn-d6:
Looks good but would appreciate a code review by another member of the reth team (or lighthouse even). Can you arrange this @rakita, if possible?

@jtraglia Am I right to think that this makes https://github.com/ethereum/c-kzg-4844/pull/440 "easier" to port from DAS to main because now the file parsing happens inside the bindings and not in the clients?

jtraglia:
> Am I right to think that this makes https://github.com/ethereum/c-kzg-4844/pull/440 "easier" to port from DAS to main

Um maybe. I'll pull in these changes into the das branch, so merging back into main isn't really a concern.

rakita:
> Looks good but would appreciate a code review by another member of the reth team (or lighthouse even). Can you arrange this @rakita, if possible?
> 
> @jtraglia Am I right to think that this makes #440 "easier" to port from DAS to main because now the file parsing happens inside the bindings and not in the clients?

Make sense, will ask for a volunteer :)

rkrasiuk (approved):
gj

PR #449, opened 2024-07-02
DAS: Implement coset FFT and use it around the codebase

asn-d6:
This commit implements https://github.com/ethereum/consensus-specs/pull/3781 in c-kzg.

Some notes:
- Commit 7856f4dc is just comments and renames to bring the function closer to the spec (more work can be done)
- The last commit (a79cf966) is optional. We can keep it or discard it.
- I decided to discard `scale_poly()` and `unscale_poly()` and instead use a single function `shift_poly()` that can be reused in other situations as well.

The PR can be reviewed either commit-by-commit or in its entirety.

### DIFF
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -2830,8 +2830,8 @@ out:
  * @param[in]   n   Length of the arrays
  * @param[in]   s   The trusted setup
  *
- * @remark The coset shift factor is RECOVERY_SHIFT_FACTOR.
- * In this function we use its inverse to implement the IFFT.
+ * @remark The coset shift factor is RECOVERY_SHIFT_FACTOR. In this function we
+ *         use its inverse to implement the IFFT.
  */
 static C_KZG_RET coset_ifft_fr(
     fr_t *out, const fr_t *in, size_t n, const KZGSettings *s
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -2868,7 +2868,7 @@ out:
  * reconstructed original. Assumes that the inverse FFT of the original data
  * has the upper half of its values equal to zero.
  *
- * @param[out]  recovered   A preallocated array for recovered cells
+ * @param[out]  reconstructed_data_out   A preallocated array for recovered cells
  * @param[in]   cells       The cells that you have
  * @param[in]   s           The trusted setup
  *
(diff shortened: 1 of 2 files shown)


src/c_kzg_4844.c
```diff
@@ -2773,83 +2773,116 @@ static const fr_t INV_SCALE_FACTOR = {
...
+
+    shift_poly(in, n, &SCALE_FACTOR);
+
+    /* Compute the FFT */
+    ret = fft_fr(out, in, n, s);
+    if (ret != C_KZG_OK) goto out;
+
+out:
+    return ret;
+}
```
jtraglia, line 2816: Following my suggestion above, so we don't mutate `in`. This is what we would add:
```suggestion
out:
    /* Shift the input back so it's not modified */
    shift_poly(in, n, &INV_SCALE_FACTOR);
    return ret;
}
```

src/c_kzg_4844.c
```diff
@@ -2773,83 +2773,116 @@ static const fr_t INV_SCALE_FACTOR = {
...
-        blst_fr_mul(&factor_power, &factor_power, &INV_SCALE_FACTOR);
+        blst_fr_mul(&factor_power, &factor_power, shift_factor);
         blst_fr_mul(&p[i], &p[i], &factor_power);
     }
 }
 
-/**
- * Unscale a polynomial in place.
+/** Do an FFT over a coset of the roots of unity. The function mutates the
+ *  input array `in`
```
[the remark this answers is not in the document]
jtraglia: Also, I don't like that `in` is mutated. I understand it's because of the initial shift. I know it doesn't really matter, because `extended_evaluation_times_zero_coeffs` (the only thing that's impacted) isn't used again.

One simple solution would be to `shift_poly(in, n, &INV_SCALE_FACTOR)` at the bottom.

src/c_kzg_4844.c
```diff
@@ -2773,83 +2773,116 @@ static const fr_t INV_SCALE_FACTOR = {
...
-    for (size_t i = 1; i < len; i++) {
-        blst_fr_mul(&factor_power, &factor_power, &SCALE_FACTOR);
-        blst_fr_mul(&p[i], &p[i], &factor_power);
-    }
+static C_KZG_RET coset_fft_fr(
+    fr_t *out, fr_t *in, size_t n, const KZGSettings *s
+) {
+    C_KZG_RET ret;
+
+    shift_poly(in, n, &SCALE_FACTOR);
```
jtraglia, line 2808: Should we rename `SCALE_FACTOR` and `INV_SCALE_FACTOR` from "scale" to "shift"?

src/test_c_kzg_4844.c
```diff
@@ -1750,6 +1758,90 @@ static void test_expand_root_of_unity__fails_wrong_root_of_unity(void) {
...
+    }
+
+    /* Keep a backup of the coefficients becasue the coset_fft_fr() is gonna
+     * mutate them */
+    memcpy(poly_coeff_backup, poly_coeff, N * sizeof(fr_t));
+
+    /* Evaluate poly using coset FFT */
+    coset_fft_fr(poly_eval, poly_coeff, N, &s);
+
+    // check: result of coset FFT are really the evaluations over the coset
```
jtraglia, line 1820: This comment style isn't consistent. `/* */` style is used everywhere.

src/test_c_kzg_4844.c
```diff
@@ -2136,6 +2228,8 @@ static void teardown(void) {
 
 int main(void) {
     setup();
+    RUN(test_fft);
+    RUN(test_coset_fft);
     RUN(test_c_kzg_malloc__succeeds_size_greater_than_zero);
```
jtraglia, line 2233: Can we move the two calls down so the order matches the declarations?

jtraglia (commented):
This is looking pretty good! Here's my initial review.

src/c_kzg_4844.c
asn-d6: I think I prefer the memory allocation approach here: https://github.com/ethereum/c-kzg-4844/pull/449/commits/b815f1d558a123b26b725842320a16390f745d1e#diff-ed568929403e623a4b0c74780ccb41dffc9b594000d41d09b21a407fa2126576R2818-R2839
more than the shift-deshift approach.

This way `in` keeps its `const`ness as well.

src/c_kzg_4844.c
```diff
@@ -2870,83 +2903,101 @@ static C_KZG_RET recover_cells_impl(
...
     if (ret != C_KZG_OK) goto out;
 
     /* We use max_width here (not zero_poly_len) intentionally */
-    ret = fft_fr(eval_zero_poly, zero_poly, s->max_width, s);
+    ret = coset_fft_fr(zero_poly_over_coset, zero_poly_coeff, s->max_width, s);
     if (ret != C_KZG_OK) goto out;
 
     /* The result of the division is Q3 */
     for (size_t i = 0; i < s->max_width; i++) {
         fr_div(
```
kevaundray, line 2958: Orthogonal: Might need to benchmark the efficiency of this

kevaundray (approved):
Functionality looks good to me

src/c_kzg_4844.c
asn-d6: Reverted to mem alloc in 9751c918

asn-d6:
Addressed all reviewer comments in the latest commit (2ad06f60). Still haven't benchmarked the diff.

kevaundray:
> Addressed all reviewer comments in the latest commit ([2ad06f6](https://github.com/ethereum/c-kzg-4844/commit/2ad06f60391fe4671f043c3babb4fc108d296c1c)). Still haven't benchmarked the diff.

The benchmarking comment was with code that was untouched, so not a blocker for this

src/c_kzg_4844.c
```diff
@@ -2870,83 +2911,101 @@ static C_KZG_RET recover_cells_impl(
...
-            &eval_poly_with_zero[i], &eval_poly_with_zero[i], &eval_zero_poly[i]
+            &extended_evaluations_over_coset[i],
+            &extended_evaluations_over_coset[i],
+            &zero_poly_over_coset[i]
         );
     }
 
+    /* reconstructed_poly_over_coset is now extended_evaluations_over_coset */
```
jtraglia, line 2973: The `reconstructed_poly_over_coset` variable doesn't exist. Can we delete this comment?

asn-d6: This was an ambiguous comment. `reconstructed_poly_over_coset` is a variable in the spec. Will reword.

asn-d6: Addressed in 0edcea20

asn-d6: Ah, push failed because of the added commits. Re-pushed at d90c4b76!

PR #450, opened 2024-07-03
fix(rust): define NDEBUG when debug assertions are off

DaniPopes:
`cc-rs` does not define `NDEBUG` by default in release builds.


Rjected (approved):
lgtm

jtraglia (approved):
LGTM too. Thanks!

PR #452, opened 2024-07-12
Add support for EIP-7594

jtraglia:
This is a big PR which merges the das branch into main.

Client teams: when you have time, please review the changes to the bindings you use.

**When we merge this, please squash the commits.**

### Notable changes to the bindings

* Add `compute_cells_and_kzg_proofs`, `recover_cells_and_kzg_proofs`, `verify_cell_kzg_proof_batch` bindings.
* Extend the trusted setup with G1 points in monomial form; there are three groups of points now.
* Change `load_trusted_setup` (bytes version) to take flat array of bytes rather than list of point bytes.
  * This means the size params represent number of total bytes, not the number of points. 
* Add `precompute` parameter to functions which load the trusted setup.
  * This is an integer between 0 and 15. It's used to speed up proof computations.
  * Refer to [this HackMD document](https://notes.ethereum.org/@jtraglia/windowed_multiplications) to compare tradeoffs.
* ~~Make some of the previously public constants private.~~

### Notable changes to the core implementation

* Replace big stack allocations with heap allocations in `load_trusted_setup_file`.
* Modify `g1_lincomb_fast` to work with points at infinity.
  * Relatively simple change, but EIP-4844 functions use it.
* Replace uses of `CHECK` macro with normal conditions.
  * There was a risk of misusing this and returning without freeing allocations.

### Other notable changes

* ~~Replace old C LLVMFuzzer with new cargo-fuzz implementation in Rust.~~
  * ~~This will also differentially fuzz with [Constantine](https://github.com/mratsim/constantine) & [PeerDAS-KZG](https://github.com/crate-crypto/peerdas-kzg) where possible.~~


bindings/csharp/Ckzg.Bindings/Ckzg.cs
```diff
@@ -2,24 +2,28 @@ namespace Ckzg;
 
 public static partial class Ckzg
 {
-    public const int BytesPerFieldElement = 32;
-    public const int FieldElementsPerBlob = 4096;
-    public const int BytesPerBlob = BytesPerFieldElement * FieldElementsPerBlob;
+    public const int BytesPerBlob = 131072;
```
flcl42, line 5: Can be some other constants multiplication

bindings/csharp/Ckzg.Test/ReferenceTests.cs
```diff
@@ -354,4 +360,158 @@ public void TestVerifyBlobKzgProofBatch()
...
+    {
+        Matcher matcher = new();
+        matcher.AddIncludePatterns(new[] { "*/*/data.yaml" });
+
+        IEnumerable<string> testFiles = matcher.GetResultsInFullPath(_computeCellsAndKzgProofsTests);
+        Assert.That(testFiles.Count(), Is.GreaterThan(0));
+
+        foreach (string testFile in testFiles)
+        {
+            string yaml = File.ReadAllText(testFile);
```
flcl42, line 388: In general cases could be separated to a source passed to TestCaseSource attribute

bindings/csharp/Ckzg.Test/ReferenceTests.cs
```diff
@@ -354,4 +360,158 @@ public void TestVerifyBlobKzgProofBatch()
...
+
+            byte[] cells = new byte[CellsPerExtBlob * Ckzg.BytesPerCell];
+            byte[] proofs = new byte[CellsPerExtBlob * Ckzg.BytesPerProof];
+            byte[] blob = GetBytes(test.Input.Blob);
+
+            try
+            {
+                Ckzg.ComputeCellsAndKzgProofs(cells, proofs, blob, _ts);
+                Assert.That(test.Output, Is.Not.EqualTo(null));
+                byte[] expectedCells = GetFlatBytes(test.Output.ElementAt(0));
```
flcl42, line 400: Can the type of it be an array of arrays instead? Could you access elements just by [index]?

jtraglia: Hmm I'm not sure about this one. Let's look into this afterwards.

bindings/csharp/Ckzg.Test/ReferenceTests.cs
jtraglia: Good idea. I only did this for the new ones. Let's follow-up with a PR for the old tests later.

bindings/java/src/main/java/ethereum/ckzg4844/CKZG4844JNI.java
```diff
@@ -43,24 +43,36 @@ public static void loadNativeLibrary() {
     }
   }
 
-  /** Scalar field modulus of BLS12-381 */
-  public static final BigInteger BLS_MODULUS =
```
StefanBratanov, line 47: I prefer to keep those public in order to use those in projects which use the library

asn-d6: Hmm, protecting those constants was a conscious decision. It was an attempt to remove cryptographic details from the clients. In some cases cryptography code can be easy to mess up, and client devs might not be aware of the various cryptographic intricacies, so this change was done so that cryptography logic gets concentrated in the libraries.

How do you use these constants in Teku?

bindings/nim/kzg.nim
```diff
@@ -61,71 +68,88 @@ template verify(res: KZG_RET, ret: untyped): untyped =
...
+    G1Len = 48
+    G2Len = 96
 
   var
     s = newStringStream(input)
-    g1: array[FIELD_ELEMENTS_PER_BLOB, G1Data]
-    g2: array[NumG2, G2Data]
+    g1MonomialBytes: seq[byte] = newSeq[byte](NumG1 * G1Len)
+    g1LagrangeBytes: seq[byte] = newSeq[byte](NumG1 * G1Len)
+    g2MonomialBytes: seq[byte] = newSeq[byte](NumG2 * G2Len)
```
jangko, line 116: I think `NumG2 * G2Len` is small enough to put in stack. `newSeq` is slow.

bindings/nim/kzg.nim
```diff
@@ -61,71 +68,88 @@ template verify(res: KZG_RET, ret: untyped): untyped =
...
-      return err("invalid number of G2, expect $1, got $2" % [
+      return err("invalid number of G2 points, expect $1, got $2" % [
         $NumG2, $numG2
       ])
 
-    for i in 0 ..< FIELD_ELEMENTS_PER_BLOB:
-      g1[i] = hexToByteArray[G1Len](s.readLine())
+    for i in 0 ..< NumG1:
+      let p = hexToByteArray[G1Len](s.readLine())
+      for j in 0 ..< G1Len:
```
jangko, line 132:
```Nim
for j in 0 ..< G1Len:
   g1LagrangeBytes[i * G1Len + j] = p[j]
```
can be replaced with

```Nim
import stew/assign2

...
assign(g1LagrangeBytes.toOpenArray(i * G1Len, ((i + 1) * G1Len) - 1), p)
```

see comments in `stew/assign2.nim`

bindings/nim/kzg.nim
```diff
@@ -61,71 +68,88 @@ template verify(res: KZG_RET, ret: untyped): untyped =
...
+        g1LagrangeBytes[i * G1Len + j] = p[j]
 
     for i in 0 ..< NumG2:
-      g2[i] = hexToByteArray[G2Len](s.readLine())
+      let p = hexToByteArray[G2Len](s.readLine())
+      for j in 0 ..< G2Len:
+        g2MonomialBytes[i * G2Len + j] = p[j]
```
jangko, line 138: ditto

bindings/nim/kzg.nim
```diff
@@ -61,71 +68,88 @@ template verify(res: KZG_RET, ret: untyped): untyped =
...
+
+    for i in 0 ..< NumG1:
+      let p = hexToByteArray[G1Len](s.readLine())
+      for j in 0 ..< G1Len:
+        g1MonomialBytes[i * G1Len + j] = p[j]
```
jangko, line 143: ditto

bindings/java/src/main/java/ethereum/ckzg4844/CKZG4844JNI.java
StefanBratanov: We have a utility class which flattens a list of commitments/proofs for example and we calculate the expected size based on these constants e.g. https://github.com/Consensys/teku/blob/master/infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/CKZG4844Utils.java

We also use in tests https://github.com/Consensys/teku/blob/master/infrastructure/kzg/src/test/java/tech/pegasys/teku/kzg/CKZG4844Test.java

I suppose not an issue to replicate those in Teku internally in the module, but would lead to duplication.

bindings/nim/kzg.nim
jtraglia: Makes sense. For consistency, I did it for all. It's ~0.5 MiB in total. Should be fine.

bindings/java/src/main/java/ethereum/ckzg4844/CKZG4844JNI.java
jtraglia: Stefan, we've decided to revert this change. So these constants are public again for all bindings. With the exception of the `BYTES_PER_G*` constants because `load_trusted_setup` takes a flat array of bytes now.

PR #454, opened 2024-07-18
g1_lincomb_fast(): Simplify point at infinity weed out (`das` branch)

asn-d6:


b-wagn (approved):
Makes code more readable 👍

PR #456, opened 2024-07-18
Increase column limit to 100 for C code

jtraglia:
This PR changes the column limit from 80 to 100. George and I had discussed how annoying the 80 char limit is. We agreed that to raise it 100, a value that modern languages like Rust and Go adhere to.


asn-d6 (approved):
LGTM!

Not merging so as to not create merge conflicts with the other PRs.

asn-d6 (approved):
LGTM! Thanks!

PR #459, opened 2024-07-19
Update recovery code to match specs

kevaundray:
This updates the computation of the zero polynomial to match the specs.

### DIFF
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -2373,11 +2373,6 @@ static C_KZG_RET ifft_fr(
 // Vanishing poly
 ///////////////////////////////////////////////////////////////////////////////
 
-typedef struct {
-    fr_t *coeffs;
-    size_t length;
-} poly_t;
-
 /**
  * Calculates the minimal polynomial that evaluates to zero for each root.
  *
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -2497,7 +2497,8 @@ static C_KZG_RET vanishing_polynomial_for_missing_cells(
                                 ...,
                                 \omega^i * \gamma^{FIELD_ELEMENTS_PER_CELL-1}
                                 }
-        where \gamma is a primitive `FIELD_ELEMENTS_PER_EXT_BLOB`-th root of unity.
+        where \gamma is a primitive `FIELD_ELEMENTS_PER_EXT_BLOB`-th root of
+       unity.
 
         This is done by shifting the degree of all coefficients in
         `short_vanishing_poly` up by `FIELD_ELEMENTS_PER_CELL` amount.
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -2429,9 +2429,9 @@ static C_KZG_RET compute_vanishing_polynomial_from_roots(
  * spaced chosen roots of unity in the domain of size `FIELD_ELEMENTS_PER_BLOB`.
  *
  * The roots of unity are chosen based on the missing cell indices. If the i'th
- * cell is missing, then the i'th root of unity will be zero on the polynomial
- * computed, along with every `CELLS_PER_EXT_BLOB` spaced root of unity in the
- * domain.
+ * cell is missing, then the i'th root of unity from `expanded_roots_of_unity`
+ * will be zero on the polynomial computed, along with every
+ * `CELLS_PER_EXT_BLOB` spaced root of unity in the domain.
  *
  * @param[in,out]   vanishing_poly          The vanishing polynomial
  * @param[in,out]   vanishing_poly_len      The length of vanishing_poly
--- a/src/c_kzg_4844.c
+++ b/src/c_kzg_4844.c
@@ -2375,7 +2375,6 @@ static C_KZG_RET ifft_fr(
 /**
  * Calculates the minimal polynomial that evaluates to zero for each root.
  *
- *
  * Uses straightforward long multiplication to calculate the product of
  * `(x - r_i)` where `r_i` is the i'th root. This results in a poly of degree
  * roots_len.
@@ -2421,8 +2420,9 @@ static C_KZG_RET compute_vanishing_polynomial_from_roots(
     return C_KZG_OK;
 }
 
-/* Computes the minimal polynomial that evaluates to zero at equally
- * spaced chosen roots of unity in the domain of size `FIELD_ELEMENTS_PER_BLOB`.
+/**
+ * Computes the minimal polynomial that evaluates to zero at equally spaced
+ * chosen roots of unity in the domain of size `FIELD_ELEMENTS_PER_BLOB`.
  *
  * The roots of unity are chosen based on the missing cell indices. If the i'th
  * cell is missing, then the i'th root of unity from `expanded_roots_of_unity`
@@ -2465,23 +2465,23 @@ static C_KZG_RET vanishing_polynomial_for_missing_cells(
     ret = new_fr_array(&short_vanishing_poly, (len_missing_cells + 1));
     if (ret != C_KZG_OK) goto out;
 
-    /* Check if max_width is divisible by CELLS_PER_EXT_BLOB*/
+    /* Check if max_width is divisible by CELLS_PER_EXT_BLOB */
     assert(s->max_width % CELLS_PER_EXT_BLOB == 0);
 
     /*
-        For each missing cell index, choose the corresponding root of unity
-        from the subgroup of size `CELLS_PER_EXT_BLOB`.
-
-        In other words, if the missing index is `i`, then we add \omega^i
-        to the roots array, where \omega is a primitive `CELLS_PER_EXT_BLOB`
-        root of unity.
-    */
+     * For each missing cell index, choose the corresponding root of unity from
+     * the subgroup of size`CELLS_PER_EXT_BLOB`.
+     *
+     * In other words, if the missing index is `i`, then we add \omega^i to the
+     * roots array, where \omega is a primitive `CELLS_PER_EXT_BLOB` root of
+     * unity.
+     */
     size_t stride = s->max_width / CELLS_PER_EXT_BLOB;
     for (size_t i = 0; i < len_missing_cells; i++) {
         roots[i] = s->expanded_roots_of_unity[missing_cell_indices[i] * stride];
     }
 
-    /* Compute the polynomial that evaluates to zero on the roots*/
+    /* Compute the polynomial that evaluates to zero on the roots */
     ret = compute_vanishing_polynomial_from_roots(
         short_vanishing_poly,
         &short_vanishing_poly_len,
@@ -2491,18 +2491,22 @@ static C_KZG_RET vanishing_polynomial_for_missing_cells(
     if (ret != C_KZG_OK) goto out;
 
     /*
-        For each root \omega^i in `short_vanishing_poly`, we compute a
-        polynomial that has roots at H =   { \omega^i * \gamma^0, \omega^i *
-        \gamma^1,
-                                ...,
-                                \omega^i * \gamma^{FIELD_ELEMENTS_PER_CELL-1}
-                                }
-        where \gamma is a primitive `FIELD_ELEMENTS_PER_EXT_BLOB`-th root of
-       unity.
-
-        This is done by shifting the degree of all coefficients in
-        `short_vanishing_poly` up by `FIELD_ELEMENTS_PER_CELL` amount.
-    */
+     * For each root \omega^i in `short_vanishing_poly`, we compute a
+     * polynomial that has roots at
+     *
+     *  H = {
+     *      \omega^i * \gamma^0,
+     *      \omega^i * \gamma^1,
+     *      ...,
+     *      \omega^i * \gamma^{FIELD_ELEMENTS_PER_CELL-1}
+     *  }
+     *
+     * where \gamma is a primitive `FIELD_ELEMENTS_PER_EXT_BLOB`-th root of
+     * unity.
+     *
+     * This is done by shifting the degree of all coefficients in
+     * `short_vanishing_poly` up by `FIELD_ELEMENTS_PER_CELL` amount.
+     */
     for (size_t i = 0; i < short_vanishing_poly_len; i++) {
         vanishing_poly[i * FIELD_ELEMENTS_PER_CELL] = short_vanishing_poly[i];
     }
@@ -2627,9 +2631,9 @@ out:
 /**
  * Helper function to check if a uint64 value is in an array.
  *
- * @param[in]   arr                      The array
- * @param[in]   arr_size                 The size of the array
- * @param[in]   value                    The value we want to search
+ * @param[in]   arr         The array
+ * @param[in]   arr_size    The size of the array
+ * @param[in]   value       The value we want to search
  *
  * @return True if the value is in the array, otherwise false.
  */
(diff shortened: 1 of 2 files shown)


src/c_kzg_4844.c
```diff
@@ -2378,231 +2379,84 @@ typedef struct {
...
+ * @param[in]       s         The trusted setup
  *
- * @remark This does not work if all indices are missing.
- * @remark Unused coefficients are set to zero.
+ * @remark This does not work if all cells are missing.
+ * @remark `missing_cell_indices` are assumed to be less than
+ * `CELLS_PER_EXT_BLOB`.
  */
-static C_KZG_RET zero_polynomial_via_multiplication(
+static C_KZG_RET zero_polynomial_periodic_roots(
```
b-wagn, line 2446: *Periodic* is not a term that we have used so far. How about `vanishing_polynomial_for_missing_cells` or something similar?

src/c_kzg_4844.c
```diff
@@ -2378,231 +2379,84 @@ typedef struct {
...
+ * cell is missing, then the i'th root of unity will be zero on the polynomial
+ * computed, along with every `CELLS_PER_EXT_BLOB` spaced root of unity in the
+ * domain.
  *
- * @param[out]  zero_poly       The zero polynomial
- * @param[out]  zero_poly_len   The zero polynomial length
- * @param[in]   missing_indices The indices of the missing coefficients
- * @param[in]   len_missing     The number of missing indices
- * @param[in]   s               The trusted setup
+ * @param[in,out]   zero_poly      The zero polynomial for cell indices
```
b-wagn, line 2436: See my comment about *vanishing* vs *zero* above.

src/c_kzg_4844.c
```diff
@@ -2378,231 +2379,84 @@ typedef struct {
...
- * Calculate the minimal polynomial that evaluates to zero for the powers of
- * roots of unity that correspond to missing indices.
+/* Computes the minimal polynomial that evaluates to zero at equally
+ * spaced chosen roots of unity in the domain of size `FIELD_ELEMENTS_PER_BLOB`.
  *
- * This is done simply by multiplying together `(x - r^i)` for all the `i` that
- * are missing indices, using a combination of direct multiplication
- * (#do_zero_poly_mul_partial) and iterated multiplication via convolution
- * (#reduce_partials).
+ * The roots of unity are chosen based on the missing cell indices. If the i'th
```
b-wagn, line 2431: I think the doc comment (and comments in the function itself) here is already pretty good. 
One thing I would add is where to get the `i'th root of unity` from, as this heavily depends on the order of the roots. 

If I understand correctly, it would be `s->expanded_roots_of_unity[i]`?

src/c_kzg_4844.c
kevaundray, line 2446: unsure about the postfix "periodic_roots" -- happy to change this to anything else that is clearer upon first read

src/c_kzg_4844.c
```diff
@@ -2611,102 +2465,68 @@ static C_KZG_RET zero_polynomial_via_multiplication(
...
-    }
+    /*
+        For each root \omega^i in `short_zero_poly`, we compute a polynomial
+        that has roots at H =   {
+                                \omega^i * \gamma^0,
+                                \omega^i * \gamma^1,
+                                ...,
+                                \omega^i * \gamma^{FIELD_ELEMENTS_PER_CELL-1}
+                                }
+        where \gamma is primitive `FIELD_ELEMENTS_PER_EXT_BLOB` generator.
```
b-wagn, line 2516: I don't fully get what `\gamma is primitive FIELD_ELEMENTS_PER_EXT_BLOB generator` is supposed to mean.

src/c_kzg_4844.c
```diff
@@ -2820,12 +2640,32 @@ static C_KZG_RET coset_ifft_fr(
...
+    return false;
+}
+
 /**
  * Given a dataset with up to half the entries missing, return the
  * reconstructed original. Assumes that the inverse FFT of the original data
  * has the upper half of its values equal to zero.
  *
  * @param[out]  reconstructed_data_out   Preallocated array for recovered cells
+ * @param[in]   cell_indices             The cell indices
```
b-wagn, line 2667:
```suggestion
 * @param[in]   cell_indices             The cell indices you have
```

src/c_kzg_4844.c
```diff
@@ -2378,231 +2379,84 @@ typedef struct {
...
-    fr_t *work = NULL;
-    fr_t *scratch = NULL;
-    poly_t *partials = NULL;
+    fr_t *roots = NULL;
+    fr_t *short_zero_poly = NULL;
+    size_t short_zero_poly_len = 0;
 
     /* If nothing is missing, return all zeros */
-    if (len_missing == 0) {
+    if (len_missing_cells == 0) {
```
b-wagn, line 2459: I am a bit puzzled about this special case (which was already around before this PR): the vanishing polynomial of empty set should be constant FR_ONE, right? Because it is defined as the polynomial that vanishes exactly on the set, or as the empty product over terms `X-r_i`.  

If it is a special case that we really need for the recovery, then we should document that. Are we expecting users to call recovery even no cell is missing? 

Otherwise, I would suggest we explicitly disallow this case, and return `C_KZG_BADARGS`, as you do in `zero_polynomial_from_roots`.

src/c_kzg_4844.c
```diff
@@ -2875,18 +2721,24 @@ static C_KZG_RET recover_cells_impl(
 
     /* Identify missing cells */
```
b-wagn, line 2722: Your approach for populating `missing_cell_indices` is very readable, that's good 😄. 

But I wonder if the following is more efficient for large `num_cells` (and still readable): 
1. Create a bool array of length `CELLS_PER_EXT_BLOB` that contains `false` everwhere. It keeps track of which cells we have.
2. Iterate over `cell_indices` and mark the cells we have in the bool array as `true`.
3. Iterate over the bool array and add every cell for which we have `false` to `missing_cell_indices`.

If `num_cells` is almost `CELLS_PER_EXT_BLOB` (i.e., only a few cells missing), then your approach scales *quadratically* in `CELLS_PER_EXT_BLOB`, whereas what I suggest would scale *linearly*, something like `3 * CELLS_PER_EXT_BLOB`.  

Am I missing something? Did you test this? Or is it just that you did not want to use the long bool array? 😅

b-wagn (commented):
Thanks for the PR, this is great!
Unifying the code with the spec is always good.
In this particular case, it also makes it much more readable and removes more complex functions.

I left a few comments, mostly about how to make it even more readable, how to deal with edge cases, and possibly one improvement.

Let me know what you think :)

src/test_c_kzg_4844.c
```diff
@@ -1970,6 +1971,79 @@ static void test_recover_cells_and_kzg_proofs__succeeds_random_blob(void) {
     }
 }
 
+static void test_zero_polynomial_from_roots(void) {
```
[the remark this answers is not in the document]
kevaundray: sounds good -- I guess the `assert` in the setup code is the exception here

src/c_kzg_4844.c
kevaundray: It might be the phrasing: It means that \gamma is able to generate the subgroup of size FIELD_ELEMENTS_PER_EXT_BLOB where each element is a root of unity

src/c_kzg_4844.c
kevaundray: returning `C_KZG_BADARGS` would be the cleanest way I think -- will change to that and add a remark that the user should explicitly handle that case.

src/c_kzg_4844.c
kevaundray: Yep I can modify that to be more explicit, and yep it is `expanded_roots_of_unity` (non-bit-reversed) -- note that the code has been modularized so that all of the vanishing polynomial calculations/functions do not depend on any bit reversal permutation

kevaundray: I've modified it to include `expanded_roots_of_unity`, though feel free to push a commit to modify it further

src/c_kzg_4844.c
kevaundray: Mentioned on telegram, will post here for prosperity -- for small lists, vectors have good cache utilization which is why I intitially went for it. I ran the below benchmark to check that I did not make a mistake and got ~0.04 milliseconds -- the bool array algorithm seems also quite simple, so I don't mind using that instead.


```c
#include <stdio.h>
#include <stdlib.h>
#include <stdbool.h>
#include <stdint.h>
#include <time.h>

#define ARRAY_SIZE 128

bool is_in_array(const uint64_t *arr, size_t arr_size, uint64_t value) {
    for (size_t i = 0; i < arr_size; i++) {
        if (arr[i] == value) {
            return true;
        }
    }
    return false;
}

int main() {
    uint64_t array[ARRAY_SIZE];
    uint64_t missing[ARRAY_SIZE];
    size_t missing_count = 0;

    // Initialize the array with some random values
    for (size_t i = 0; i < ARRAY_SIZE; i++) {
        array[i] = rand() % ARRAY_SIZE;
    }

    // Time the for loop
    clock_t start = clock();

    // Check for missing numbers, but make sure that we
    // search the whole array each time by using numbers greater
    // than ARRAY_SIZE.
    for (uint64_t i = ARRAY_SIZE + 1; i <= 2 * ARRAY_SIZE; i++) {
        if (!is_in_array(array, ARRAY_SIZE, i)) {
            missing[missing_count++] = i;
        }
    }

    clock_t end = clock();
    double cpu_time_used = ((double) (end - start)) / CLOCKS_PER_SEC;

    // Print the execution time
    printf("Execution time: %f seconds\n", cpu_time_used);

    return 0;
}
```

src/c_kzg_4844.c
```diff
@@ -2378,231 +2379,84 @@ typedef struct {
 } poly_t;
```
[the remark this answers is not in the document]
kevaundray: Hmm I'm not sure, I'll make a PR to remove it

src/test_c_kzg_4844.c
```diff
@@ -1970,6 +1971,79 @@ static void test_recover_cells_and_kzg_proofs__succeeds_random_blob(void) {
...
+
+    ret = zero_polynomial_periodic_roots(
+        zero_poly, &zero_poly_len, missing_cell_indices, len_missing_cells, &s
+    );
+
+    // Check return status
+    assert(ret == C_KZG_OK);
+
+    // Check polynomial length
+    assert(zero_poly_len == MAX_WIDTH);
```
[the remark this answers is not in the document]
kevaundray: Yep, for prosperity, this is just checking the length -- I think that now that the edge case has been removed, the zero_poly_len can be removed too

src/test_c_kzg_4844.c
```diff
@@ -1970,6 +1971,79 @@ static void test_recover_cells_and_kzg_proofs__succeeds_random_blob(void) {
...
+
+    // Compute FFT of zero_poly
+    fft_fr(fft_result, zero_poly, MAX_WIDTH, &s);
+
+    // Check FFT results
```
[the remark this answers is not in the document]
kevaundray: I added more explanation on how to derive these roots using `expanded_roots_of_unity` in it -- feel free to move it to where you think it is appropriate

src/c_kzg_4844.c
```diff
@@ -2378,231 +2379,84 @@ typedef struct {
...
- * @param[out]  out             Polynomial with len_out fields allocated
- * @param[in]   len_out         Domain evaluation length, a power of two
- * @param[in]   scratch         Scratch space, at least 3x len_out
- * @param[in]   len_scratch     Scratch size, at least 3x len_out
- * @param[in]   partials        Array of polys to be multiplied together
- * @param[in]   partial_count   The number of polys to be multiplied together
- * @param[in]   s               The trusted setup
+ * @param[in,out]   poly      The zero polynomial for roots
+ * @param[in,out]   poly_len  The length of poly
+ * @param[in]       roots     The array of roots
```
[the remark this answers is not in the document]
kevaundray: I added a remark about them not needing to be roots of unity, but the roots of polynomial -- domain is also a bit overloaded in the codebase already (I have a preference for root, but wouldn't be sad if it changed to domain)

src/c_kzg_4844.c
```diff
@@ -2369,344 +2370,150 @@ static C_KZG_RET ifft_fr(
...
+    ret = new_fr_array(&roots, len_missing_cells);
+    if (ret != C_KZG_OK) goto out;
+    ret = new_fr_array(&short_vanishing_poly, (len_missing_cells + 1));
+    if (ret != C_KZG_OK) goto out;
+
+    /* Check if max_width is divisible by CELLS_PER_EXT_BLOB*/
+    if (s->max_width % CELLS_PER_EXT_BLOB != 0) {
         ret = C_KZG_BADARGS;
         goto out;
     }
```
jtraglia, line 2473: I feel that this particular check should be an assert, as it should never happen.

src/test_c_kzg_4844.c
```diff
@@ -2,6 +2,7 @@
  * This file contains unit tests for C-KZG-4844.
  */
 #include "c_kzg_4844.c"
+#include "c_kzg_4844.h"
```
jtraglia, line 5: Why was this added?

src/c_kzg_4844.c
```diff
@@ -2820,12 +2627,32 @@ static C_KZG_RET coset_ifft_fr(
...
+/**
+ * Helper function to check if a uint64 value is in an array.
+ *
+ * @param[in]   arr                      The array
+ * @param[in]   arr_size                 The size of the array
+ * @param[in]   value                    The value we want to search
+ *
+ * @return True if the value is in the array, otherwise false.
+ */
+bool is_in_array(const uint64_t *arr, size_t arr_size, uint64_t value) {
```
jtraglia, line 2639:
```suggestion
static bool is_in_array(const uint64_t *arr, size_t arr_size, uint64_t value) {
```

src/c_kzg_4844.c
```diff
@@ -2369,344 +2369,147 @@ static C_KZG_RET ifft_fr(
...
-        );
-        if (ret != C_KZG_OK) goto out;
+    /*
+        For each root \omega^i in `short_vanishing_poly`, we compute a
+        polynomial that has roots at H =   { \omega^i * \gamma^0, \omega^i *
+        \gamma^1,
+                                ...,
+                                \omega^i * \gamma^{FIELD_ELEMENTS_PER_CELL-1}
+                                }
+        where \gamma is primitive `FIELD_ELEMENTS_PER_EXT_BLOB` generator.
```
b-wagn, line 2500:
```suggestion
        where \gamma is a primitive `FIELD_ELEMENTS_PER_EXT_BLOB`-th root of unity.
```

src/test_c_kzg_4844.c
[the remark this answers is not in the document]
b-wagn: This is a very good explanation, thank you!

src/c_kzg_4844.c
b-wagn: ok, cool. Thanks for the benchmark.
Then I think we can just keep the `is_in_array`-code. :)

src/c_kzg_4844.c
```diff
@@ -2369,344 +2369,151 @@ static C_KZG_RET ifft_fr(
...
  *
- * @remark This will pad the polynomials, perform FFTs, point-wise multiply the
- *         results together, and apply an inverse FFT to the result.
+ * @remark When all of the cells are missing, this algorithm has
+ * an edge case. We return C_KZG_BADARGS in that case.
+ * @remark When none of the cells are missing, recovery
+ * is trivial. We expect the caller to handle this case,
+ * and return C_KZG_BADARGS if not.
+ * @remark `missing_cell_indices` are assumed to be less than
+ * `CELLS_PER_EXT_BLOB`.
```
b-wagn, line 2443:
```suggestion
 * `CELLS_PER_EXT_BLOB` and at least one.
```

b-wagn (approved):
LGTM 👍

src/c_kzg_4844.c
jtraglia: The previous remark should be good enough.

jtraglia (approved):
Thanks Kev! This is much better 😃

Issue #413, opened 2024-03-29
[rust] Add compute_* verify_* functions for KzgSettings


mattsse:
currently, all `KzgProof` functions take `KzgSettings` as argument:

https://github.com/ethereum/c-kzg-4844/blob/1ef67c17bf659547749750e6ebaaf09563e0149d/bindings/rust/src/bindings/mod.rs#L321-L325

for convenience, all of those could also be implemented for `KzgSettings` directly that delegate with `&self`

Since these are new functions, this would make it possible to change the API and accept fixed arrays directly instead of the c-kzg Bytes wrapper types, which would be even more convenient.

jtraglia:
I think this is a good idea. @asn-d6 @pawanjay176 what do y'all think about this?

mattsse:
perhaps a bit more context re the last part, building the c-kzg crate can be a problem on certain platforms, hence in alloy the kzg types that are used for rpc are using alloy-primitives, which are also just array wrappers:

https://github.com/alloy-rs/alloy/blob/1fc24a7545ad51f26b09747a3c7c2b204a56c869/crates/consensus/src/transaction/eip4844.rs#L963-L970

so we can use the type without needing c-kzg, they have the same array repr and could be casted to the c-kzg, but ideally we can move this step behind the `KzgSettings` API so that we can verify types that are `AsRef<[]>` for example.

asn-d6:
> currently, all `KzgProof` functions take `KzgSettings` as argument:
> 
> https://github.com/ethereum/c-kzg-4844/blob/1ef67c17bf659547749750e6ebaaf09563e0149d/bindings/rust/src/bindings/mod.rs#L321-L325
> 
> for convenience, all of those could also be implemented for `KzgSettings` directly that delegate with `&self`
> 

No strong opinion about this change. 

Gotta say that it seems a bit arbitrary to me to consider `compute_kzg_proof()` a method of `KZGSettings`. Because in a similar way, I could think of it as a method of `Blob`.

That said, if this makes people happy, I don't have a strong opinion about it. I think Justin also likes it because it will inadvertently also group all the public functions together.

> Since these are new functions, this would make it possible to change the API and accept fixed arrays directly instead of the c-kzg Bytes wrapper types, which would be even more convenient.

Hm, this seems like an unrelated and more complicated change. I would make a separate issue for it.

[...]

PR #461, opened 2024-07-19
chore(rust): move Rust bindings to KZGSettings

jtraglia:
This moves the public functions to KZGSettings. And we no longer need to pass `&kzg_settings` as an argument.

Associated with #413.

### DIFF
--- a/bindings/rust/benches/kzg_benches.rs
+++ b/bindings/rust/benches/kzg_benches.rs
@@ -36,7 +36,8 @@ pub fn criterion_benchmark(c: &mut Criterion) {
     let commitments: Vec<Bytes48> = blobs
         .iter()
         .map(|blob| {
-            KzgCommitment::blob_to_kzg_commitment(blob, &kzg_settings)
+            kzg_settings
+                .blob_to_kzg_commitment(blob)
                 .unwrap()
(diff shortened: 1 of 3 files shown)


asn-d6:
Hmm. I think this is fine, but I have no strong feelings on this PR in either direction.

I will let @mattsse decide if it works better for their codebase.

IIUC this is a non-backwards compatible change and Lighthouse will also need to adjust their codebase.

mattsse (commented):
this should work but I believe these functions could be kept on Kzgcommitment for backwards compat and then the new fns on kzgsettings would call them?

jtraglia:
> I believe these functions could be kept on Kzgcommitment for backwards compat and then the new fns on kzgsettings would call them?

Hmm that would work too, but it feels a little confusing. I'm okay with breaking backwards compatibility this time. It's a good time to do this, as we're going to make a v2 release soon-ish. I like having all the public functions accessible in the same way.

mattsse:
changes should be minimal, @Rjected for vis

Rjected:
This API makes sense to me, fine with updating our code when it breaks

asn-d6 (approved):
LGTM! Thanks!

PR #462, opened 2024-07-19
fix: poly_lagrange_to_monomial

kevaundray:
This function had the parameters in the wrong order. It did not break because it was internally computing the monomial form but naming it the lagrange form. I've also modified the description, the previous wording made it seem like you needed to do bit reversal to convert a polynomial from lagrange to monomial form.


asn-d6 (approved):
LGTM! Thanks!

(Pushed a small commit as well)

jtraglia (approved):
Oops! Looks great. Thanks for fixing this.

PR #463, opened 2024-07-19
fix: remove extraneous memset

kevaundray:
calloc should zero out the memory when you allocate so the memset here would be un-needed


asn-d6 (approved):
LGTM! Thanks!

jtraglia (commented):
LGTM 👍

PR #464, opened 2024-07-26
Split core implementation into multiple files

asn-d6:
Splits codebase into: common/4844/7594 files


src/common.c
```diff
@@ -363,15 +363,14 @@ void compute_powers(fr_t *out, const fr_t *x, uint64_t n) {
...
  * Deserialize a blob (array of bytes) into a polynomial (array of field elements).
  *
- * @param[out] p    The output polynomial (array of field elements)
- * @param[in]  blob The blob (an array of bytes)
+ * @param[out]  p           The output polynomial (array of field elements)
+ * @param[in]   blob        The blob (an array of bytes)
+ * @param[in]   num_fields  The number of field elements in the blob
  */
-C_KZG_RET blob_to_polynomial(Polynomial *p, const Blob *blob) {
+C_KZG_RET blob_to_polynomial(fr_t *p, const uint8_t *blob, size_t num_fields) {
```
asn-d6: Hm. Why was this function made generic? Seems like we always use it with `n=FIELD_ELEMENTS_PER_BLOB` anyway.

BTW, IMO `num_fields` as a var name can be improved. "fields" is not a common way to call "field elements". I think `num_fields -> n` would make things better, but perhap we can just remove this change altogether if we don't need the function to be generic.

src/common.c
```diff
@@ -653,16 +653,16 @@ static void fft_g1_fast(
  * @param[in]   s   The trusted setup
  *
  * @remark The array lengths must be a power of two.
- * @remark Use ifft_g1() for inverse transformation.
+ * @remark Use g1_ifft() for inverse transformation.
  */
-C_KZG_RET fft_g1(g1_t *out, const g1_t *in, size_t n, const KZGSettings *s) {
+C_KZG_RET g1_fft(g1_t *out, const g1_t *in, size_t n, const KZGSettings *s) {
```
asn-d6, line 658: Why are we renaming these functions? The spec also names them `fft_field` and `coset_fft_field` (the type is in the suffix).

asn-d6 (commented):
Thanks for the changes! Great work.

Left some comments. I can address them myself after I hear your thoughts.

BTW 075979cd3f942c64002e9c43d9a66a7161bda5b2 is hard to review because it moves and restructures code (e.g. breaks into lines) at the same time. In the future, it would be great if such massive refactors are split into two commits: one that moves, and another that restructures.

src/common.h
```diff
@@ -0,0 +1,257 @@
...
+
+/*
+ * Point Operations:
+ */
+void g1_sub(g1_t *out, const g1_t *a, const g1_t *b);
+void g1_mul(g1_t *out, const g1_t *a, const fr_t *b);
+bool pairings_verify(const g1_t *a1, const g2_t *a2, const g1_t *b1, const g2_t *b2);
+void g1_lincomb_naive(g1_t *out, const g1_t *p, const fr_t *coeffs, uint64_t len);
+C_KZG_RET g1_lincomb_fast(g1_t *out, const g1_t *p, const fr_t *coeffs, size_t len);
+C_KZG_RET g1_fft(g1_t *out, const g1_t *in, size_t n, const KZGSettings *s);
```
b-wagn, line 252: Not sure if `g1_fft` (and `g1_ifft`) should be in common. We only use it in eip7594, right? 

On the other hand, it is a somewhat general operation that we may want to use elsewhere at some point.

src/eip4844.h
```diff
@@ -0,0 +1,114 @@
...
+/** A basic blob data. */
+typedef struct {
+    uint8_t bytes[BYTES_PER_BLOB];
+} Blob;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+// Public Functions
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+C_KZG_RET load_trusted_setup(
```
b-wagn, line 57: The trusted setup functions (`load_trusted_setup`, `load_trusted_setup_file`, `free_trusted_setup`) are also in setup.h. 
Accident or on purpose?

bindings/rust/build.rs
```diff
@@ -155,9 +156,6 @@ fn replace_ckzg_ret_repr(mut bindings: String) -> String {
         .rfind(repr_to_replace)
         .expect("Could not find repr to replace in bindings");
 
-    // Sanity check that it's an attribute of `C_KZG_RET` and not another type.
-    assert!(repr_start > bindings[..ckzg_ret].rfind('}').unwrap());
```
jtraglia, line 159: Just to note, removed this because it doesn't work if `C_KZG_RET` is the first type.

src/common.c
jtraglia: In C, the type is the prefix. For example:

telegram-cloud-photo-size-1-4994634266591735274-y

src/common.c
jtraglia: This function is defined in common, but references Blob and Polynomial. Blob should be defined in eip4844.h and Polynomial shouldn't be exposed. This is the best way I thought to handle this.

asn-d6: Hmm, I see.
I think Blob and FIELD_ELEMENTS_PER_BLOB should be moved to common, since they are used both by 4844 and 7594. I did this in e4d6c1cc.

Then in 551b7c53, I restored `blob_to_polynomial()` closer to what it once was. To minimize the diffs of audited 4844 code.

asn-d6:
I did a bit of `#include` cleanup in 82deee52. I think we should include header files from the source files when needed, and only include stuff in the header files if it's needed by the header file itself. It seems like this is [the general (but not mandatory) convention](https://stackoverflow.com/questions/3943352/where-to-put-include-statements-header-or-source).

I think this way it's cleaner to see what each source file needs, and also cleaner for the compiler to understand what each header file actually needs.

As a sidenote, see how 82deee529856466b78fdf8cff2fdd80ca232b0c7 had to modify the python C wrapper. The wrapper file was using boolean types even tho it was not including `stbool`. The compilation worked because `<stdbool.h>` was silently (and needlessly) included by an include of `ckzg.h`.

asn-d6:
I think there is an issue with the Makefile now. For example, run `make`, then do a modification to `common.c` and run `make` again. The Makefile won't notice that `common.c` was modified, and will just move on with running the tests.

jtraglia (approved):
LGTM 👍 let's plan to do more splits in the future.

PR #466, opened 2024-08-01
Move FFT functions to fft.{c,h} files

jtraglia:
Move the `*_{i,}fft` functions (except for `coset_{i,}fft`) to a new `fft.c` file. Exposed via `fft.h`.


src/ckzg.c
```diff
@@ -17,4 +17,5 @@
 #include "common.c"
 #include "eip4844.c"
 #include "eip7594.c"
+#include "fft.c"
```
asn-d6, line 20: Hm. Is this needed? Do we need to expose `fft.c`? Does it hold any public methods?

src/tests.c
```diff
@@ -1,12 +1,8 @@
 /*
  * This file contains unit tests for C-KZG-4844.
  */
-#include "common.c"
+#include "ckzg.c"
```
asn-d6, line 4: IMO it's better to include the .c files that actually have the functions, instead of the mother file. I think it makes dependencies more explicit. No strong opinion.

src/Makefile
```diff
@@ -40,7 +40,8 @@ BLST_BUILDSCRIPT_FLAGS = -D__BLST_PORTABLE__
 LIBS = $(BLST_LIBRARY)
 
 # List of individual implementation files.
-IMPL_FILES = common.c eip4844.c eip7594.c setup.c
+IMPL_FILES = common.c debug.c eip4844.c eip7594.c fft.c setup.c
```
asn-d6, line 43: I guess a question for this PR is whether we want to introduce some directory structure like:
- `src/ckzg.c`
- `src/setup.c`
- `src/eip4844/eip4844.c`
- `src/eip7594/eip7594.c`
- `src/eip7594/fft.c`

So that we don't dump all the files on the root dir. I think I like the structure I suggest above, but no strong opinion. AFAIU it should not complicate the building process.

asn-d6 (changes requested):
Thanks! Left some comments!

src/Makefile
jtraglia: This is a good idea. We should definitely do this.

src/ckzg.c
jtraglia: Yes this is needed. Otherwise there is no fft implementation. We just don't want to include this in the header file.

Without it we get these errors:

image

src/tests.c
jtraglia: Because this is a test file for everything, I think it makes sense to include everything.

src/Makefile
jtraglia: Let's do this in a separate PR.

Issue #439, opened 2024-06-18
Documentation of KZGSettings (Branch `das`)


b-wagn:
Note: this issue is about [this](https://github.com/ethereum/c-kzg-4844/tree/das) branch.

When reading the code, I felt that the different fields related to roots of unity in `KZGSettings` ([here](https://github.com/ethereum/c-kzg-4844/blob/das/src/c_kzg_4844.h#L159)) are a bit confusing.
This includes the fields `roots_of_unity`, `expanded_roots_of_unity`, and `reverse_roots_of_unity`.

As a result, it is a bit hard to understand code that uses these fields, or to modify code properly if needed.
For example, here are some questions that I can not immediately resolve from the documentation.
A better documentation and more consistent naming could help to identify what they are.
The main questions that the comments/naming should address for each field are the following:

* Is the field related to a blob or to an extended blob?
* How long is the field, i.e., how many elements does it contain?
* Are these roots in reverse bit ordering or not?

patniemeyer:
FYI on the documentation topic - I wrote a short, developer-level article that shows how to use the ckzg libs from Python and walks through the nature and the sizes of the datatypes (how to chunk the data) including how to use the roots of unity from KZGSettings to address your data elements.  If it is useful it could be expanded upon and turned into documentation.  Link:  [https://pat.net/ckzg/](https://pat.net/ckzg/)

asn-d6:
@b-wagn fixed the big issues with the roots of unity in https://github.com/ethereum/c-kzg-4844/pull/467.
There were a few additional concerns with the `KZGSettings` fields raised in the PR. Feel free to take a stab at them if you would like.

For now, I'm closing this issue. But feel free to open a new one with more specific tasks.

PR #467, opened 2024-08-02
Cleanup roots of unity in KZGSettings

asn-d6:
This PR makes some steps towards https://github.com/ethereum/c-kzg-4844/issues/439

Leftover things to do:
- ~~Completely remove `max_width`. This is not a variable field so it doesn't need to be in `KZGSettings`. It's a parameter of the system, and by having it as a variable it makes things more confusing. IMO we should just use `FIELD_ELEMENTS_PER_EXT_BLOB` directly where it's appropriate, and `FIELD_ELEMENTS_PER_BLOB` in FK20.~~
- At some point we should stop this `domain_size + 1` artifact stemming from `expand_root_of_unity()` since it's needless and confusing. I decided to not do it because it would also change the 4844 code and I would ideally like it to remain as it was.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -100,14 +100,18 @@ typedef Bytes48 KZGProof;
 
 /** Stores the setup and parameters needed for computing KZG proofs. */
 typedef struct {
-    /** The length of `roots_of_unity`, a power of 2. */
+    /** The size of our multiplicative subgroup (the roots of unity). This is the size of
+     *  the extended domain (after the RS encoding has been applied), so the size of
+     *  the subgroup is FIELD_ELEMENTS_PER_EXT_BLOB. */
     uint64_t max_width;
-    /** Powers of the primitive root of unity determined by `SCALE2_ROOT_OF_UNITY` in bit-reversal
-     * permutation order, length `max_width`. */
+    /** Roots of unity in bit-reversal permutation order.
+     *  The array contains `domain_size` elements. */
     fr_t *roots_of_unity;
-    /** The expanded roots of unity. */
+    /** Roots of unity for the subgroup of size `domain_size`.
+     *  The array contains `domain_size + 1` elements, it starts and ends with Fr::one(). */
     fr_t *expanded_roots_of_unity;
-    /** The bit-reversal permuted roots of unity. */
+    /** Roots of unity for the subgroup of size `domain_size` reversed. Only used in FFTs.
+     *  The array contains `domain_size + 1` elements: it starts and ends with Fr::one(). */
     fr_t *reverse_roots_of_unity;
     /** G1 group elements from the trusted setup in monomial form. */
     g1_t *g1_values_monomial;
--- a/src/common.h
+++ b/src/common.h
@@ -111,13 +111,13 @@ typedef struct {
      *
      * The array contains `domain_size + 1` elements, it starts and ends with Fr::one().
      */
-    fr_t *expanded_roots_of_unity;
+    fr_t *roots_of_unity;
     /**
      * Roots of unity in bit-reversed order.
      *
-     * This array is derived by applying a bit-reversal permutation to `expanded_roots_of_unity`
+     * This array is derived by applying a bit-reversal permutation to `roots_of_unity`
      * excluding the last element. Essentially:
-     *   `brp_roots_of_unity = bit_reversal_permutation(expanded_roots_of_unity[:-1])`
+     *   `brp_roots_of_unity = bit_reversal_permutation(roots_of_unity[:-1])`
      *
      * The array contains `domain_size` elements.
      */
@@ -125,8 +125,8 @@ typedef struct {
     /**
      * Roots of unity for the subgroup of size `domain_size` in reversed order.
      *
-     * It is the reversed version of `expanded_roots_of_unity`. Essentially:
-     *    `reverse_roots_of_unity = reverse(expanded_roots_of_unity)`
+     * It is the reversed version of `roots_of_unity`. Essentially:
+     *    `reverse_roots_of_unity = reverse(roots_of_unity)`
      *
      * This array is primarily used in FFTs.
      * The array contains `domain_size + 1` elements, it starts and ends with Fr::one().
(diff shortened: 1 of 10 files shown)


src/common.h
```diff
@@ -100,14 +100,18 @@ typedef Bytes48 KZGProof;
...
-    uint64_t max_width;
-    /** Powers of the primitive root of unity determined by `SCALE2_ROOT_OF_UNITY` in bit-reversal
-     * permutation order, length `max_width`. */
-    fr_t *roots_of_unity;
-    /** The expanded roots of unity. */
+    /** The size of our multiplicative subgroup (the roots of unity). This is the size of
+     *  the extended domain (after the RS encoding has been applied), so the size of
+     *  the subgroup is FIELD_ELEMENTS_PER_EXT_BLOB. */
+    uint64_t domain_size;
+    /** Roots of unity in bit-reversal permutation order.
```
b-wagn, line 107: Comments here are much better than before!
Nit: "bit-reversal permutation order" vs. "bit-reversed order" vs "reverse bit order(ing)" (in some blog posts). 
We should have a consistent name for that, at least within this code.

src/common.h
```diff
@@ -100,14 +100,18 @@ typedef Bytes48 KZGProof;
...
+     *  The array contains `domain_size` elements. */
+    fr_t *brp_roots_of_unity;
```
b-wagn, line 109: For `g1_values_lagrange_brp` few lines below, we have the `brp` as a suffix. Maybe make it consistent with that?

src/common.h
```diff
@@ -100,14 +100,18 @@ typedef Bytes48 KZGProof;
...
+    /** Roots of unity for the subgroup of size `domain_size`.
+     *  The array contains `domain_size + 1` elements, it starts and ends with Fr::one(). */
     fr_t *expanded_roots_of_unity;
-    /** The bit-reversal permuted roots of unity. */
+    /** Roots of unity for the subgroup of size `domain_size` reversed. Only used in FFTs.
+     *  The array contains `domain_size + 1` elements: it starts and ends with Fr::one(). */
     fr_t *reverse_roots_of_unity;
     /** G1 group elements from the trusted setup in monomial form. */
```
b-wagn, line 116: For these (`g1_values_monomial`, `g1_values_lagrange_brp`, `g2_values_monomial`) we could also add the size of the arrays.

src/common.h
```diff
@@ -100,14 +100,18 @@ typedef Bytes48 KZGProof;
...
+    /** Roots of unity in bit-reversal permutation order.
+     *  The array contains `domain_size` elements. */
+    fr_t *brp_roots_of_unity;
+    /** Roots of unity for the subgroup of size `domain_size`.
+     *  The array contains `domain_size + 1` elements, it starts and ends with Fr::one(). */
     fr_t *expanded_roots_of_unity;
-    /** The bit-reversal permuted roots of unity. */
+    /** Roots of unity for the subgroup of size `domain_size` reversed. Only used in FFTs.
+     *  The array contains `domain_size + 1` elements: it starts and ends with Fr::one(). */
     fr_t *reverse_roots_of_unity;
```
b-wagn, line 115: The term "reverse" here is dangerously close to "reverse-bit". Is it possible to use a different term to separate these two concepts more clearly? Also, is `reverse_roots_of_unity` the reversed form of `expanded_roots_of_unity`? Or of `brp_roots_of_unity`? Maybe we can make this a bit more clear 😅.

b-wagn (commented):
Thanks, that's a very important change. 
I left some minor comments.

Also, I totally agree with removing `domain_size` / `max_width`.

src/common.h
asn-d6: Took a stab at it in cb63fb0d7089a2c1076af78c16a12e93ce877ddf
Feel free to suggest alternatives.

src/common.h
asn-d6: I felt like the other roots of unity structures had the adjective as a prefix, so I did the same. Perhaps we can introduce this convention in a separate PR?

src/common.h
asn-d6: IMO let's do this at a separate PR?

src/common.h
asn-d6: Completely agree this was not clear. Took a bit of a maximalist approach in cb63fb0d7089a2c1076af78c16a12e93ce877ddf. Let me know if you don't like it.

src/common.h
b-wagn: Sure 👍

src/common.h
b-wagn: Sure 👍

src/common.h
```diff
@@ -100,14 +100,37 @@ typedef Bytes48 KZGProof;
...
+    uint64_t domain_size;
+    /**
+     * Roots of unity for the subgroup of size `domain_size`.
+     *
+     * The array contains `domain_size + 1` elements, it starts and ends with Fr::one().
+     */
     fr_t *expanded_roots_of_unity;
-    /** The bit-reversal permuted roots of unity. */
+    /**
+     * Roots of unity in bit-reversed order.
```
jtraglia, line 116: Should this be?
```suggestion
     * Roots of unity for the subgroup of size `domain_size` in bit-reversed order.
```

src/common.h
```diff
@@ -100,14 +100,37 @@ typedef Bytes48 KZGProof;
...
+     * The array contains `domain_size` elements.
+     */
+    fr_t *brp_roots_of_unity;
+    /**
+     * Roots of unity for the subgroup of size `domain_size` in reversed order.
+     *
+     * It is the reversed version of `expanded_roots_of_unity`. Essentially:
+     *    `reverse_roots_of_unity = reverse(expanded_roots_of_unity)`
+     *
+     * This array is primarily used in FFTs.
```
jtraglia, line 131: I don't feel that this remark adds all that much.

asn-d6: I added the python pseudocode in an attempt to highlight the difference between this and `brp_roots_of_unity` (both in terms of "reverse" vs "reverse-bit order" and in terms of excluding the last element).

With regards to "This array is primarily used in FFTs.", I was confused on why we need the reverse roots in the first place, so this would be useful to me, but I understand that it might be superfluous.

asn-d6:
Thanks a lot for the help @jtraglia !

PR #469, opened 2024-08-02
Fix Makefile dependencies

### LINKED ISSUE #468: Improve build system (post-file-split Makefile issues)
After the file split (#464) the build system does not seem to detect file changes correctly.

We should probably clarify to the Makefile what's the [dependencies](https://stackoverflow.com/questions/62989870/make-file-doesnt-detect-changes-in-source-files) of our targets.

### How to reproduce:
- Run `make`
- Edit `setup.c`
- Run `make` again (it will not compile `setup.c`)

jtraglia:
This PR defines three new lists:
* `SOURCE_FILES`
* `HEADER_FILES`
* `OBJECT_FILES`

When running `make`, we build each object file individually. This ensures includes are correct.

Fix include issues, discovered from doing this.

When running tests, list all source/header files as dependencies. If there's a change, it will recompile.

Fixes #468.


asn-d6 (approved):
LGTM! Thanks!

PR #471, opened 2024-08-05
Split C code into more files

jtraglia:
This PR continues to split the core C code into more manageable files.

Four new directories have been created:

* `common` -- Contains general code, some of which is shared between eip4844 and eip7594.
  * For example, it contains definitions for BLS points/fields, memory allocations, and various utilities.
* `eip4844` -- As the name implies, contains the implementation for EIP-4844.
  * The core file in this directory (`eip4844.c`) has been mostly left as-as. Don't poke the audited dragon. 
* `eip7594` -- As the name implies, contains the implementation for EIP-7594.
  * There are several files in this directory, such as `fk20.c` which handles FK20 proof generation, `recovery.c` which implements cell recovery, and `fft.c` which provides methods for FFT.
* `setup` -- Contains code pertaining to loading the trusted setup.
  * This also contains the definition for `KZGSettings`, which didn't quite make sense in `common`. This is a bit of a weird situation. We don't want to include it in `eip4844` because it's been updated for `eip7594`.

We needed to include the C source directory in the Rust bindings generator so the updated includes would work.

There are various other small changes, but no changes to the implementation have been made here.

Added `InsertNewlineAtEOF: true` to the clang formatter configuration file. Several of my new files did not contain a newline at the bottom, which showed an ugly crossed out circle on GitHub.


asn-d6:
This looks great!

Some thoughts:
- Not sure if I prefer having separate files for `constants.h` and `types.h` over just stuffing them into the (already quite small) `api.h`.
- I'm not sure if I love the `eip7594/api.c` entry point. I would go for `eip7594/eip7594.c` but I understand that might not be best either. 

Both points above are just personal preference, and I hold no strong opinions.

asn-d6 (approved):
Awesome work! Thanks a lot!

PR #473, opened 2024-08-06
Fix profile & analyze

jtraglia:
So `make profile` and `make analyze` have been broken since the file split. This PR makes the following changes:

* Run analyzer on each source file individually. Doesn't work when bundled together.
* Fix some allocation nits in the tests. There were some potential leaks that the analyzer pointed out.
  * I'm using more stack variables now. They are big, but not big enough to cause problems.
  * This includes storing the result & doing `c_kzg_free` before doing asserts.
* Fix `commitments` issue in one of the profiler functions.

~~To fix the broken test, please merge #474 first.~~

### DIFF
--- a/src/eip7594/recovery.c
+++ b/src/eip7594/recovery.c
@@ -135,6 +135,11 @@ static C_KZG_RET vanishing_polynomial_for_missing_cells(
     );
     if (ret != C_KZG_OK) goto out;
 
+    /* Zero out all the coefficients */
+    for (size_t i = 0; i < FIELD_ELEMENTS_PER_EXT_BLOB; i++) {
+        vanishing_poly[i] = FR_ZERO;
+    }
+
     /*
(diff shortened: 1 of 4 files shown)


src/eip7594/recovery.c
```diff
@@ -135,6 +135,11 @@ static C_KZG_RET vanishing_polynomial_for_missing_cells(
     );
     if (ret != C_KZG_OK) goto out;
 
+    /* Zero out all the coefficients */
+    for (size_t i = 0; i < FIELD_ELEMENTS_PER_EXT_BLOB; i++) {
+        vanishing_poly[i] = FR_ZERO;
+    }
+
```
jtraglia, line 142: Please take note of this.

asn-d6 (approved):
LGTM! Thanks!

PR #474, opened 2024-08-06
Do not rely on calloc zeroing out allocations

jtraglia:
There were two spots which relied on `calloc` zeroing out values. I don't like this.

I left the `MEMSET_TO_FF` macro option to test this. Could be useful in the future. Open to removing it.

And yeah, `malloc` doesn't zero out allocations, but wanted to be consistent.

Also fix a random comment's wrap that I noticed.

### DIFF
--- a/src/common/alloc.c
+++ b/src/common/alloc.c
@@ -21,10 +21,6 @@
 #include <stdbool.h> /* For bool */
 #include <stddef.h>  /* For size_t & NULL */
 #include <stdlib.h>  /* For malloc */
-#include <string.h>  /* For memset */
-
-/** Enable to test non-zeroized allocations. */
-#define MEMSET_TO_FF 0
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 // Memory Allocation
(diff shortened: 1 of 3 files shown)


asn-d6:
Just curious, why you dislike relying on calloc to zero out values?

Also, why do you loop and manually set arrays to `FR_ZERO`, over a memset? Do you dislike relying on `FR_ZERO` being all zeroes?

jtraglia:
> Just curious, why you dislike relying on `calloc` to zero out values?

So `calloc` is required by the C standard to zero out the data, so it should be perfectly safe to rely on this. But it kind of tells me that we're depending on a feature that isn't exactly clear to everyone. It's possible someone unfamiliar with C doesn't know that `calloc` behaves that way.

> Also, why do you loop and manually set arrays to `FR_ZERO`, over a `memset`?

I think the loop is more readable than memset. I think performance difference is negligible & probably optimized away.

>Do you dislike relying on `FR_ZERO` being all zeroes?

To small degree, yes 😄

asn-d6 (approved):
LGTM! Thanks!

PR #475, opened 2024-08-08
Run cargo update for fuzzer

jtraglia:
Simply do a `cargo update` in the fuzz directory so the latest version of rust-eth-kzg is used. This update includes the semi-recent changes to batch verification. We can remove the placeholder argument now too.


asn-d6 (approved):
LGTM! Thanks!

PR #476, opened 2024-08-08
Remove benchmarks for Rust & Java

jtraglia:
Remove benchmarks in the Rust and Java bindings. This is mostly a maintenance issue. I would rather delete them than update them for EIP-7594. In my opinion, these benchmarks don't provide that much extra value.

Now, there will only be benchmarks in the Go bindings.


asn-d6 (approved):
I've been using the rust benchmarks, but I totally get the point about maintainance costs. I will start using the go benchmarks from now on.

LGTM! Thanks!

PR #477, opened 2024-08-08
Remove unnecessary tests in the bindings

jtraglia:
Also controversial, but there are a bunch of "old" tests in the bindings that existed before there were reference tests. So each of the bindings kind of did their own thing, then added the reference tests. So they are testing the same thing, but are considerably more complex. We haven't added non-reference tests for EIP-7594 to each of the bindings. This is not something that I would like to do either.


asn-d6:
I assume this has been a maintainance burden everytime we change the API right?
If not, I would just keep them around. If yes, I agree, let's remove them and rely on the test vectors.

jtraglia:
It's not really a maintenance burden in that sense. Closing.

PR #478, opened 2024-08-08
Replace `size_t` with `uint64_t`

jtraglia:
Opening as a draft. Not entirely sure I want this. I've had some mild concerns with size_t being a dynamic size. I'm not sure if my concerns are justified. This PR replaces all instances of `size_t` with `uint64_t`.


asn-d6:
Can you elaborate on what you think is the issue here? It would be weird if we were accepting `size_t` as input on a public method, since there could be cross-platform disagreements, but I don't see anything wrong with using the type in internal functions.

jtraglia:
Hmm I don't have a strong opinion on what internal types we should use. I thought it might be good to use `uint64_t` everywhere for consistency, but that's not a great reason. Maybe I should revert that, and only do `uint64_t` for the public API functions, like you suggested. This is originally what I had in mind, but I'm on the fence as to what I think is best.

Screenshot of the task list:
image

jtraglia:
I will make these changes sometime next week. Doesn't feel like a high priority right now.

asn-d6:
This looks good! Thanks!

As a question, is it possible to replace the `ulong` of the C# bindings to `uint64` as well? https://github.com/ethereum/c-kzg-4844/blob/b58cd9d00a0d1ee6362d6bec6d99299b0a7cad65/bindings/csharp/Ckzg.Bindings/Ckzg.Bindings.cs#L67-L68

PR #480, opened 2024-08-09
Enable more compiler warnings

jtraglia:
So once again, I feel like I've been tricked by the compiler. I thought that `-Wall -Wextra` was everything, but it's not.

This PR enables these warnings & fixes the issues. The most noticeable change is probably the due to the implicit conversions from `uint64_t` to `uint32_t` to `int` etc that we were doing. Rather than dealing with conversions, we can use `uint64_t` in these places. I actually like this better & think we should have done this from the beginning.

See this link for a list of flags. Not all of them are listed here 🤷 

https://gcc.gnu.org/onlinedocs/gcc-4.4.0/gcc/Warning-Options.html#Warning-Options

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -20,7 +20,16 @@ ifeq ($(PLATFORM),Darwin)
 endif
 
 # The base compiler flags. More can be added on command line.
-CFLAGS += -I. -I../inc -O2 -Wall -Wextra
+CFLAGS += -I. -I../inc -O2
+# Enable all of the optional warnings.
+CFLAGS += -pedantic -Wall -Wextra -Wconversion -Wsign-conversion -Wsign-compare \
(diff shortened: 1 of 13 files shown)


asn-d6:
This PR does not build on my system.
I'm using `clang version 16.0.6 (27+b1)`. Maybe you are building with gcc?

jtraglia:
I'm actually using Apple's clang. They have their own fork if I understand correctly.

```
$ clang -v
Apple clang version 15.0.0 (clang-1500.3.9.4)
```

What are the issues?

asn-d6 (approved):
LGTM! Thanks!

asn-d6:
The issues were an artifact of a separate build system issue. Will address on another PR.

PR #482, opened 2024-08-12
Simplify nim bindings

jtraglia:
I would like to simplify the Nim bindings before the big v2.0 release. This PR does the following:

* Remove the "zero overhead aliases" which I find pretty confusing.
  * I think one could argue that these are over-complicated & prone to issues.
  * When adding EIP-7594 support, I made several mistakes.
* Remove duplicate tests, only use reference tests.
  * The tests for each API (kzg, kzg_abi, kzg_ex) were too much.
  * It's okay to only run through the reference tests.
* Use a global `KzgCtx` like the other bindings do.
  * This is simpler & also less prone to mistakes.
  * Requires that `free_trusted_setup` is called on exit.

Some thoughts on this from our discussion on discord:
image

Opening as a draft until @agnxsh tests it.

I would also appreciate a review from @jangko, the author of these bindings.


jangko (approved):
LGTM. Besides `loadTrustedSetupFromString`, Nim actually can optimize it further via compile time mechanism to load baked in array from executable binary. But that better put into another PR.

jtraglia:
@jangko Oh cool. Yes let's do that in another PR.

Any thoughts on the optimization issue on Windows? I've disabled compiler optimizations as a bandaid fix until we figure it out. I asked about it on Discord (in Nimbus's PeerDAS channel):

> Also, I think I've run into a Nim compiler bug. For some reason, the Windows release build will crash when calling `computeCellsAndKzgProofs`. The debug build works and the release build with `--opt:none` works too. I spent a couple hours trying to figure it out. No issues on Linux either. Is there a Nim expert that could look into this?

To be clear, this happens when the high level wrapper function (`computeCellsAndKzgProofs`), not the ABI function, is called & it is not crashing due to anything the function is doing. I put an echo before the function was called and another at the top of the function. Only the first one printed.

jtraglia:
Thank you @jangko for identifying the issue:

> Maybe related to `completeStruct` pragma being absent from `KzgCell`? Because I see other types/struct are using `completeStruct`. Not completely sure though.

After adding this to the type, the Windows CI tests pass again.

jtraglia:
This branch was tested by Agnish (nimbus dev) and confirmed it compiles & passes the reference tests. Going to merge.

PR #483, opened 2024-08-13
Improve the build system

jtraglia:
This makes a few notable improvements to the build system:

* If the blst submodule has been updated, it will re-build & copy files.
* Do not build object files individually. The build prints won't show now.
  * I kind of liked this though, because it ensured imports were correct.
* Add prints for each rule, so the developer knows what's going on.
* Also, fix build for profiler (it was using too much stack).
* Also, build blst in the Java bindings. If the library didn't exist, it would error.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+src/.blst_hash
 src/tests
 src/tests_*
 coverage.html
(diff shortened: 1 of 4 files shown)


jtraglia:
Should be non-controversial. George said it worked for him last week. Going to merge.

PR #484, opened 2024-08-13
Rename Kzg to KZG in bindings

jtraglia:
This PR touches a lot of file, but it's pretty straightforward. Renames `Kzg` to `KZG` in the bindings. For consistency. I used find/replace to do this, so I'm pretty sure all of the instances are updated.

I'm afraid these changes might be too annoying for clients though.


jtraglia:
I've changed my mind. I think this is too much. Closing.

PR #485, opened 2024-08-14
Upgrade blst to v0.3.13

jtraglia:
blst made a new release a few weeks ago:

https://github.com/supranational/blst/releases/tag/v0.3.13

This also updates the version in the Go & Rust bindings.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -118,9 +118,9 @@ checksum = "37b2a672a2cb129a2e41c10b1224bb368f9f37a2b16b612598138befd7b37eb5"
 
 [[package]]
 name = "cc"
-version = "1.1.8"
+version = "1.1.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


jtraglia:
I'll fix the windows build issues tomorrow morning.

jtraglia:
So there's a problem with the `build.bat` file used to build blst on Windows. On [this line](https://github.com/supranational/blst/blob/52cc60d78591a56abb2f3d0bd1cdafc6ba242997/build.bat#L3), `%PATH%` is expanded twice (once in the cmd & once in the comment) and will grow past the 8192-character limit if PATH is greater than 4096 characters. When using [`msvc-dev-cmd`](ttps://github.com/ilammy/msvc-dev-cmd) and several other actions (including calling `VsDevCmd.bat` myself) PATH will be greater than that 4096-character limit. We could modify PATH to only include the paths we need, but that's a lot of work and would be especially difficult in the Python release script, which we cannot modify PATH. I've talked with Andy (dot-asm) about this and he is aware of the issue. I expect the next release will include a fix for this. Therefore, I suggest we skip this release (it's pretty minor) and wait for the next one.

PR #486, opened 2024-08-15
Use `TestCaseSource` in C# reference tests

jtraglia:
When adding support for EIP-7594, @flcl42 recommended we use `TestCaseSource`. I liked this, but only wanted to change the new tests, not the old EIP-4844 ones. This PR makes these changes for the old tests.

See: https://github.com/ethereum/c-kzg-4844/pull/452#discussion_r1676406466


jtraglia:
This should be non-controversial. Going to merge.

PR #487, opened 2024-08-15
Enable -Werror for Windows & cleanup warnings

jtraglia:
For some reason, the Windows compiler shows more accurate warnings than Unix.

It accurately made these observations:

* `-Wc++-compat` and `-Wredundant-move` are only for C++ code.
* `-Wlarger-than=512` should (and does) happen all the time.
  * Lots of our stack variables are large than 512 bytes.
* `-Wlong-long` might also happen all the time.


jtraglia:
This should be non-controversial. Going to merge.

PR #488, opened 2024-08-15
Update deps in Go/Rust/Fuzz

jtraglia:
* Configure blst to be v0.3.12 for both (we forgot to do this).
* Run `go get -u github.com/supranational/blst@v0.3.12` and `go mod tidy`.
* Run `cargo update` in `c-kzg-4844/` and `c-kzg-4844/fuzz`.


asn-d6:
LGTM! Thanks!

PR #490, opened 2024-08-15
Convert some internal variables from `uint64_t` to `size_t`

jtraglia:
I reviewed the C code to see where we could use `size_t` instead of `uint64_t`.


asn-d6:
LGTM! Thanks!

PR #494, opened 2024-08-16
`verify_cell_kzg_proof_batch()`: Abstract commitment to interpolation poly

asn-d6:
After a bunch of failed attempts to simplify `verify_cell_kzg_proof_batch()`, I found a good spot to apply a scalpel.

This refactoring is fairly clean and moves away four allocations from the main function into a smaller contained function.

The first commit only moves code (can be verified with `git diff --color-moved`) while the second applies a few minor changes.

There is a few more minor improvements that can be done, but they all have drawbacks. This seems quite clean IMO.

### DIFF
--- a/src/eip7594/eip7594.c
+++ b/src/eip7594/eip7594.c
@@ -459,12 +459,12 @@ out:
  *
  * This function computes `RLI = [sum_k r^k interpolation_poly_k(s)]` from the spec.
  *
- * @param[out]  commitment_out      Commitment to the aggregated interpolation poly
- * @param[in]   r_powers            Precomputed powers of the random challenge
- * @param[in]   cell_indices        Indices of the cells
- * @param[in]   cells               Array of cells
- * @param[in]   num_cells           Number of cells
- * @param[in]   s                   The trusted setup
+ * @param[out]  commitment_out  Commitment to the aggregated interpolation poly
+ * @param[in]   r_powers        Precomputed powers of the random challenge
+ * @param[in]   cell_indices    Indices of the cells
+ * @param[in]   cells           Array of cells
+ * @param[in]   num_cells       Number of cells
+ * @param[in]   s               The trusted setup
  */
 static C_KZG_RET compute_commitment_to_aggregated_interpolation_poly(
     g1_t *commitment_out,
--- a/src/eip7594/eip7594.c
+++ b/src/eip7594/eip7594.c
@@ -480,14 +480,10 @@ static C_KZG_RET compute_commitment_to_aggregated_interpolation_poly(
     fr_t *aggregated_interpolation_poly = NULL;
     bool *is_cell_used = NULL;
 
-    /*
-     * This function first aggregates cells from the same column by scaling them with the
-     * corresponding powers of the random challenge. It then computes interpolation polynomials
-     * using the aggregated cells. It finally sums up the interpolation polynomials and commits to
-     * the aggregated interpolation polynomial.
-     */
+    ////////////////////////////////////////////////////////////////////////////////////////////////
+    // Array allocations
+    ////////////////////////////////////////////////////////////////////////////////////////////////
 
-    /* Allocate memory */
     ret = new_fr_array(&aggregated_column_cells, FIELD_ELEMENTS_PER_EXT_BLOB);
     if (ret != C_KZG_OK) goto out;
     ret = new_fr_array(&column_interpolation_poly, FIELD_ELEMENTS_PER_CELL);
@@ -496,6 +492,10 @@ static C_KZG_RET compute_commitment_to_aggregated_interpolation_poly(
     if (ret != C_KZG_OK) goto out;
     ret = new_bool_array(&is_cell_used, FIELD_ELEMENTS_PER_EXT_BLOB);
     if (ret != C_KZG_OK) goto out;
+
+    ////////////////////////////////////////////////////////////////////////////////////////////////
+    // Aggregates cells from the same column
+    ////////////////////////////////////////////////////////////////////////////////////////////////
 
     /* Start with zeroed out columns */
     for (size_t i = 0; i < CELLS_PER_EXT_BLOB; i++) {
@@ -523,6 +523,10 @@ static C_KZG_RET compute_commitment_to_aggregated_interpolation_poly(
             is_cell_used[index] = true;
         }
     }
+
+    ////////////////////////////////////////////////////////////////////////////////////////////////
+    // Compute interpolation polynomials using the aggregated cells
+    ////////////////////////////////////////////////////////////////////////////////////////////////
 
     /* Start with a zeroed out poly */
     for (size_t i = 0; i < FIELD_ELEMENTS_PER_CELL; i++) {
@@ -571,6 +575,10 @@ static C_KZG_RET compute_commitment_to_aggregated_interpolation_poly(
             );
         }
     }
+
+    ////////////////////////////////////////////////////////////////////////////////////////////////
+    // Commit to the aggregated interpolation polynomial
+    ////////////////////////////////////////////////////////////////////////////////////////////////
 
     /* Commit to the final aggregated interpolation polynomial */
     ret = g1_lincomb_fast(
--- a/src/eip7594/eip7594.c
+++ b/src/eip7594/eip7594.c
@@ -506,14 +506,20 @@ static C_KZG_RET compute_commitment_to_aggregated_interpolation_poly(
         }
     }
 
-    /* Scale each cell's data points */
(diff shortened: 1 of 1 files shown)


src/eip7594/eip7594.c
```diff
@@ -454,6 +454,125 @@ static C_KZG_RET compute_r_powers_for_verify_cell_kzg_proof_batch(
...
+ * This function first aggregates cells from the same column by scaling them with the corresponding
+ * powers of the random challenge. It then computes the sum of the interpolation polynomials from
+ * the aggregated columns and finally commits to the aggregated interpolation polynomial.
+ *
+ * @param[out]  commitment_out              Commitment to the aggregated interpolation poly
+ * @param[in]   r_powers                    Precomputed powers of the random challenge
+ * @param[in]   cell_indices                Indices of the cells
+ * @param[in]   cells                       Array of cells
+ * @param[in]   num_cells                   Number of cells
+ * @param[in]   s                           The trusted setup
```
jtraglia, line 471: We can delete some space between param name & description.

src/eip7594/eip7594.c
```diff
@@ -454,6 +454,125 @@ static C_KZG_RET compute_r_powers_for_verify_cell_kzg_proof_batch(
     return ret;
 }
 
+/**
+ * Aggregate columns, compute the sum of interpolation polynomials, and commit to the result.
+ *
+ * This function computes: [sum_k r^k interpolation_poly_k(s)]
```
jtraglia, line 460: What is `interpolation_poly` exactly? Should this be?
```suggestion
 * This function computes: [sum_k r^k aggregated_interpolation_poly_k(s)]
```

src/eip7594/eip7594.c
```diff
@@ -454,6 +454,125 @@ static C_KZG_RET compute_r_powers_for_verify_cell_kzg_proof_batch(
...
+    size_t num_cells,
+    const KZGSettings *s
+) {
+    C_KZG_RET ret;
+    fr_t *aggregated_column_cells = NULL;
+    fr_t *column_interpolation_poly = NULL;
+    fr_t *aggregated_interpolation_poly = NULL;
+    bool *is_cell_used = NULL;
+
+    // Allocate memory
```
jtraglia, line 487: Inconsistent comment style. We prefer `/* ... */`.

src/eip7594/eip7594.c
```diff
@@ -454,6 +454,125 @@ static C_KZG_RET compute_r_powers_for_verify_cell_kzg_proof_batch(
...
+    if (ret != C_KZG_OK) goto out;
+
+    // Initialize aggregated columns and usage flags
+    for (size_t i = 0; i < CELLS_PER_EXT_BLOB; i++) {
+        for (size_t j = 0; j < FIELD_ELEMENTS_PER_CELL; j++) {
+            size_t index = i * FIELD_ELEMENTS_PER_CELL + j;
+            aggregated_column_cells[index] = FR_ZERO;
+            is_cell_used[index] = false;
+        }
+    }
```
jtraglia, line 504: Hmm "flags" isn't the ideal word here. Not sure what is though. Maybe "markers"?

src/eip7594/eip7594.c
```diff
@@ -454,6 +454,125 @@ static C_KZG_RET compute_r_powers_for_verify_cell_kzg_proof_batch(
...
+ * @param[in]   cells                       Array of cells
+ * @param[in]   num_cells                   Number of cells
+ * @param[in]   s                           The trusted setup
+ */
+static C_KZG_RET compute_commitment_to_aggregated_interpolation_poly(
+    g1_t *commitment_out,
+    const fr_t *r_powers,
+    const uint64_t *cell_indices,
+    const Cell *cells,
+    size_t num_cells,
```
jtraglia, line 478: Ugh. Technically, `num_cells` from the caller is a `uint64_t` type.

src/eip7594/eip7594.c
```diff
@@ -454,6 +454,125 @@ static C_KZG_RET compute_r_powers_for_verify_cell_kzg_proof_batch(
...
+        shift_poly(column_interpolation_poly, FIELD_ELEMENTS_PER_CELL, &inv_coset_factor);
+
+        for (size_t k = 0; k < FIELD_ELEMENTS_PER_CELL; k++) {
+            blst_fr_add(
+                &aggregated_interpolation_poly[k],
+                &aggregated_interpolation_poly[k],
+                &column_interpolation_poly[k]
+            );
+        }
+    }
```
jtraglia, line 557: You deleted a bunch of old comments. Can we add those back?

src/eip7594/eip7594.c
```diff
@@ -615,95 +722,17 @@ C_KZG_RET verify_cell_kzg_proof_batch(
...
-                &aggregated_interpolation_poly[k],
-                &column_interpolation_poly[k]
-            );
-        }
-    }
-
-    /* Commit to the final aggregated interpolation polynomial */
-    ret = g1_lincomb_fast(
-        &evaluation, s->g1_values_monomial, aggregated_interpolation_poly, FIELD_ELEMENTS_PER_CELL
+    /* Aggregated cells from same columns, sum interpolation polynomials, and commit */
```
jtraglia, line 728: Oh one more thing. This should be:
```suggestion
    /* Aggregate cells from same columns, sum interpolation polynomials, and commit */
```

src/eip7594/eip7594.c
jtraglia: Note, they don't have to be the same comments, just anything to help the reader.

src/eip7594/eip7594.c
```diff
@@ -615,95 +722,17 @@ C_KZG_RET verify_cell_kzg_proof_batch(
...
-            );
-
-            /* Mark the cell as being used */
-            is_cell_used[index] = true;
-        }
-    }
-
-    ////////////////////////////////////////////////////////////////////////////////////////////////
-    // Compute sum of the interpolation polynomials
+    // Commmit to aggregated interpolation polynomial: [sum_k r^k interpolation_poly_k(s)]
```
jtraglia, line 725: I think maybe this section should be about `final_g1_sum` now, since it's the final "result" of it.

jtraglia, line 725: Typo and that last part is duplicate information & not all that useful here.
They can check the helper function's docs.
```suggestion
    // Commit to aggregated interpolation polynomial
```

src/eip7594/eip7594.c
```diff
@@ -454,6 +454,125 @@ static C_KZG_RET compute_r_powers_for_verify_cell_kzg_proof_batch(
...
+            ret = bytes_to_bls_field(&cell_fr, (const Bytes32 *)&cells[i].bytes[offset]);
+            if (ret != C_KZG_OK) goto out;
+            blst_fr_mul(&scaled_fr, &cell_fr, &r_powers[i]);
+            size_t index = cell_indices[i] * FIELD_ELEMENTS_PER_CELL + j;
+            blst_fr_add(
+                &aggregated_column_cells[index], &aggregated_column_cells[index], &scaled_fr
+            );
+
+            // Mark the cell as being used
+            is_cell_used[index] = true;
```
b-wagn, line 520: As this PR only moved this, this is a comment that probably should be addressed in a separate PR:
We set `is_cell_used` at indices which are not a multiple of `FIELD_ELEMENTS_PER_CELL`, but only read it at multiples of `FIELD_ELEMENTS_PER_CELL`.
This seems unnecessary. Why not just set `is_cell_used[cell_indices[i]] = true` in the outer loop and making the `is_cell_used` array smaller?

src/eip7594/eip7594.c
```diff
@@ -454,6 +454,125 @@ static C_KZG_RET compute_r_powers_for_verify_cell_kzg_proof_batch(
     return ret;
 }
 
+/**
+ * Aggregate columns, compute the sum of interpolation polynomials, and commit to the result.
+ *
+ * This function computes: [sum_k r^k interpolation_poly_k(s)]
+ *
+ * This function first aggregates cells from the same column by scaling them with the corresponding
```
b-wagn, line 462: I feel like this comment is not describing *what* the function does, but instead it describes *how* it does it.
Maybe as the user of the function I don't need to know that. So maybe move this comment into the function?

b-wagn (commented):
That's a good idea! Thanks.
Left two comments.

src/eip7594/eip7594.c
asn-d6: Clarified that this is a quote from the spec. It might help people who read the spec and match it with the code.

jtraglia (approved):
LGTM. Thanks!

b-wagn (approved):
LGTM! Thank you both :)

PR #495, opened 2024-08-16
Refactor is_cell_used array & usage

jtraglia:
The `is_cell_used` array was a little misleading. It tracked individual field elements, not cells.


src/eip7594/eip7594.c
```diff
@@ -524,12 +523,23 @@ static C_KZG_RET compute_commitment_to_aggregated_interpolation_poly(
...
+    // Determine which cells are used
+    ////////////////////////////////////////////////////////////////////////////////////////////////
+
+    /* Start with false values */
+    for (size_t i = 0; i < CELLS_PER_EXT_BLOB; i++) {
+        is_cell_used[i] = false;
+    }
+
+    /* Mark each cell index as used */
+    for (uint64_t i = 0; i < num_cells; i++) {
```
b-wagn, line 539: Why is `i` a `size_t` above, and a `uint64_t` here? Any particular reason?

b-wagn (approved):
Nice!

src/eip7594/eip7594.c
jtraglia: Yeah it's a bit of an odd situation. It's because of `num_cells` type. `num_cells` is a `uint64_t` because it comes from the public API while we try to use `size_t` internally.

b-wagn: ok!

PR #497, opened 2024-08-19
Update links in README

jtraglia:
* Remove link to Deneb's polynomial commitment API. Feels weird not linking EIP-7594's doc too. I would like to link it, but maybe we should do it later, when it's part of the Electra spec.
* Remove link to Deneb's KZG tests, point to consensus-spec-tests instead.
* Fix link to C tests.
* Delete an unnecessary "the".
* Use relative links for internal files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 # C-KZG-4844
 
-A minimal implementation of the [Polynomial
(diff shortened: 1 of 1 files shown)


asn-d6 (approved):
LGTM! Thanks!

PR #501, opened 2024-08-23
Add param linter script for C source

jtraglia:
Add a python script which auto-formats the `@param` lines in the C source code.


jtraglia:
I'm going to merge this. It shouldn't be controversial.

PR #502, opened 2024-08-26
fix: node.js publish

matthewkeil:
When publishing 4.0.0 there was an error and the bundle published without the `deps` folder.  The `make command` did not correctly switch directories and published the root folder, not the `dist` folder.  I updated the process to use the canonical method that node uses to avoid future issues.

I also updated the supported/tested versions of node that are run in CI

### DIFF
--- a/.github/workflows/nodejs-tests.yml
+++ b/.github/workflows/nodejs-tests.yml
@@ -17,8 +17,9 @@ jobs:
           - macos-latest
           - windows-latest
         node:
-          - 16
           - 18
+          - 20
+          - 22
     steps:
       - uses: actions/checkout@v3
         with:
--- a/.github/workflows/nodejs-tests.yml
+++ b/.github/workflows/nodejs-tests.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v4
         with:
-          python-version: '3.10'
+          python-version: "3.10"
       - name: Install setuptools
         run: |
           python -m pip install --upgrade pip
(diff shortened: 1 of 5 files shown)


.github/workflows/nodejs-tests.yml
```diff
@@ -30,7 +31,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v4
         with:
-          python-version: '3.10'
+          python-version: "3.10"
```
matthewkeil, line 34: This was an auto-format change.  Can revert if you would like.

bindings/node.js/package.json
```diff
@@ -1,10 +1,10 @@
 {
   "name": "c-kzg",
-  "version": "4.0.0",
+  "version": "4.0.1",
```
matthewkeil, line 3: Found the issue after running the publish command with 4.0.0 so needed to bump the minor.  Not sure why this showed up now though.

bindings/node.js/Makefile
```diff
@@ -44,17 +44,6 @@ build: install clean
...
-# Bundle the distribution, also helpful for cross compatibility checks
-.PHONY: bundle
-bundle: build
-	@mkdir dist
-	@mv deps dist
-	@cp -r lib dist/lib
-	@cp -r src dist/src
-	@cp README.md dist/README.md
-	@cp package.json dist/package.json
-	@cp binding.gyp dist/binding.gyp
```
matthewkeil, line 56: The `files` key in the `package.json` is what controls this behavior.  There is no need to build the `dist` folder.  Went with this before because you guys seemed to like it so rode the wave but the error showed up so switched this to the canonical method of using the `files` key to control what gets bundled in the npm tarball

jtraglia (approved):
LGTM. Thank you!

PR #504, opened 2024-08-30
Add "/v2" to Go module name

jtraglia:
Seems that this is necessary for a v2 release...

https://pkg.go.dev/github.com/ethereum/c-kzg-4844?tab=versions

### DIFF
--- a/bindings/go/README.md
+++ b/bindings/go/README.md
@@ -6,7 +6,7 @@ exported functions in [C-KZG-4844](https://github.com/ethereum/c-kzg-4844).
 ## Installation
 
 ```
-go get github.com/ethereum/c-kzg-4844
+go get github.com/ethereum/c-kzg-4844/v2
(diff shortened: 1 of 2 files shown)


kevaundray (approved):
quite unfortunate :(

jtraglia:
Hey @tarassh thanks for the comment. I'm definitely not an expert on this but I think it's enough to append v2 to the module name. We will still need to make a new v2.0.1 release for this to show up on pkg.go.dev though. Are you sure the v2 code must go into a new v2 directory?

jtraglia:
> Eventually you should be able to see this https://pkg.go.dev/github.com/googleapis/gax-go?tab=versions with this repo.

I see. But that project chose that way of handling things. It's not strictly necessary. Unless it's required, I'd rather not maintain two versions of the codebase.

PR #505, opened 2024-09-03
Update version to v2.0.1 and do cargo updates

jtraglia:
This increments the version number in preparation for a new minor release.

Also run `cargo update` in the Rust bindings and fuzzing directory.

(We forgot to update `kzg4844.nimble` for the v2 release 😢)


asn-d6 (approved):
LGTM! Thanks!

PR #506, opened 2024-09-07
Fix profiling instructions for Linux

agnxsh:
The debian package name `gperftools` is now deprecated, and is replaced with `libgoogle-perftools-dev`. Additionally the `pprof` binary seems to be decoupled from perftools, for debian we would separately install it from Go packages.

image

Overall the instructions would look like this now:
image

### DIFF
--- a/src/PROFILE.md
+++ b/src/PROFILE.md
@@ -10,7 +10,7 @@ took, but it will help you determine which functions are the most expensive.
 
 On Linux (Debian), you need to install:
 ```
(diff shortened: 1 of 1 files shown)


jtraglia:
Thanks! I'm going to wait to merge this until you've successfully run the profiler. There might be other issues; it's been a while since I've tested this on Linux.

jtraglia (approved):
LGTM. Thank you!

PR #508, opened 2024-09-16
Fix findings from Dmitry's review

jtraglia:
This PR is a work in progress. It does the following things:

* Update `com.diffplug.spotless` to work with Java 22.
* Rename `compute_fk20_proofs` to `compute_fk20_cell_proofs`.
* Hardcode `compute_fk20_cell_proofs` to accept `FIELD_ELEMENTS_PER_BLOB` length poly.
* Add instructions to install `build` in Python bindings.
* Add comment about `fr_is_null` edge case in recovery.
* Handle zero-length inputs to FFT functions.
* Optimize calculation for `coset_factor_pow`.
* Replace signed integers with unsigned types.
* Assert that the domain strings are the correct length.
* Add precomputed array of RBL cell indices.

Big thanks to @khovratovich 🙏

If we choose to, we can do larger renaming fixes in a separate PR.

### DIFF
--- a/src/eip7594/eip7594.c
+++ b/src/eip7594/eip7594.c
@@ -538,9 +538,9 @@ out:
  * Compute the inverse coset factor h_k^{-1},
  *  where `h_k` is the coset factor for cell with index `k`.
  *
- * @param[out]  inv_coset_factor_out  Pointer to store the computed inverse coset factor
- * @param[in]   cell_index            The index of the cell
- * @param[in]   s                     The trusted setup
+ * @param[out]  inv_coset_factor_out    Pointer to store the computed inverse coset factor
+ * @param[in]   cell_index              The index of the cell
+ * @param[in]   s                       The trusted setup
  */
 static void get_inv_coset_shift_for_cell(
     fr_t *inv_coset_factor_out, uint64_t cell_index, const KZGSettings *s
@@ -570,9 +570,9 @@ static void get_inv_coset_shift_for_cell(
 /**
  * Compute h_k^{n}, where `h_k` is the coset factor for cell with index `k`.
  *
- * @param[out]  coset_factor_out  Pointer to store h_k^{n}
- * @param[in]   cell_index        The index of the cell
- * @param[in]   s                 The trusted setup
+ * @param[out]  coset_factor_out    Pointer to store h_k^{n}
+ * @param[in]   cell_index          The index of the cell
+ * @param[in]   s                   The trusted setup
  */
 static void get_coset_shift_pow_for_cell(
     fr_t *coset_factor_out, uint64_t cell_index, const KZGSettings *s


src/eip7594/eip7594.c
```diff
@@ -648,10 +667,8 @@ static C_KZG_RET compute_commitment_to_aggregated_interpolation_poly(
         if (ret != C_KZG_OK) goto out;
 
         /* Now divide by the coset shift factor */
-        uint64_t pos = reverse_bits_limited(CELLS_PER_EXT_BLOB, i);
-        fr_t inv_coset_factor;
-        blst_fr_eucl_inverse(&inv_coset_factor, &s->roots_of_unity[pos]);
-        shift_poly(column_interpolation_poly, FIELD_ELEMENTS_PER_CELL, &inv_coset_factor);
+        uint64_t pos = -CELL_INDICES_RBL[i] % FIELD_ELEMENTS_PER_EXT_BLOB;
```
asn-d6, line 670: Please see my other comment about `coset_shift_for_cell()`. This change can be confusing IMO. Not sure what `-CELL_INDICES_RBL[i]` is supposed to be, let alone comboed the modulus operator.

src/eip4844/eip4844.c
```diff
@@ -106,9 +103,10 @@ static C_KZG_RET fr_batch_inv(fr_t *out, const fr_t *a, int len) {
 
     blst_fr_eucl_inverse(&accumulator, &accumulator);
 
-    for (i = len - 1; i >= 0; i--) {
-        blst_fr_mul(&out[i], &out[i], &accumulator);
-        blst_fr_mul(&accumulator, &accumulator, &a[i]);
+    for (size_t i = len; i > 0; i--) {
+        size_t index = i - 1;
+        blst_fr_mul(&out[index], &out[index], &accumulator);
+        blst_fr_mul(&accumulator, &accumulator, &a[index]);
```
asn-d6, line 109: Hmm. IMO it's confusing to have one loop index, and then edit it first thing inside the loop.  Can we just leave the code as it was (but with size_t)? It's 4844 code as well, so I like not changing it if it's not broken.

src/eip7594/eip7594.c
```diff
@@ -709,11 +726,11 @@ static C_KZG_RET computed_weighted_sum_of_proofs(
...
 
     for (uint64_t i = 0; i < num_cells; i++) {
-        uint64_t pos = reverse_bits_limited(CELLS_PER_EXT_BLOB, cell_indices[i]);
-        fr_t coset_factor = s->roots_of_unity[pos];
-        // Compute h_k^n, with h_k and n as in the spec.
-        fr_pow(&coset_factor_pow, &coset_factor, FIELD_ELEMENTS_PER_CELL);
-        // Scale the power of r by h_k^n
+        /* Compute h_k^n, with h_k and n as in the spec */
+        uint64_t pos = CELL_INDICES_RBL[cell_indices[i]];
+        coset_factor_pow = s->roots_of_unity[pos * FIELD_ELEMENTS_PER_CELL];
```
asn-d6, line 731: IMO this is not a straightforward change. I understand what we are trying to do, but not sure exactly why this should work. Is it possible to introduce a function similar to `coset_shift_for_cell()` from the spec (feel free to have our own naming) which does this logic with a bit of documentation?

asn-d6 (changes requested):
Thanks for this. Highlighting some things that can be improved IMO.

src/eip7594/eip7594.c
jtraglia: That's equivalent to the following:

```c
FIELD_ELEMENTS_PER_EXT_BLOB - CELL_INDICES_RBL[i] % FIELD_ELEMENTS_PER_EXT_BLOB
```

It's the formula I came up with to get the inverse root of unity.

src/eip4844/eip4844.c
asn-d6: Thanks for the fix.

src/eip7594/eip7594.c
khovratovich: You probably meant

`(FIELD_ELEMENTS_PER_EXT_BLOB - CELL_INDICES_RBL[i]) % FIELD_ELEMENTS_PER_EXT_BLOB`

jtraglia: Yes, thank you.

src/eip7594/cell.c
```diff
@@ -26,7 +26,7 @@
  */
 void print_cell(const Cell *cell) {
     for (size_t i = 0; i < FIELD_ELEMENTS_PER_CELL; i++) {
-        const Bytes32 *field = (const Bytes32 *)&cell->bytes[i * BYTES_PER_FIELD_ELEMENT];
-        print_bytes32(field);
+        const Bytes32 *field_element = (const Bytes32 *)&cell->bytes[i * BYTES_PER_FIELD_ELEMENT];
```
b-wagn, line 29: Using something like `element_fr` would be slightly more consistent with the rest of the code.
But it's a bit ugly.

src/eip7594/fft.c
```diff
@@ -130,7 +138,7 @@ C_KZG_RET fr_ifft(fr_t *out, const fr_t *in, size_t n, const KZGSettings *s) {
 
     fr_t inv_len;
     fr_from_uint64(&inv_len, n);
-    blst_fr_inverse(&inv_len, &inv_len);
+    blst_fr_eucl_inverse(&inv_len, &inv_len);
```
b-wagn, line 141: Just noting here that `inv_len` is not a good name if this is the inverse of `n`. Why not `inv_n`?
Currently, it reads more like "the length of the inverse".

src/eip7594/eip7594.c
b-wagn: I still think it is confusing. Maybe a more detailed comment would help? 😅

b-wagn (commented):
Thank you @jtraglia  and also @khovratovich ! 

I left some comments, but most of it looks good :)

src/eip7594/eip7594.c
```diff
@@ -728,10 +745,10 @@ static C_KZG_RET computed_weighted_sum_of_proofs(
  * Given some cells, verify that their proofs are valid.
  *
  * @param[out]  ok                  True if the proofs are valid
- * @param[in]   commitments_bytes   The commitments for the cells
- * @param[in]   cell_indices        The cell indices for the cells
- * @param[in]   cells               The cells to check
- * @param[in]   proofs_bytes        The proofs for the cells
+ * @param[in]   commitments_bytes   The commitments for the cells, length `num_cells`
+ * @param[in]   cell_indices        The indices for the cells, length `num_cells`
+ * @param[in]   cells               The cells to check, length `num_cells`
```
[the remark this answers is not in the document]
jtraglia: Yeah good point. Maybe we don't make these changes? Or just do it for everything?

src/eip7594/fk20.c
```diff
@@ -113,7 +112,9 @@ C_KZG_RET compute_fk20_proofs(g1_t *out, const fr_t *p, size_t n, const KZGSetti
 
     /* Compute toeplitz coefficients and organize by column */
     for (size_t i = 0; i < FIELD_ELEMENTS_PER_CELL; i++) {
-        ret = toeplitz_coeffs_stride(toeplitz_coeffs, p, n, i, FIELD_ELEMENTS_PER_CELL);
+        ret = toeplitz_coeffs_stride(
+            toeplitz_coeffs, p, FIELD_ELEMENTS_PER_BLOB, i, FIELD_ELEMENTS_PER_CELL
```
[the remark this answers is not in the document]
jtraglia: Hmm we could hardcode more constants if we think that's fine. Do you think we should?

src/eip7594/cell.c
jtraglia: Hmm, it would need to be `element_bytes` in this case, as that's the type. I'll do that 👍

src/eip7594/eip7594.c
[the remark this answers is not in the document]
b-wagn: If it is not too much work, I would prefer doing it for everything. But that could be a separate PR.

src/eip7594/fk20.c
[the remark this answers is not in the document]
b-wagn: I think at this particular point, it was not clear why we don't do it consistently. So at least for `toeplitz_coeffs_stride`, I think we should do it.

src/eip7594/eip7594.c
```diff
@@ -159,6 +176,7 @@ C_KZG_RET compute_cells_and_kzg_proofs(
  * @param[in]   num_cells           The number of available cells provided
  * @param[in]   s                   The trusted setup
  *
+ * @remark At least 50% of CELLS_PER_EXT_BLOB cells must be provided.
```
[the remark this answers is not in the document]
khovratovich: Moreover the current code actually has the expansion rate `2` hardcoded in several places as an integer rather than as a named constant. If we need the code scalable to other rates, then the constant should be named.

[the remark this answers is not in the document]
jtraglia: Good idea. I'll make an `EXPANSION_FACTOR` constant.

src/eip7594/fk20.c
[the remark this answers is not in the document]
jtraglia: I decided to clean this function up. It was pretty confusing. Do y'all think this is better?

src/eip7594/eip7594.c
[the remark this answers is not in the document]
jtraglia: Also decided to make a `CELLS_PER_BLOB` constant.

src/eip7594/eip7594.c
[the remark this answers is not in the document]
jtraglia: I made these changes here. I think I got them all.

src/eip7594/eip7594.c
jtraglia: Hmm yes & no. If we did that, we'd have to manually take the power of it again. And it would only be used once.

src/eip7594/eip7594.c
jtraglia: I've added some slightly better comments. Let me know if those are fine.

src/eip7594/cell.h
```diff
@@ -30,6 +30,15 @@
 /** The number of bytes in a single cell. */
 #define BYTES_PER_CELL (FIELD_ELEMENTS_PER_CELL * BYTES_PER_FIELD_ELEMENT)
 
+/** The Reed-Solomon erasure coding expansion factor. */
+#define EXPANSION_FACTOR 2
+
+/** The number of field elements in an extended blob */
```
b-wagn, line 36:
```suggestion
/** The number of field elements in an extended blob. */
```

src/eip7594/poly.c
```diff
@@ -31,7 +31,7 @@
  * Multiplies each coefficient by `shift_factor ^ i`. Equivalent to creating a polynomial that
  * evaluates at `x * shift_factor` rather than `x`.
  *
- * @param[in,out]   p               The polynomial coefficients to be scaled
+ * @param[in,out]   p               The polynomial coefficients to be scaled, length `p`
```
b-wagn, line 34: Should it be the following?
```suggestion
 * @param[in,out]   p               The polynomial coefficients to be scaled, length `len`
```

b-wagn (approved):
Changes look good to me 👍

src/eip7594/cell.h
jtraglia: Ah nice catch. Good nit pick.

src/eip7594/eip7594.c
khovratovich, line 1: Some divisions by 2 should be replaced by divisions by `EXPANSION_FACTOR`

src/eip7594/poly.c
jtraglia: Yes, it should... good catch again.

src/eip7594/eip7594.c
jtraglia: Hmm yes. In this file, I only saw one. Were there others?

asn-d6 (approved):
LGTM! Thanks a lot everyone here!

PR #513, opened 2024-09-26
7594 Bindings: Trivial improvements

asn-d6:
Went over the C#/Java/Nim bindings. Here are some trivial improvements. Still need to go over Rust and Golang.


jtraglia (approved):
Ah nice catches. Thank you!

PR #514, opened 2024-10-08
Fix inconsistent parameter doc in blob.c

b-wagn:
An input parameter `n` was specified in the documentation.
As it was not present in the function, we remove it in this PR.


jtraglia (approved):
Ah thank you. Good catch.

Issue #512, opened 2024-09-25
Constant for Expansion Factor


b-wagn:
### Background

The current code uses the fact that the number of field elements in a blob and the number of field elements in an extended blob relate by a factor of two. In other words, we currently use a rate 1/2 code. It came up in [this](https://github.com/ethereum/c-kzg-4844/pull/508) PR that this is *potentially* an issue, say if we later want to go to a different rate. 

As part of the PR, an `EXPANSION_FACTOR = 2` constant has been introduced and certain hardcoded 2's have been replaced by `EXPANSION_FACTOR`. However, it was clear that (1) not all occurrences have been found, and (2) that not all 2's are actually related to the rate of the code. It was therefore not clear that the code would still work for other reasonable constants like `EXPANSION_FACTOR = 4`. Therefore, this constant has been removed again.

### Purpose of this Issue 
This issue is to make us not forget about this discussion. If in the future someone wants to do a PR introducing the `EXPANSION_FACTOR` constant properly, then it is in particular important to pay attention to (1) and (2), and to test and document for which values of `EXPANSION_FACTOR` the code works. It is also still subject to discussion whether it is likely that we need to use a different rate in the future.

PR #515, opened 2024-10-09
Add LOG_EXPANSION_FACTOR constant

b-wagn:
This PR is to address #512.

Note that in contrast to what is stated in the issue, the code will *not* work for other constants as it is.
This is because it contains precomputed constants such as `ROOT_OF_UNITY` that are tailored to the expansion factor being two (`LOG_EXPANSION_FACTOR = 1`). We could have added code to compute the constants on the fly based on the expansion factor, but this would add significant overhead. Therefore, introducing the constant is more a way to improve readability instead of a way to make the code more general.

### DIFF
--- a/src/eip7594/cell.h
+++ b/src/eip7594/cell.h
@@ -30,8 +30,15 @@
 /** The number of bytes in a single cell. */
 #define BYTES_PER_CELL (FIELD_ELEMENTS_PER_CELL * BYTES_PER_FIELD_ELEMENT)
 
+/**
+ * The logarithm (base 2) of the expansion factor.
+ * In other words, this defines the rate (blob / extended blob).
+ * Note that the code is not guaranteed to work anymore if this is changed.
+ */
+#define LOG_EXPANSION_FACTOR 1
+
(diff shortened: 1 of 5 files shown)


jtraglia (approved):
LGTM! Thanks for this, this is better 😄

src/eip7594/cell.h
```diff
@@ -30,8 +30,15 @@
 /** The number of bytes in a single cell. */
 #define BYTES_PER_CELL (FIELD_ELEMENTS_PER_CELL * BYTES_PER_FIELD_ELEMENT)
 
+/**
+ * The logarithm (base 2) of the expansion factor.
+ * In other words, this defines the rate (blob / extended blob).
+ * Note that the code is not guaranteed to work anymore if this is changed.
```
asn-d6, line 36:
```suggestion
/**
 * The logarithm (base 2) of the expansion factor of our Reed-Solomon code.
 * In other words, this defines the rate of the Reed-Solomon code (blob / extended blob).
 * Note that our codebase is not guaranteed to work anymore if this is changed.
```

PR #523, opened 2024-12-13
Fix typo 

dannbbb1:
Changes:

yaml → YAML
bindings\nim → bindings/nim

This change is useful because:

YAML: The term "YAML" is an acronym (YAML Ain't Markup Language) and should always be capitalized to maintain consistency and correctness in technical documentation. Using "yaml" in lowercase can lead to confusion or make the text appear less professional.

bindings/nim: Changing bindings\nim to bindings/nim corrects an incorrect use of the newline character (\n), ensuring the path or reference is formatted properly. This avoids errors in file paths or URLs, which could otherwise prevent the system from locating the correct resources.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ A minimal implementation of the Polynomial Commitments API for
 
 While the core implementation is in C, bindings are available for various
 high-level languages, providing convenient wrappers around C functions. These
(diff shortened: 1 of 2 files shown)


flechemano:
Yes, little things are important things

PR #525, opened 2024-12-30
Update github actions versions due to deprecation

parithosh:
Artifact actions v3 will be closing down by January 30, 2025 and github is going to begin temporarily failing jobs that don't update before this deadline. The PR fixes the versions used. The migration docs [here](https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md) were followed.

PR #529, opened 2025-01-15
Fix NuGet packaging

### LINKED ISSUE #528: The NuGet package missing the `runtimes` directory
The latest version [2.0.1.1254](https://www.nuget.org/packages/Ckzg.Bindings/2.0.1.1254) is missing its `runtimes` directory for some reason which makes the package unusable. This version must be deprecated (and may be replaced with a newer one) as it's the default choice and causes crashes.

cc @jtraglia

rubo:
Fixes #528 

- Fixed the issue with NuGet package artifacts introduced in #525
- Removed checks for artifacts of the `runtimes` directory in the `.csproj` file to fail the build when those files are missing
- Updated and consolidated the target .NET versions from 6 and 7 to the latest LTS version of 8
- Updated dependencies
- Optimized artifacts downloading in GitHub Actions workflow

### DIFF
--- a/.github/workflows/csharp-tests.yml
+++ b/.github/workflows/csharp-tests.yml
@@ -58,7 +58,7 @@ jobs:
     - name: Upload artifacts
       uses: actions/upload-artifact@v4
       with:
-        name: ckzg-library-wrapper-${{ matrix.target.location }}-${{ matrix.target.arch || 'default' }}
+        name: ckzg-library-wrapper-${{ matrix.target.location }}
         path: bindings/csharp/Ckzg.Bindings/runtimes/${{ matrix.target.location }}/native
 
   test-ckzg-dotnet:
@@ -84,7 +84,7 @@ jobs:
       run: cd bindings/csharp && dotnet restore
     - uses: actions/download-artifact@v4
       with:
-        pattern: 'ckzg-library-wrapper-${{ matrix.target.location }}-*'
+        pattern: ckzg-library-wrapper-${{ matrix.target.location }}
         path: bindings/csharp/Ckzg.Bindings/runtimes/${{ matrix.target.location }}/native
         merge-multiple: true
     - name: Test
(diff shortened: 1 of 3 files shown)


.github/workflows/csharp-tests.yml
```diff
@@ -99,34 +106,25 @@ jobs:
...
-        pattern: ckzg-library-wrapper-*
-        path: bindings/csharp/Ckzg.Bindings/runtimes/win-x64/native
-    - uses: actions/download-artifact@v4
-      with:
-        pattern: ckzg-library-wrapper-*
-        path: bindings/csharp/Ckzg.Bindings/runtimes/osx-arm64/native
-    - uses: actions/download-artifact@v4
-      with:
-        pattern: ckzg-library-wrapper-*
-        path: bindings/csharp/Ckzg.Bindings/runtimes/linux-arm64/native
```
rubo, line 118: These actions were downloading all 5 artifacts 5 times for no reason. Now, it does once and later puts each of them where they should be.

.github/workflows/csharp-tests.yml
```diff
@@ -84,9 +84,16 @@ jobs:
...
+        pattern: ckzg-library-wrapper-*
+        path: bindings/csharp/Ckzg.Bindings/runtimes
+    - name: Move artifacts
+      working-directory: bindings/csharp/Ckzg.Bindings/runtimes
+      run: |
+        mv ckzg-library-wrapper-linux-arm64/ckzg.so linux-arm64/native/ckzg.so
+        mv ckzg-library-wrapper-linux-x64/ckzg.so linux-x64/native/ckzg.so
+        mv ckzg-library-wrapper-osx-arm64/ckzg.dylib osx-arm64/native/ckzg.dylib
+        mv ckzg-library-wrapper-osx-x64/ckzg.dylib osx-x64/native/ckzg.dylib
+        mv ckzg-library-wrapper-win-x64/ckzg.dll win-x64/native/ckzg.dll
```
rubo, line 96: Since the checks have been removed from the `.csproj` file, we need to ensure that each artifact is present even though it is not used on the platform being run.

jtraglia (approved):
This looks great. Thanks for the other updates too!

PR #531, opened 2025-01-22
feat: add windows-gnu to rust tests

Rjected:
This allows the build script to work with `windows-gnu`, by using `flag_if_supported` on the `/std:c11` flag, and gating the `i32` repr change to msvc.


jtraglia (approved):
Nice. LGTM, thank you!

PR #537, opened 2025-02-19
fix: fix issue with find_space_after returning None

indmind:
There was a problem in the `find_space_after` function: if no space was found after the given `start_index`, the function was returning `None`, which doesn't match the expected return type of `int`.
I’ve fixed this by adding `return -1` in case no space is found.
This ensures the function always returns an integer value as expected.


jtraglia (approved):
Thanks for noticing this! Your change doesn't really make the error any better though, the exception will still be confusing. I've pushed an additional commit which improves this.

PR #539, opened 2025-02-26
Allocate polynomials on heap instead of stack

jtraglia:
I think we should allocate polynomials on the heap instead of the stack. These are quite large and could cause problems if, for whatever reason, there isn't much available stack space. In this PR, I've also renamed some variables for clarity.

### DIFF
--- a/src/eip4844/eip4844.c
+++ b/src/eip4844/eip4844.c
@@ -48,29 +48,20 @@ static const char *FIAT_SHAMIR_PROTOCOL_DOMAIN = "FSBLOBVERIFY_V1_";
 static const char *RANDOM_CHALLENGE_DOMAIN_VERIFY_BLOB_KZG_PROOF_BATCH = "RCKZGBATCH___V1_";
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
-// Types
-////////////////////////////////////////////////////////////////////////////////////////////////////
-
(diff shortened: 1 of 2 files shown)


src/eip4844/eip4844.c
```diff
@@ -417,26 +410,22 @@ C_KZG_RET compute_kzg_proof(
...
-    fr_t *y_out,
-    const Polynomial *polynomial,
-    const fr_t *z,
-    const KZGSettings *s
+    KZGProof *proof_out, fr_t *y_out, const fr_t *poly, const fr_t *z, const KZGSettings *s
 ) {
     C_KZG_RET ret;
     fr_t *inverses_in = NULL;
     fr_t *inverses = NULL;
+    fr_t *q = NULL;
```
asn-d6, line 423: We could consider renaming this to `q_poly` or `quotient_poly`, to match the convention of renaming all polynomials to `poly` instead of using single-letter variable names.

asn-d6 (approved):
Looks great! Left a small variable-name nit. Thanks!

src/eip4844/eip4844.c
jtraglia: I agree. Good suggestion. Will make this change in a bit.

asn-d6: Doing it and pushing here.

PR #550, opened 2025-03-14
Fix elixir default path

BlazeWasHere:
allows you to run elixir based commands in both root folder and `bindings/elixir`

```sh
c-kzg-4844 ~> mix test
# all good
c-kzg-4844 ~> cd bindings/elixir
c-kzg-4844/bindings/elixir ~> mix test
# all good
```


jtraglia (approved):
Nice, LGTM 👍

PR #553, opened 2025-03-28
feat: add arbitrary feature

Rjected:
This gates all arbitrary code behind a feature

### DIFF
--- a/bindings/rust/src/bindings/mod.rs
+++ b/bindings/rust/src/bindings/mod.rs
@@ -8,9 +8,6 @@ mod serde;
 #[cfg(test)]
 mod test_formats;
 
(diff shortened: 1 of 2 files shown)


bindings/rust/src/bindings/mod.rs
```diff
@@ -8,6 +8,7 @@ mod serde;
 #[cfg(test)]
 mod test_formats;
 
+#[cfg(feature = "arbitrary")]
 use arbitrary::Arbitrary;
```
mattsse, line 12: can we use full paths insteads and remove feature gated import?

mattsse (changes requested):
nit

but this should really be optional

PR #561, opened 2025-04-08
fix: rm unsafe from extern that breaks compilation

rakita:
Found here: https://github.com/foundry-rs/foundry-fork-db/actions/runs/14331517446/job/40168425837
happens on 1.8.2 and newer rust

Have removed `unsafe` from the generated code so CI can pass, and it will not be brought back when we regenerate bindings in future.


rakita:
All good, no need for this PR, unsafe will be required in rust 2024

PR #562, opened 2025-04-28
Rename namespace so it's different from main class, and we don't need…

flcl42:
… to always write Ckzg.Ckzg.Foo


jtraglia (approved):
Ah 😅 LGTM, thanks!

agnxsh (approved):

PR #563, opened 2025-04-28
Use correct constant in coeffs allocation

jtraglia:
When testing with different cell sizes, I noticed that the constant used in this allocation is wrong; it should have used the `FIELD_ELEMENTS_PER_CELL` constant instead. When `FIELD_ELEMENTS_PER_CELL` is 64, then `CELLS_PER_BLOB` (non-extended) is also 64. But when `FIELD_ELEMENTS_PER_CELL` is 32, `CELLS_PER_BLOB` is 128. This means that we would over-allocate when using a different FEPC count.

For reference, see how `coeffs` is used in this section of code (might need to scroll):

https://github.com/ethereum/c-kzg-4844/blob/0f5e47dc26d79f10bd37084d33e6a174ff06d64b/src/eip7594/fk20.c#L112-L148


agnxsh (approved):

asn-d6 (approved):
LGTM! Thanks!

PR #565, opened 2025-04-28
Allow n<=1 in bit_reversal_permutation

jtraglia:
When testing `FIELD_ELEMENTS_PER_CELL=1`, I got a `C_KZG_BADARGS` from `bit_reversal_permutation` because we treated 1 as an invalid length. We just hadn't considered this being valid, but it is. For 0 and 1, we can just return without doing anything. This function is only used with controlled constants, so it shouldn't really affect anything.


agnxsh (approved):

asn-d6 (approved):
LGTM! Thanks!

PR #566, opened 2025-04-28
Move FIELD_ELEMENTS_PER_EXT_BLOB to blob.h

jtraglia:
I believe the `FIELD_ELEMENTS_PER_EXT_BLOB` definition makes more sense here.


asn-d6 (approved):
LGTM! Thanks!

PR #568, opened 2025-04-28
Remove global settings from go bindings

jtraglia:
Note: this is an API change and would require a thoughtful release.

If/when we decide to reduce cell sizes (ie lower `FIELD_ELEMENTS_PER_CELL`) clients will probably need to have two versions of ckzg loaded at the same time. When the fork happens, the client will switch contexts and proceed like normal.


jtraglia:
Going to close this. May re-open in the future if/when we need this.

PR #569, opened 2025-04-28
Add commitment cache optimization

jtraglia:
This feels a little dirty, but I've been looking for any optimization that would speed up cell verification. I believe this is a worthy optimization. Let's establish the context:

* When clients receive a column, they will call `verify_cell_kzg_proof_batch` on the cells for that column.
* They may do this in a multi-threaded way; they will spawn a thread to verify the column.
* For each column in a slot, the commitments are expected to be the same.
* Columns will usually be for the same slot; random columns for other slots will not be received.
* At t+0 we'll start to receive columns, by t+4 we'll probably receive most columns for that slot. 

For the first column, there will be no improvement, but on subsequent calls, there will be a ~3ms improvement. For a basic node, who will custody 4 columns, this will save at least 9ms when verifying columns it custodies. For a super node, which custodies 128 columns, this will save at least 381ms of CPU time.

### Before
```
Benchmark/VerifyColumns(count=1)                      76          15628912 ns/op
Benchmark/VerifyColumns(count=2)                      55          20550682 ns/op
Benchmark/VerifyColumns(count=4)                      37          31408771 ns/op
Benchmark/VerifyColumns(count=8)                      22          52450987 ns/op
Benchmark/VerifyColumns(count=16)                     12          93017035 ns/op
Benchmark/VerifyColumns(count=32)                      6         172958548 ns/op
Benchmark/VerifyColumns(count=64)                      4         328076802 ns/op
Benchmark/VerifyColumns(count=128)                     2         639059375 ns/op
```

### After
```
Benchmark/VerifyColumns(count=1)                      97          12343080 ns/op
Benchmark/VerifyColumns(count=2)                      67          17212544 ns/op
Benchmark/VerifyColumns(count=4)                      42          28790411 ns/op
Benchmark/VerifyColumns(count=8)                      22          49900112 ns/op
Benchmark/VerifyColumns(count=16)                     12          91822337 ns/op
Benchmark/VerifyColumns(count=32)                      6         172098938 ns/op
Benchmark/VerifyColumns(count=64)                      4         324300136 ns/op
Benchmark/VerifyColumns(count=128)                     2         631046188 ns/op
```

### DIFF
--- a/bindings/rust/build.rs
+++ b/bindings/rust/build.rs
@@ -102,6 +102,8 @@ fn make_bindings(
         /*
          * Cleanup instructions.
          */
+        // Use core::ffi for ctypes
+        .ctypes_prefix("core::ffi")
         // Remove stdio definitions related to FILE.
         .opaque_type("FILE")
         // Remove the definition of FILE to use the libc one, which is more convenient.
--- a/bindings/rust/src/bindings/generated.rs
+++ b/bindings/rust/src/bindings/generated.rs
@@ -106,7 +106,7 @@ pub struct KZGSettings {
     #[doc = " The scratch size for the fixed-base MSM."]
     scratch_size: usize,
     #[doc = " A lock for the commitment cache."]
-    comm_cache_lock: *mut u8,
+    comm_cache_lock: *mut ::std::os::raw::c_long,
     #[doc = " The bytes used as the commitment cache key."]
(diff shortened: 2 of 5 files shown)


jtraglia:
Let's shelve this PR for now. If clients express a desire for this, we can reconsider it.

PR #570, opened 2025-04-29
Split settings initialization into a new function

jtraglia:
I would like to initialize the settings struct prior to both `load_trusted_setup` functions. Splitting this into a function makes this cleaner/easier.


asn-d6 (approved):
LGTM! Thanks!

PR #573, opened 2025-05-19
Update fk20.c

khovratovich:
* Expand the documentation to include the details of the FK20 multiproof algorithm
* Rename some variables to be consistent with the paper
* Replace an incorrect usage of `circulant_domain_size` with `CELLS_PER_EXT_BLOB`

### DIFF
--- a/src/eip7594/fk20.c
+++ b/src/eip7594/fk20.c
@@ -25,14 +25,30 @@
 /**
  * Reorder and extend polynomial coefficients for the toeplitz method, strided version.
  *
- * @param[out]  out     The reordered polynomial, length `CELLS_PER_EXT_BLOB`
+ * This is an auxiliary function that selects the polynomial coefficient for the circulant matrix
+ * in the FK20 multiproof algorithm (Section 3)
+ *
+ * The constants in this function correspond to the FK20 notation as follows:
+ * FIELD_ELEMENTS_PER_CELL =  `l`
+ * CELLS_PER_BLOB = `r`
+ * CELLS_PER_EXT_BLOB = `n`
+ * FIELD_ELEMENTS_PER_BLOB = `d` +1
+ * @offset = `i`
+ *
+ * This function outputs the first column of the circulant matrix `F''_i`,
+ * The matrix `F''_i` is the padding of the Toeplitz matrix of size (r-1)*(r-1) to
+ * the size 2r*2r
+ *
+ * It is supposed to output (in[d-i],0,...,0 (`r` zeros),p[d-(r-2)l-i,p[d-(r-3)l-i,...,p[d-l-i])
+ *
+ * @param[out]  out     The reordered polynomial, length `2*CELLS_PER_BLOB`
  * @param[in]   in      The input polynomial, length `FIELD_ELEMENTS_PER_BLOB`
- * @param[in]   offset  The offset
+ * @param[in]   offset  The offset, the integer between 0 and FIELD_ELEMENTS_PER_BLOB-1, inclusive
  */
-static void toeplitz_coeffs_stride(fr_t *out, const fr_t *in, size_t offset) {
+static void circulant_coeffs_stride(fr_t *out, const fr_t *in, size_t offset) {
     /* Calculate starting indices */
     size_t out_start = CELLS_PER_BLOB + 2;
-    size_t in_start = CELLS_PER_EXT_BLOB - offset - 1;
+    size_t in_start = 2*CELLS_PER_BLOB - offset - 1;
 
     /* Set the first element */
     out[0] = in[FIELD_ELEMENTS_PER_BLOB - 1 - offset];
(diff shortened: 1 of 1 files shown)


jtraglia:
Hey @khovratovich thanks for the PR. Just need to fix the linting and we can get this merged.

src/eip7594/fk20.c
```diff
@@ -25,14 +25,30 @@
...
  * @param[in]   in      The input polynomial, length `FIELD_ELEMENTS_PER_BLOB`
- * @param[in]   offset  The offset
+ * @param[in]   offset  The offset, the integer between 0 and FIELD_ELEMENTS_PER_BLOB-1, inclusive
  */
-static void toeplitz_coeffs_stride(fr_t *out, const fr_t *in, size_t offset) {
+static void circulant_coeffs_stride(fr_t *out, const fr_t *in, size_t offset) {
     /* Calculate starting indices */
     size_t out_start = CELLS_PER_BLOB + 2;
-    size_t in_start = CELLS_PER_EXT_BLOB - offset - 1;
+    size_t in_start = 2*CELLS_PER_BLOB - offset - 1;
```
asn-d6, line 51: Hm. Does this change anything in practice? We use an RS code with rate=1/2, so we always have `CELLS_PER_EXT_BLOB == 2*CELLS_PER_BLOB`.

IIUC Justin created the [original PR](https://github.com/ethereum/c-kzg-4844/pull/567) because this function broke if we changed `FIELD_ELEMENTS_PER_CELL` from 64 to 32. I don't quite see how this patch fixes this.

src/eip7594/fk20.c
```diff
@@ -25,14 +25,30 @@
...
+ * This function outputs the first column of the circulant matrix `F''_i`,
+ * The matrix `F''_i` is the padding of the Toeplitz matrix of size (r-1)*(r-1) to
+ * the size 2r*2r
+ *
+ * It is supposed to output (in[d-i],0,...,0 (`r` zeros),p[d-(r-2)l-i,p[d-(r-3)l-i,...,p[d-l-i])
+ *
+ * @param[out]  out     The reordered polynomial, length `2*CELLS_PER_BLOB`
  * @param[in]   in      The input polynomial, length `FIELD_ELEMENTS_PER_BLOB`
- * @param[in]   offset  The offset
+ * @param[in]   offset  The offset, the integer between 0 and FIELD_ELEMENTS_PER_BLOB-1, inclusive
```
asn-d6, line 46: Can we document what this "offset" does in practice?

src/eip7594/fk20.c
```diff
@@ -49,7 +65,57 @@ static void toeplitz_coeffs_stride(fr_t *out, const fr_t *in, size_t offset) {
 }
```
asn-d6, line 65: This approach for filling the array fills more complex than it should be. Here are a few things we could improve:
- `out_start` is not a great var name, especially when the array is called `out`. Also what is `in_start`? What starts based on those variables? Maybe we should doc or rename.
- This function manages to use all the constants somehow: `CELLS_PER_BLOB`, `FIELD_ELEMENTS_PER_BLOB`, `CELLS_PER_EXT_BLOB`, `FIELD_ELEMENTS_PER_CELL`. It's really hard to read it and be confident that it actually fills the entire array.

If the bug is still here, is there a way to make this code a bit more readable?

asn-d6 (changes requested):
Thanks for the improvements! I left some questions I think we should address before merging.

src/eip7594/fk20.c
jtraglia: Yeah it should be `2 * FIELD_ELEMENTS_PER_CELL`. That's what he said in the other PR.

asn-d6: I think the fact that we can change those constants without any asserts triggering is problematic. We might be able [to design an assert like this](https://github.com/ethereum/c-kzg-4844/blob/main/src/eip7594/eip7594.c#L452) to be more defensive.

khovratovich: The rate of the RS code is another constant, called `LOG_EXPANSION_FACTOR`. However, the factor 2 in this case is of completely different kind, as it is a feature of the FK20 algorithm. I.e. the code will break whenever a different rate will be used, and this will be quite hard to catch.

I am not sure we need asserts for these constants as the code is supposed to work for other power-of-two constants.

src/eip7594/fk20.c
khovratovich: See lines 35 and 41

asn-d6:
Pushed an attempt at improving readability. What I did:
- Added helper vars that map to the paper
- Explicitly stated the array values in the funcdoc
- Zeroed out the entire array in the beginning (tiny bit of perf loss. hopefully compiler replaces it with a memset)
- Replaced reverse loop with forward loop
- Introduced `d_minus_i` helper var which appears in a bunch of places
Let me know what you think! :-)

asn-d6 (approved):
LGTM! Woo!

PR #574, opened 2025-05-21
fixed misleading error message copy-pastes

felix314159:
I wanted to create a proof for Prague and used `z = bytes([2])` and got error about my blob not having correct size. Inspected code and it seems same error message has been copied lots of times and twice it was forgotten to replace blobs with actual thing. So my blob was fine I just had to change z to
```
z = 2
z_valid_size: bytes = z.to_bytes(Spec.BYTES_PER_FIELD_ELEMENT, byteorder="big")
```


asn-d6 (approved):
LGTM! Thanks!

PR #587, opened 2025-07-22
src/Makefile: Enable CC override and remove -Wformat=2 from CFLAGS

leon-anavi:
* Enable overriding of CC in the Makefile to support alternative build systems.
* Remove -Wformat=2 from CFLAGS to avoid Werror issues like: [-Werror=missing-prototypes], [-Werror=conversion], [-Werror=nested-externs], [-Werror=redundant-decls], [-Werror=strict-aliasing], [-Werror=strict-aliasing] and [-Werror=sign-conversion]

These changes are necessary for compatibility with different build systems. Similar patches have already been applied in [Fedora](https://src.fedoraproject.org/rpms/python-ckzg/tree/f42) as well as in [the Yocto Project and OpenEmbedded layer meta-openembedded](https://git.openembedded.org/meta-openembedded/commit/?h=master-next&id=57b88d28b3b8b866cab0f06f1cc73700d67b70b8).

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -20,7 +20,7 @@ ifeq ($(PLATFORM),Darwin)
 endif
 
 # The base compiler flags. More can be added on command line.
-CFLAGS += -I. -I../inc -O2 -Werror
+CFLAGS += -I. -I../inc
 # Enable a bunch of optional warnings.
 CFLAGS += \
 	-pedantic \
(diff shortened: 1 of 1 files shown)


lemenkov:
@leon-anavi thanks for stepping up with this one!
@jtraglia looks quite straightforward to me - doesn't change anything for a current workflows and adds a bit of flexibility for those who keen to experiment. What do you think?

jtraglia:
Hey @leon-anavi & @lemenkov. Sorry about not leaving a review here. I understand why you want to make these changes, but I disagree with them. Most notably, I do not want to remove `-Werror` as I like treating warnings as errors. Unfortunately this project does not work very well with gcc, so clang is a required dependency. I'd prefer not to relax error reporting in order to support other compilers, that's how bugs get introduced.

PR #588, opened 2025-07-25
Remove point-at-infinity filter in g1_lincomb_fast

jtraglia:
Now that blst properly handles this, we can remove this filtering 🙂

See: https://github.com/supranational/blst/commit/ae643bad61ac6fd618bffd771d0ccfddfddf8aee


asn-d6 (approved):
LGTM! This feels much cleaner.

Perhaps it's time to also remove the second `@remark` in the function.

jtraglia:
> Perhaps it's time to also remove the second @remark in the function.

I'll follow up with another PR after the current audit is complete :)

PR #589, opened 2025-07-31
In Java bindings, delete more local references

jtraglia:
Teku shared with me that their memory usage with ckzg is higher than expected (~10gb). I believe the issue is that we were not properly deleting local references.

### DIFF
--- a/bindings/java/ckzg_jni.c
+++ b/bindings/java/ckzg_jni.c
@@ -19,6 +19,7 @@ void reset_trusted_setup(void) {
 void throw_exception(JNIEnv *env, const char *message) {
   jclass exception_class = (*env)->FindClass(env, "java/lang/RuntimeException");
   (*env)->ThrowNew(env, exception_class, message);
+  (*env)->DeleteLocalRef(env, exception_class);
 }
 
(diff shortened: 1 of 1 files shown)


asn-d6:
What do you think about trying out a goto-based free strategy, like we do in the core library code?
The calls to `DeleteLocalRef` in the functions that can throw errors get quite complicated. 
IIUC we could implement goto-based freeing since [the docs say](https://docs.oracle.com/en/java/javase/11/docs/specs/jni/functions.html?utm_source=chatgpt.com#deletelocalref) that `localRef: a local reference. The function does nothing in the case of a NULL value passed here.`.

Also, have we tested that these changes fix the memleaks? Should we get a Teku devs to do some testing?

tbenr:
I spent some time on this and I don't think those changes matter than much. We are not facing a memory leak issue here.

the relevant reading is this: https://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/design.html#wp1242

I don't think we fall into the situation in which is better to manually call `DeleteLocalRef`.

As per doc:
`Local references are valid for the duration of a native method call, and are automatically freed after the native method returns.` this should be good enough for us.

PR #590, opened 2025-07-31
Check if blob is nil in ComputeCells/ComputeCellsAndKZGProofs

jtraglia:
Missing checks here. If provided nil inputs, it would panic.


asn-d6 (approved):
LGTM! Thanks!

PR #600, opened 2025-09-15
Update blst to fix point-at-infinity issue in `g1_lincomb_fast`

jtraglia:
Reported by Teku & Nimbus. There was an issue with `verify_cell_kzg_proof_batch` when there were at least 8 cells being verified & one of the cell's commitments was the point-at-infinity. It seems that blst's v0.3.15 support for points at infinity in Pippenger MSM was insufficient. Regarding why the existing reference tests did not catch this... When there are fewer than 8 inputs, ckzg uses the naive linear combination implementation which works as intended. Supranational recently pushed a commit to blst which fixes the issue, but this isn't included in a release yet. See these PRs for reference:

* https://github.com/supranational/blst/commit/f48500c1fdbefa7c0bf9800bccd65d28236799c1
* https://github.com/ethereum/c-kzg-4844/pull/588
* https://github.com/ethereum/consensus-specs/pull/4590

This PR bumps blst to the latest commit; the submodule & binding dependencies.

A regression test from consensus-specs is included too. The CI check for spec consistency is expected to fail until there's another spec release.


kevaundray (approved):

jtraglia:
Going to merge now. Not going to wait for a new spec release 😅

PR #637, opened 2026-03-09
Use std OnceLock to avoid wasted CPU time

DaniPopes:
`once_cell`'s `OnceBox` allows races during initialization, and then only one thread gets to set the value. `std`'s `OnceLock` waits for the first thread to finish computation.


jtraglia (approved):
Nice, this is much nicer. Do you need a release with this?

DaniPopes:
That would be nice, yes

PR #641, opened 2026-04-06
Add zig bindings

omerfirmak:


bindings/zig/src/root.zig
```diff
@@ -0,0 +1,263 @@
...
+            sliceConstPtr(Blob, blobs),
+            sliceConstPtr(Bytes48, commitments_bytes),
+            sliceConstPtr(Bytes48, proofs_bytes),
+            blobs.len,
+            &self.inner,
+        ));
+        return ok;
+    }
+
+    pub fn computeCellsAndKzgProofs(
```
jtraglia, line 170: We should also provide a binding for `computeCells`.

bindings/zig/src/tests.zig
jtraglia, line 1: We need to run the reference tests here. For example, see:

* https://github.com/ethereum/c-kzg-4844/blob/main/bindings/python/tests.py

.gitignore
```diff
@@ -23,6 +23,8 @@ __pycache__
 
 # nimble build dir
 build/
+.zig-cache/
+zig-out/
```
jtraglia, line 27: Just a nit, let's move this to the bottom of the file with a new `# zig` comment.

omerfirmak:
@jtraglia addressed your comments, all green.

jtraglia:
> @jtraglia addressed your comments, all green.

Very nice. I will give this another review soon!

jtraglia (approved):
LGTM, nice work! I pushed a few small changes for nits I saw while reviewing; please review those when you have time. I don't like that the tests convert all of the YAML files to JSON files, but it appears that [`zig-yaml`](https://github.com/kubkon/zig-yaml) does not work with zig v0.16.0 yet and there are no other good alternatives. So I've accepted this as a necessary evil for now. If/when there is better YAML support in zig, let's revisit this. Other than that, everything looks pretty good. The bindings themselves are quite simple 🙂

omerfirmak:
> but it appears that [`zig-yaml`](https://github.com/kubkon/zig-yaml) does not work with zig v0.16.0 yet and there are no other good alternatives

Yeap, that is the main reason I had to go this route. Not having any external Zig dependencies is great for keeping up with zig compiler versions as well.

> please review those when you have time

they all LGTM, thanks for the touch up:)

PR #655, opened 2026-07-21
Add test for compute_cells_and_kzg_proofs with precomputation

cavdarahmet:
The precompute path (used when precompute > 0, as the README recommends for
applications that compute cell proofs frequently) wasn't covered by any test
in the C test suite -- the whole suite only loads the trusted setup with
precompute=0. This adds a test that computes cells/proofs with
precomputation enabled and checks the result matches the non-precomputed
path exactly.

Tested with `make test`, `make sanitize`, and `make coverage` -- the
previously-0%-covered precompute branches in `setup.c` and `fk20.c` are now
exercised.

### DIFF
--- a/src/test/tests.c
+++ b/src/test/tests.c
@@ -1934,6 +1934,44 @@ static void test_recover_cells_and_kzg_proofs__succeeds_random_blob(void) {
     }
 }
 
+static void test_compute_cells_and_kzg_proofs__succeeds_with_precompute(void) {
+    C_KZG_RET ret;
+    FILE *fp;
+    KZGSettings s_precompute;
+    Blob blob;
+    Cell cells[CELLS_PER_EXT_BLOB];
+    Cell cells_precompute[CELLS_PER_EXT_BLOB];
+    KZGProof proofs[CELLS_PER_EXT_BLOB];
+    KZGProof proofs_precompute[CELLS_PER_EXT_BLOB];
+    int diff;
+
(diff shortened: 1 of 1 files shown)


jtraglia:
Hi, thanks for the contribution. Just curious how you stumbled upon this? I suppose extra coverage is fine, but we are very confident that this function works as intended.

cavdarahmet:
Thanks for taking a look! I was doing a coverage-driven pass over the
codebase and noticed the `precompute` branch in `init_fk20_multi_settings`
(setup.c) and `compute_fk20_cell_proofs` (fk20.c) had 0% test coverage --
verified with `llvm-cov show`, every line inside the `if (precompute)`
blocks reports 0 hits, while the sibling non-precompute path is fully
exercised. This stood out because the README recommends `precompute=8`
or `precompute=9` for applications that call these functions frequently.

Before writing anything, I wanted to rule out an actual bug rather than
just closing a coverage gap, so I independently built the Python
bindings and compared `compute_cells_and_kzg_proofs` output across
precompute=0/8/9 on a few blobs (including an all-zero one) -- all
byte-identical and independently valid. Only after confirming that did
I add this as a regression test, which now asserts that equality
directly via memcmp.

Totally understand if you'd rather keep the test minimal or drop it --
my goal wasn't to flag a current bug (there isn't one), but to add a
guard so a future change to the precompute path can't silently regress
without any test catching it, especially since it's the path your own
docs recommend for frequent use.

PR #658, opened 2026-08-11
Fix YAML folding

rubo:
Fixed YAML folding, added missing `|` to the `Publish package` step


jtraglia (approved):
Haha oops. I missed this too. Thanks.

PR #660, opened 2026-08-17
Add function to recover cells (and not proofs) to bindings

jtraglia:
I am working on a separate library which uses a GPU to compute cell KZG proofs. In order to use that when recovering cells/proofs, clients will need the ability to recover cells (ie blobs) and *not* the proofs, so that the other library can handle that part. For that reason, this PR introduces a "recover cells" function to all of the bindings.

### DIFF
--- a/bindings/csharp/Ckzg.Bindings/Ckzg.cs
+++ b/bindings/csharp/Ckzg.Bindings/Ckzg.cs
@@ -248,6 +248,36 @@ public static partial class Ckzg
     }
 
     /// <summary>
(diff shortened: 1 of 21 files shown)


cskiraly:
@jtraglia , maybe worth checking https://github.com/ethereum/go-ethereum/pull/35531 and related PRs I've added to Geth, where I was tackling similar issues as well.

jtraglia:
So this isn't actually necessary for me anymore. But I'll leave it open for a bit while I consider merging it anyway. As you mentioned, it might still be valuable.

PR #661, opened 2026-08-17
perf(fft): iterative in-place fr_fft with staged twiddle lookup

cskiraly:
Note: this PR is part of a set of changes coming out of my AI-assisted performance investigation of the
EIP-7594 libraries. The analysis and implementation were done with Anthropic's Claude (Opus/Fable), and
cross-reviewed with OpenAI's Codex (GPT-5.6). The change is differential-tested against the previous
implementation, which is kept as a reference in the test suite — byte-for-byte across every power-of-two size
from 1 to 8192 — clean under the sanitizers and the static analyzer, and the consensus spec vectors pass
through the bindings.

This is a small PR, and it only touches the scalar-field FFT:
`compute_cells`: **2.50 ms → 2.32 ms** (−7%); the transform itself: **0.95 ms → 0.82 ms** (−14%) at n=8192.

## Summary

`fr_fft_fast` is the classic recursive formulation: each level recurses twice with doubled strides and writes
out of place, and every butterfly pays a `blst_fr_mul` — including the ones whose twiddle is `w^0 = 1`.

This makes it iterative: one O(n) pass gathers the input in bit-reversed order (an amortized reversed-carry
counter, not a per-element bit reversal — the per-element version measures *slower* than the recursion), then
log2(n) stages of in-place butterflies. **No new tables**: the existing natural-order `roots_of_unity` array
already contains every twiddle at every power-of-two stride, so each stage just reads it at a different step.
The `j = 0` butterfly skips its multiplication, which removes `n − 1` of the `(n/2)·log2(n)` multiplications —
**15.4% of them at n=8192** — and is exact, so output is byte-identical, not merely equivalent.

`fr_ifft` shares the core; its final 1/n scaling is unchanged. The G1 FFT is not touched.

One internal simplification fell out: the recursive helper's `stride` parameter is gone. Its non-recursive
entry calls only ever passed 1 — the strides of 2, 4, … existed only inside the recursion's own even/odd
traversal, which the bit-reversed gather replaces — and the in/out no-overlap contract the callers (recovery,
fk20, eip7594, poly) already satisfied is now documented on the helper.

## What this is and isn't

Motivation for calibration: go-eth-kzg recently delegated its scalar FFT to gnark-crypto's iterative
staged-twiddle implementation, and on the same machine its RS encode runs 1.60 ms against this library's 2.50.
This PR ports the *structural* part of that difference — iteration order, in-place stages, table lookup,
skipped identity multiplications — and that part is worth the ~14% here. The remainder of the gap lives in
gnark's fused add/sub butterfly assembly and per-operation field arithmetic, which cannot be ported without
changes to blst itself, so this PR deliberately does not chase it. The proof-bearing calls
(`compute_cells_and_kzg_proofs`, recovery-plus-proofs) are elliptic-curve-dominated and do not move.

## Correctness

- **Structurally the same butterfly DAG, and differential-tested byte-for-byte.** The equivalence argument is
  that only the scheduling of independent butterflies changes and the skipped multiplications are by the exact
  Montgomery one; the test pins it: the recursive implementation survives as `reference_fr_fft_recursive` in
  `tests.c`, and `test_fr_fft__matches_recursive_reference` compares `fr_fft` and `fr_ifft` against it with
  `memcmp` for every power-of-two size 1..8192, plus an `ifft(fft(x)) == x` roundtrip, with the suite's
  deterministic RNG.
- **The consensus spec vectors pass** through the Go bindings (run with a fresh build cache).
- `make test` 94/94; address and undefined sanitizers pass; `make analyze` clean; clang-format clean.

## Benchmarks

Apple M1, Go bindings at `precompute=6`, interleaved prebuilt binaries, medians of 6:

| | main | this PR | |
|---|---:|---:|---|
| `fr_fft`, n=8192 (C, `-O2`) | 0.95 ms | **0.82 ms** | −14% |
| `fr_ifft`, n=8192 | 1.05 ms | **0.93 ms** | −11% |
| `ComputeCells` | 2.504 ms | **2.323 ms** | −7.2% |
| `ComputeCellsAndKZGProofs` | 206.1 ms | 206.5 ms | ~ |
| `RecoverCellsAndKZGProofs` | 224.9 ms | 225.7 ms | ~ |

The last two *should* be unchanged — they are EC-dominated — and that they are is a check on the measurement.

## Limits

- Measured on one machine, `darwin/arm64`. Not reproduced elsewhere.
- This is a minor win stated as such: ~14% of the transform, −7% on the one API call that is FFT-bound.
- Independent of the recovery PR from the same investigation: the two merge cleanly (no shared hunks), and
  measured together recovery-plus-proofs goes 224.9 → 208.9 ms (the recovery PR's contribution; this PR's share
  there is within noise).

### DIFF
--- a/src/eip7594/fft.c
+++ b/src/eip7594/fft.c
@@ -58,30 +58,60 @@ static const fr_t INV_RECOVERY_SHIFT_FACTOR = {
 /**
  * Fast Fourier Transform.
  *
- * Recursively divide and conquer.
+ * Iterative decimation-in-time: gather the input in bit-reversed order, then run log2(n) stages
+ * of in-place butterflies over blocks of doubling size. The roots table is in natural order and
+ * spans the full domain, so it already holds every twiddle at every power-of-two stride; each
+ * stage just reads it at a larger step. The first twiddle of every block is w^0 = 1, so that
+ * butterfly is a plain add/sub, saving n - 1 of the (n/2) * log2(n) multiplications.
  *
  * @param[out]  out             The results, length `n`
- * @param[in]   in              The input data, length `n * stride`
- * @param[in]   stride          The input data stride
+ * @param[in]   in              The input data, length `n`
  * @param[in]   roots           Roots of unity, length `n * roots_stride`
  * @param[in]   roots_stride    The stride interval among the roots of unity
  * @param[in]   n               Length of the FFT, must be a power of two
+ *
+ * @remark The input and output arrays must not overlap.
  */
 static void fr_fft_fast(
-    fr_t *out, const fr_t *in, size_t stride, const fr_t *roots, size_t roots_stride, size_t n
+    fr_t *out, const fr_t *in, const fr_t *roots, size_t roots_stride, size_t n
 ) {
-    size_t half = n / 2;
-    if (half > 0) {
-        fr_t y_times_root;
-        fr_fft_fast(out, in, stride * 2, roots, roots_stride * 2, half);
-        fr_fft_fast(out + half, in + stride, stride * 2, roots, roots_stride * 2, half);
-        for (size_t i = 0; i < half; i++) {
-            fr_mul(&y_times_root, &out[i + half], &roots[i * roots_stride]);
-            fr_sub(&out[i + half], &out[i], &y_times_root);
-            fr_add(&out[i], &out[i], &y_times_root);
+    if (n < 2) {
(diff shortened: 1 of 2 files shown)


cskiraly:
test failed with `urllib.error.HTTPError: HTTP Error 504: Gateway Timeout` . Please re-run the failing test

PR #662, opened 2026-08-18
Reformulate verify_kzg_proof to drop the G2 scalar multiplication

rakita:
Move `z` from G2 to G1 in the single-proof pairing check via bilinearity `e(Q, -[z]) = e(-z·Q, [1])`, so both G2 pairing inputs collapse to constants (the G2 generator and `[X] = s->g2_values_monomial[1]`):

```
e(P - [y] + z·Q, [1]) == e(Q, [X])
```

This replaces the per-call G2 scalar multiplication (and the G2 subtraction) with a single cheaper G1 scalar multiplication.

Notably, c-kzg already uses this reformulation in one place: [`verify_kzg_proof_batch`](https://github.com/ethereum/c-kzg-4844/blob/b7e4098813e02c866d8def4e88524be8a6bedc7f/src/eip4844/eip4844.c#L733-L751) computes `rhs = Σ rⁱ(Cᵢ - [yᵢ]) + Σ rⁱzᵢQᵢ` and checks `e(Σ rⁱQᵢ, [X]) == e(rhs, [1])`. The new single-proof code mirrors it completely — same helpers (`g1_mul` on the generator for `[y]`, `g1_sub`, `g1_add`), same operand order, same two constant G2 inputs, and the same `pairings_verify` argument order — making it exactly the n = 1 specialization (where r⁰ = 1, and the one-element lincomb is a direct scalar multiplication). So the single-proof and batch paths now perform the same check, and the now-dead `g2_mul`/`g2_sub` helpers are removed from `eip4844.c` (`g2_mul` moves into `tests.c`, which still uses it for the `pairings_verify` tests).

Infinity semantics are unchanged: `pairings_verify` uses the per-pair `blst_miller_loop`, which handles infinity, and an infinity proof still degenerates to the same `commitment == [y]` condition as before.

Tests: adds an all-finite tampered-`y` rejection (pins the pairing check to the exact evaluation relation rather than the field-element guards) and an infinity-rhs/finite-proof rejection (commitment = proof = G1 generator, `y = 0`, `z = r - 1`, making `P - [y] + z·Q` the point at infinity while the proof is finite — valid only if the proof is infinity too, so it must be rejected). Full unit suite and the official reference vectors (via the Rust bindings) pass.

Quick benchmark (portable blst build, 2000 iters on the reference vector, interleaved runs): ~632–639 µs/op before, ~590–615 µs/op after, roughly a 4–7% speedup of the full `verify_kzg_proof` call. A natural follow-up would be caching Miller-loop line preparation for the two now-constant G2 points.

This mirrors the equivalent change made in revm (bluealloy/revm#3836).

### DIFF
--- a/src/eip4844/eip4844.c
+++ b/src/eip4844/eip4844.c
@@ -105,32 +105,6 @@ static C_KZG_RET fr_batch_inv(fr_t *out, const fr_t *a, int len) {
     return C_KZG_OK;
 }
 
-/**
- * Multiply a G2 group element by a field element.
- *
- * @param[out]  out The result, `a * b`
- * @param[in]   a   The G2 group element
- * @param[in]   b   The multiplier
- */
-static void g2_mul(g2_t *out, const g2_t *a, const fr_t *b) {
-    blst_scalar s;
-    blst_scalar_from_fr(&s, b);
-    blst_p2_mult(out, a, s.b, BITS_PER_FIELD_ELEMENT);
-}
-
-/**
- * Subtraction of G2 group elements.
- *
- * @param[out]  out The result, `a - b`
- * @param[in]   a   A G2 group element
- * @param[in]   b   The G2 group element to be subtracted
- */
-static void g2_sub(g2_t *out, const g2_t *a, const g2_t *b) {
-    g2_t bneg = *b;
-    blst_p2_cneg(&bneg, true);
-    blst_p2_add_or_double(out, a, &bneg);
-}
-
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 // BLS12-381 Helper Functions
 ////////////////////////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


4waan:
I was thinking of adding a test `test_g1_add__doubles_equal_points` because for a wrong sum in proof generation paths (`g1_fft_fast()` butterfly `(g1_add(&out[i], &out[i], &y_times_root))` that catches up to`g1_fft()` in fk20.c and setup.c) a rejected proof would be better. 

since `g1_add()` is `blst_p1_add_or_double()` its fine for now. 

glad to ship this separately if you'd like to keep the pr focused, because `test_verify_kzg_proof__fails_infinity_rhs_finite_proof` doesn't actually catch the doubling behaviour. 
With `commitment = proof = G1 and z = BLS_MODULUS - 1` the addition is `G1 + (-G1)` and that would be `a == -b,` not a == b, and blst's incomplete formula gets a == -b right (H = U2 - U1 = 0, so Z3 = 0, and infinity is correct there).

PR #666, opened 2026-09-14
Add Windows ARM64 support

ndabas:
As the title says -- I have added builds/tests for Python, C#, Rust, and Node.js.


flcl42 (approved):

ndabas:
@flcl42 thanks for approving and for kicking off the checks.

This PR does not include any changes to the Nim workflow but that job is failing. I did do some experiments and it looks like it is related to Nim being unable to copy a directory of symlinks, and it works if I disable symlinks in git for that job. I did actually [add that commit](https://github.com/ethereum/c-kzg-4844/commit/be9047444af8772a24d2f72da2202c3a65df984f) to this PR but then removed it, because it's not really part of what we're doing in this PR, and also the Windows runners should actually have symlinks properly enabled anyway, so it seemed like a hacky fix.

Let me know if you would like me to add that quick fix back to get the workflow to run.

.github/workflows/python-release.yml
jtraglia, line 1: Hey @ndabas, this is just for the release. It appears that this PR does not update `python-tests.yml`. Could you update it too?

ndabas: @jtraglia thanks for pointing that out -- I have updated `python-tests.yml` as well. I did try a few things but the latest push should be all good.

jtraglia (approved):
Thanks @ndabas! LGTM 👍

PR #668, opened 2026-09-21
nim: avoid allocation for global context

arnetheduck:
KZGSettings is a trivial and small type - instead of heap-allocating it, we can simply make it part of the global context so that we don't have to involve the nim alloctor - apart from being simpler, this has the collateral benefit that the setup can be freed in a different thread than it was allocated (due to limitations in the refc allocator).

Although it's not strictly needed (this gets loaded from the C header file anyway), this PR also adds the KzgSettings fields which gives sizeof information to Nim - `-d:checkAbi` in tests ensures that Nim and C agree on them.

While we're here, get rid of some pre-1.6 code.


bindings/nim/kzg_abi.nim
```diff
@@ -83,11 +92,10 @@ type
   KzgProof* = KzgBytes48
 
   # A single cell for a blob.
-  KzgCell* {.importc: "Cell",
-    header: "ckzg.h", completeStruct.} = object
+  KzgCell* {.importc: "Cell", kzg_type.} = object
     bytes*: array[BYTES_PER_CELL, uint8]
 
-{.pragma: kzg_abi, importc, cdecl, header: "ckzg.h".}
+{.pragma: kzg_abi, importc, cdecl, header: "ckzg.h", noSideEffect.}
```
arnetheduck, line 98: this prepares kzg for an upcoming change where nim will consider C functions to have side effects unless otherwise noted, which would propagate to callers.

